PR #2, opened 2019-11-04
Documenting Process

kenoir:
In order that we work in a consistent and effective way
I've documented what I think should be/is currently our "process"
This replaces the existing definition of done and needs agreement from the team.


process.md
```diff
@@ -0,0 +1,57 @@
...
+
+Delivery over process - do not follow these guidelines if they prevent you from delivering value as defined by our objectives.
+
+## Projects
+
+Projects are quarterly iterations of the projects [available in Github](https://github.com/wellcometrust/platform/projects).
+
+Issues are binned or recycled on a quarterly basis.
+
+## Pull Requests (PRs)
```
jamieparkinson: I feel like (most of) these PR requirements might be more appropriate as ticket/issue requirements; if they only become visible when the PR is opened then that feels a bit too late to me - does that make sense? That is based on the assumption that usually tickets and PRs are one-to-one.

process.md
```diff
@@ -0,0 +1,57 @@
...
+
+```
+In order to make searches better for users
+We've ordered searches by date
+By implementing an update to the `ElasticFlarp GargleMap` in `object FleebChunker`
+
+Depends on #123, required for #456
+```
+ 
+## Definition of done
```
jamieparkinson: Further to the above, it would be nice to add some documentation of how we write issues/tickets (maybe add an issue template to the repo). So far I've struggled a bit with understanding issues that lack acceptance criteria and/or that seem to require context that isn't documented or linked to them.

process.md
```diff
@@ -0,0 +1,57 @@
...
+All Pull Requests should have in general:
+
+- What is the benefit of this change (and for who)
+  e.g. In order to make searches better for users
+  
+- What is the change functionally
+  e.g. We've ordered searches by date
+  
+- What is the change
+  e.g. By implementing an update to the `ElasticFlarp GargleMap` in `object FleebChunker`
```
jamieparkinson: Do we think that this is necessary if we're writing sufficiently modular PRs? Maybe I'm misunderstanding the example but it seems like the diff is the documentation here

process.md
```diff
@@ -0,0 +1,57 @@
...
+
+Projects are quarterly iterations of the projects [available in Github](https://github.com/wellcometrust/platform/projects).
+
+Issues are binned or recycled on a quarterly basis.
+
+## Pull Requests (PRs)
+
+All Pull Requests should have in general:
+
+- What is the benefit of this change (and for who)
```
jamieparkinson: nit: s/benefit/outcome?

jamieparkinson: even bigger nit, I'm sorry that I'm like this: s/who/whom

kenoir: I love it. I'm a sucker for nit picks.

process.md
kenoir: Intended that you maybe write a one liner technical overview - not describe the entire change - bad example probably?

process.md
kenoir: Intended that the above _is_ that, seems like there is something i'm missing. Probably need to talk this through in person/at retro?

process.md
kenoir: > nit: s/benefit/outcome?

This is why would should do it, next point it what it is.

jamieparkinson: Ah yeah, I was thinking outcomes in the sense of "user outcomes", this is clearer

kenoir: This is deliberately not a BDD "narrative" because _why_ is more important than _what_.

PR #3, opened 2019-12-16
Add some notes on searching with the CALM API

alexwlchan:


README.md
```diff
@@ -1,9 +1,8 @@
...
 
-This is documentation for Wellcome Collection developers. Documentation for users of the API can be found at https://developers.wellcomecollection.org.
-
+This is documentation for Wellcome Collection developers. Documentation for users of the API can be found at <https://developers.wellcomecollection.org>.
 
 * [Projects](projects/README.md): A list of projects and services supported by Wellcome Collection developers.
 * [Process](process.md): How we work.
 * [RFCs](rfcs): A collection of RFCs for Wellcome Collection developers.
-
+* [Research](research): Notes and research on technical stuff that might be useful later.
```
jamesgorrie: 🚀

research/2019-12-16_searching_with_the_calm_api.md
```diff
@@ -0,0 +1,241 @@
+# Searching with the CALM API
+
+### 16 December 2019, AWLC
```
jamesgorrie: `AWLC`?

research/2019-12-16_searching_with_the_calm_api.md
```diff
@@ -0,0 +1,241 @@
...
+
+This has some notes on using the CALM API, and an example script for searching all the records in the CALM catalogue.
+
+
+
+## Motivation
+
+As part of the born-digital content migration, I had a collection of directories with accession-like numbers, and I wanted to get a list of all the accession numbers in CALM, so I could match the directories to accession numbers:
```
jamesgorrie: Where did you get the list of directories?

research/2019-12-16_searching_with_the_calm_api.md
```diff
@@ -0,0 +1,241 @@
...
+
+  > In library science, an *accession* is all the material that gets collected at one time from a single source.
+  > The *accession identifier* is the unique identifier given to that batch of material.
+  >
+  > At Wellcome, the accession identifier is made of two parts: the *accession number* (sequential, currently up to 2551) and a *collection reference* (such as WT, PPMTW, SANCT).
+  > The two are written together, separated by a slash.
+  > Examples:
+  >
+  > The accession number is globally unique: if there's an accession `WT/123`, you won't see `123` with any other collection reference.
+</details>
```
jamesgorrie: What syntax is this?

jamesgorrie: Nice note though!

research/2019-12-16_searching_with_the_calm_api.md
```diff
@@ -0,0 +1,241 @@
...
+      ...
+    </xml>
+    ```
+
+    One would serialise to JSON as a string, the other a list.
+
+    This might indicate an error somewhere in CALM that we want to fix (and now we have the API, we can find all those places!)
+
+*   AFAICT, the CALM API is read-only -- so there's less risk of a coding error triggering irreparable damage to the CALM database.
+    (I thought I'd seen that written down somewhere, but I can't find the reference now.)
```
jamesgorrie: Saying this - I wonder if there's some long running queries we could run to lock it up?

research/2019-12-16_searching_with_the_calm_api.md
```diff
@@ -0,0 +1,241 @@
...
+## Searching the API
+
+Useful pieces for searching the API:
+
+*   **Make a Search request using the SOAP API.**
+
+    Example request/responses are here: <http://archives.wellcomelibrary.org/CalmAPI/help/example_search.htm>
+    This returns an XML response that tells you how many results there are.
+
+    I know you can search the "Catalog", "Accessn" and "Locations" databases; there might be more I'm unaware of.
```
amme2: There are lots more but not all of them are populated / relevant. Of those which are "Persons" is the most significant - contains authority records for people and functions, most notably in relation to the Wellcome Trust corporate archive.

research/2019-12-16_searching_with_the_calm_api.md
```diff
@@ -0,0 +1,241 @@
...
+    AAUICW_2185 ~> AAUICW/2185 ?
+    AAUNAM_2213 ~> AAUNAM/2213 ?
+    AAUNAM_2254 ~> AAUNAM/2254 ?
+
+Exporting a list of accession numbers from the CALM GUI would have been tricky (I don't have access to CALM on my Mac), so I tried using the CALM API to export the data.
+
+<details>
+  <summary>What is an accession/accession number?</summary>
+
+  > In library science, an *accession* is all the material that gets collected at one time from a single source.
```
amme2: I'd recommend saying 'In archives practice' here. Libraries tend to use 'acquisition' instead, which is subtly different to accession.

research/2019-12-16_searching_with_the_calm_api.md
alexwlchan: My initials

Because research notes are a point-in-time snapshot rather than a living thing, I thought it might be useful to record who wrote them – although all the info *should* be here, knowing who to ask is probably useful.

I'll replace it with a full name.

research/2019-12-16_searching_with_the_calm_api.md
alexwlchan: From the V drive; I'll clarify in the document.

research/2019-12-16_searching_with_the_calm_api.md
alexwlchan: This is the HTML5 `<details>` tag: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

It gives you a widget that allows you to selectively show/hide information. I expect at least some people reading this will know what an accession (number) is, so they can skip this bit – but if you don’t, you won’t be lost.

research/2019-12-16_searching_with_the_calm_api.md
alexwlchan: Probably, but I don’t know what those are. 😉

research/2019-12-16_searching_with_the_calm_api.md
alexwlchan: Indeed, "Persons" does return a bunch of search results!

research/2019-12-16_searching_with_the_calm_api.md
alexwlchan: This is the HTML5 `<details>` tag: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

It gives you a widget that allows you to selectively show/hide information. I expect at least some people reading this will know what an accession (number) is, so they can skip this bit – but if you don’t, you won’t be lost.

research/2019-12-16_searching_with_the_calm_api.md
```diff
@@ -0,0 +1,241 @@
...
+  > The *accession identifier* is the unique identifier given to that batch of material.
+  >
+  > At Wellcome, the accession identifier is made of two parts: the *accession number* (sequential, currently up to 2551) and a *collection reference* (such as WT, PPMTW, SANCT).
+  > The two are written together, separated by a slash.
+  > Examples:
+  >
+  > The accession number is globally unique: if there's an accession `WT/123`, you won't see `123` with any other collection reference.
```
amme2: Hmm, is this how Trudy expressed them in the spreadsheet? There a slight danger of accidentally duplicating catalogue references for early accessions and I wonder whether it would be sensible to adopt some other separator than the slash here to avoid confusion later. e.g. if Acc_No 1 was a WT accession, WT/1 could refer to _either_ WT accession 1 _or_ WT item 1 in the catalogue.

alexwlchan: Not sure, this might be a convention that I've accidentally invented! I think I saw what looked like `WT/123` hanging around in Preservica, which is where I got the idea. I can’t find anything from Trudy that specifically uses this convention. Definitely see the point about ambiguity.

I don't think Archivematica or the Wellcome Storage mind what you put here, so we can choose a different character. Free choice!

The V drive uses underscores as separators. We could do that.

If not, maybe a hash `#`, because that's commonly used to mean “number”? e.g. `WT#1` or even `WT#acc1`.

Your team will have to deal with these identifiers, so happy to take your lead on this one.

research/2019-12-16_searching_with_the_calm_api.md
```diff
@@ -0,0 +1,241 @@
...
+*   **I want to get everything.**
+
+    Make a Search request with database `Catalog` and search term `*`.
+    That tells you how many results there are, then iterate from `0 to (hit count – 1)` (the results are 0-indexed), and call SummaryHeader for each.
+
+*   **I want to get recently updated stuff.**
+
+    You don't want to make a SummaryHeader request for each record -- that's slow and unnecessary.
+
+    Make a Search request with database `Catalog` and search term `*`.
```
amme2: * as a wildcard sometimes gives slightly odd results in the Calm client, so it might be worthwhile comparing the results of this search in the API with a similar search in the client, to check they are as expected. * in the client only matches 'text-type' fields and only where the field is not empty, and there's a separate keyboard shortcut (which must map to something programatically but I don't know what) for 'find all records in database'. As there are no field(s) specified here, maybe this isn't an issue though... but then what is the purpose of the <elementSet>DC</elementSet> below?

alexwlchan:
> Happy from a clarity POV, might be worth making sure it's inline with @amme2's understanding and approved there before merging.

:nod:

I think we've skated over the issue of "what accession identifiers do we put in the storage service", and we should double-check that before we start migrating stuff.

alexwlchan:
Here's an example from Preservica where it's a bit iffy: http://sdb.wellcome.ac.uk/explorer/explorer.html#browse:7&cdc1a761-d04f-401c-99e6-316fa13e547d&T1AO/1

This is labelled as MS8565/1550/1. It's a collection of sixteen text files that appear to have been given individual reference numbers (although I can't see those as piece-level records in CALM, just an Item-level record with RefNo `MS8565`. That’s where I got the “accession identifier has a slash” from.

Quick refresher on records in the storage service. They have a two-part identifier:

- The *storage space* is the broad category of stuff, e.g. `digitised` or `born-digital`. Uncatalogued born-digital material is going to go into a space labelled `born-digital-accessions` (this has been discussed with Trudy), so it's visibly separated from the catalogued material.

   (The storage space cannot contain a slash.)

- The *external identifier* is a reference to an external system that you can use to look up material in the space. For example, in the `digitised` space we use the Sierra b numbers, and for the `born-digital` space probably the AltRefNo (tbc).

    (The external identifier can contain slashes.)

My current plan is to use `{collection reference}/{accession number}` as the external identifier for material in the uncatalogued space. Open to suggestions for alternatives from @amme2 or Trudy, but I want to decide soon-ish so I can get stuff out of the V drive.

research/2019-12-16_searching_with_the_calm_api.md
alexwlchan: The `<elementSet>DC</elementSet>` is something the CALM API docs suggest; I haven't experimented to see what happens if you omit that field.

alexwlchan:
I've had a chat with @amme2 about accession numbers and we've agreed an approach. I'm going to write that up as a separate howto. I've added a note about accession number ambiguity to this note, and I'll merge as is – the howto will be another PR.

PR #5, opened 2019-12-17
Add some instructions for uploading born digital accessions

alexwlchan:
This is based on a discussion with @amme2 just now.

We're going to store uncatalogued accessions under **their accession number only**, and then include the collection reference in `metadata.csv` (so if the number is mistyped, you can work out what collection it came from).

The "start a transfer in Archivematica" Lambda will need modifying to:

* Check these three columns are present
* Extract the value from the `accession_number` column, and pass that to Archivematica

Interested in hearing how people feel about this approach before I start writing code for it.


jtweed:
I'd like to do this in a way that it can be wired into Archivematica. Ie, we extend the metadata.csv import in Archivematica, not as a pre-ingest step.

alexwlchan:
I wonder if the `identifiers.json` that was mentioned a while back would work for this? I’ll have a look. https://www.archivematica.org/en/docs/archivematica-1.10/user-manual/transfer/transfer/#transfers-with-existing-persistent-identifiers

jtweed:
Possibly, but in both cases I think our problem is associating data with the overall transfer, rather than objects within it. Hmm. Can you set the accession ID when the automation tools create the transfer from the lambda?

alexwlchan:
> Can you set the accession ID when the automation tools create the transfer from the lambda?

Yes!

Sorry, I should have made this explicit – when the Lambda reads the `metadata.csv`, it will use the value in the `accession_number` field as the `accession` parameter to the "Start Transfer" API (see https://wiki.archivematica.org/Archivematica_API#Start_Transfer). It'll end up in the same place as if you'd typed into the Accession Number field in the dashboard UI, but with less clicking involved.

jtweed:
In that case, your plan makes perfect sense I think.

amme2:
Although the comment above about 'associating data with the overall transfer, rather than objects within it' also makes me think I'd like to start looking at the impact on the cataloguing workflow sooner rather than later. At that point we would also want to associate the data with objects within the transfer, or at least not lose some kind of connection to a specific accession package when objects are moved into the catalogued storage partition. 

That said, I still think this is a much neater solution for new accession workflow - JS just pointed out to me that occasionally collection references are changed at point of cataloguing, so this would also allow for that eventuality.

PR #8, opened 2020-01-28
RFC: Add pipeline tracing

jamieparkinson:
Rendered markdown: https://github.com/wellcomecollection/docs/blob/pipeline-tracing/rfcs/018-pipeline_tracing/README.md

### DIFF
--- a/rfcs/018-pipeline_tracing/README.md
+++ b/rfcs/018-pipeline_tracing/README.md
@@ -1,0 +1,28 @@
+## Pipeline Tracing
+
+When things go wrong in the pipeline, debugging them involves a confusing and slow mix of checking DLQs and reading through the logs of several separate services in order to figure out what went wrong, for which documents, and where. This slows down development across the catalogue and makes it harder to catch bugs.
+
+As well as more comprehensive, structured logging in the constituent services of the pipeline, it would be beneficial to track the flow of data through it, from the adapters right through to ingest.
+
+### Proposed solution
+
+The [OpenTracing](https://opentracing.io/) project defines a common API and methodology for distributed tracing: tracing requests through multiple services whilst propagating and accumulating context throughout, enabling profiling and monitoring of the whole application.
+
+Whilst tracing is typically used in a request/response context where traces/spans stack hierarchically, it can also be used for monitoring unidirectional data flow: ie, for a pipeline.
+
+Even with no further tracing within a service, a top-level trace can tell us which data is in which service, and when. This would be a great starting point for knowing where to focus debugging efforts.
+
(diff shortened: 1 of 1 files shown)


jamesgorrie:
Good start - could we make it a draft similar to [this](https://github.com/wellcomecollection/docs/tree/pipeline-tracing/rfcs/016-holdings_service#rfc016holdingsservice), then we can merge it in and keep adding.

warrd:
Nice work, looking forward to this making it's way in. Do the RFCs have a set format? I notice other ones have other sections like "Alternatives" etc

jamesgorrie:
> Do the RFCs have a set format?

@warrd 
Yes - looking at past, fully featured ones works well although always up for improving, as you can see if you look at the first ones.

Maybe we could have an RFC for the format of RFCs?

jamesgorrie:
@warrd [things like this do exist](https://github.com/rust-lang/rfcs/blob/master/0000-template.md)

jamesgorrie:
@jamieparkinson it would be good to get some alternatives, and think about any drawbacks to this response.

warrd:
:+1: cheers. Yeah I am aware of the format used for Rust RFCs (and @alexwlchan mentioned they based the idea on that), that's why I thought to ask.

jamieparkinson:
@jamesgorrie @warrd added some thoughts on alternatives and downsides; didn't go the full Rust RFC template route as I think the more specific implementation details will come out in the wash of actually doing this bit of work.

kenoir:
Approach to tracing is sound, opentracing is a good choice. 

Can we provide a specific example of where this would have helped us debug a pipeline issue? 

My impression is that there is a deeper problem identifying **what an error actually is in the applications**. If an error is never flagged up, we will have no mechanism to lead us to a trace.

Some questions I think we need to answer:

- How will this feed into alerting?
- Will this change how we deal with errors in the pipeline now? Will we stop retrying things that throw exceptions, or distinguish between things that need to be retried and things that don't?

kenoir:
> My impression is that there is a deeper problem identifying **what an error actually is in the applications**. If an error is never flagged up, we will have no mechanism to lead us to a trace.

After chatting with @jamieparkinson it is clearer that this will force us to add context to each step in the pipeline that will allow us to pull a set of pipeline events into a thread, thereby enabling more effective debugging.

> - How will this feed into alerting?
> - Will this change how we deal with errors in the pipeline now? Will we stop retrying things that throw exceptions, or distinguish between things that need to be retried and things that don't?

Are questions we should handle separately (not in this RFC).

jamieparkinson:
Merging in anticipation of further PRs on developments

PR #20, opened 2020-03-06
RFC: Data science in the pipeline

jamieparkinson:
https://github.com/wellcomecollection/docs/blob/data-science-pipeline/rfcs/021-data_science_in_the_pipeline/README.md

### DIFF
--- a/rfcs/021-data_science_in_the_pipeline/README.md
+++ b/rfcs/021-data_science_in_the_pipeline/README.md
@@ -25,9 +25,9 @@ We can write down some desiderata for any proposed solution:
 
 These suggest 3 distinct, but loosely coupled, services:
 
-- **Inferrer**: A Python service that provides a synchronous API that consumes whatever is needed to infer the new data, which it outputs.
+- **Inferrer**: A Python service that provides a synchronous API (most likely RESTful HTTP) that consumes whatever is needed to infer the new data, which it outputs.
 - **Model Trainer**: A Python service that can consume records from the catalogue index in bulk in order to train a model, and outputs/stores a persistent representation of this model for the *inferrer* to use.
-- **Inference Manager**: A Scala service that lives in the pipeline and contains "the usual" Wellcome message-passing, Akka, etc logic & libraries, which sychronously calls the `inferrer` and attaches the new data to the work/image before passing it along.
+- **Inference Manager**: A Scala service that lives in the existing pipeline and contains "the usual" Wellcome message-passing, Akka, etc logic & libraries, which sychronously calls the `inferrer` and attaches the new data to the work/image (by populating a field) before passing it along.
 
 The usage of these services would look like this:
 
--- a/rfcs/021-data_science_in_the_pipeline/README.md
+++ b/rfcs/021-data_science_in_the_pipeline/README.md
@@ -27,7 +27,7 @@ These suggest 3 distinct, but loosely coupled, services:
 
 - **Inferrer**: A Python service that provides a synchronous API that consumes whatever is needed to infer the new data, which it outputs.
 - **Model Trainer**: A Python service that can consume records from the catalogue index in bulk in order to train a model, and outputs/stores a persistent representation of this model for the *inferrer* to use.
-- **Inference Manager**: A Scala service that lives in the pipeline and contains "the usual" Wellcome message-passing, Akka, etc logic & libraries, which sychronously calls the *inferrer* and attaches the new data to the work/image before passing it along.
+- **Inference Manager**: A Scala service that lives in the pipeline and contains "the usual" Wellcome message-passing, Akka, etc logic & libraries, which sychronously calls the `inferrer` and attaches the new data to the work/image before passing it along.
 
 The usage of these services would look like this:
 
@@ -43,4 +43,12 @@ The usage of these services would look like this:
 
 > Are the network requests for (eg) image files going to be very expensive and/or slow? How could we mitigate this?
 
-> How do we deal with wanting to infer several separate things? A chain of inferrer-manager pairs would be simplest and cleanest, but this would require `n` times as many network requests for images as one inferrer doing all the work.
+> How do we deal with wanting to infer several separate things for one piece of content (eg a feature vector _and_ a palette vector for a single image)? A chain of inferrer-manager pairs would be simplest and cleanest, but this would require `n` times as many network requests for images as one inferrer doing all the work.
+
+> Should DS APIs be open to the public?
+
+> Should model artefacts be open to the public?
+
+> Should DS inferrers work only for things we know (ie feature vectors can be inferred only for images within the WC ecosystem, by passing an image ID) or with any content (ie feature vectors can be inferred for any image, by passing a publicly accessible URI)
+
+> It seems like we'll have to reinfer on every existing work/image when a model is retrained, which will significantly add to the expense of retraining based on the above considerations about network requests etc. Is this necessarily the case?
--- a/rfcs/021-data_science_in_the_pipeline/README.md
+++ b/rfcs/021-data_science_in_the_pipeline/README.md
@@ -6,7 +6,7 @@
 
 ## Motivation
 
-We want to be augment works and images with data inferred off of them using data science techniques: for example, feature vectors and colour palettes for images.
+We want to augment works and images with data inferred from them using data science techniques: for example, feature vectors and colour palettes for images.
 
 Currently, we do this by holding both some form of index (usually a set of points in a vector space) as well as a model in a separate service - for example, https://labs.wellcomecollection.org/feature-similarity. This has significant drawbacks: data is duplicated, outdated, patchy, and perhaps most importantly, must be searched as a wholly separate all-or-nothing operation outside of our ES indices.
 
--- a/rfcs/021-data_science_in_the_pipeline/README.md
+++ b/rfcs/021-data_science_in_the_pipeline/README.md
@@ -1,0 +1,46 @@
+# RFC 021: Data science in the pipeline
+
+**Status:** 🏗 Draft 🚧
+
+**Last updated:** 2020/03/06
+
+## Motivation
+
+We want to be augment works and images with data inferred off of them using data science techniques: for example, feature vectors and colour palettes for images.
+
+Currently, we do this by holding both some form of index (usually a set of points in a vector space) as well as a model in a separate service - for example, https://labs.wellcomecollection.org/feature-similarity. This has significant drawbacks: data is duplicated, outdated, patchy, and perhaps most importantly, must be searched as a wholly separate all-or-nothing operation outside of our ES indices.
+
+To resolve this, we want to bring data science services into the pipeline, and use them to augment works/images with data that can be indexed and searched in ES. This RFC details how that might look.
+
+## Proposal
+
+We can write down some desiderata for any proposed solution:
+
+- Data science (DS) logic can be written in Python
+- DS services do not know about our data model(s)
+- Pipeline services do not know about data science
+- DS services do not know about SQS, message passing, the Actor model, etc
+- DS models are persistent and separate to DS inferrers
+- DS models can be retrained on demand
+
+These suggest 3 distinct, but loosely coupled, services:
+
+- **Inferrer**: A Python service that provides a synchronous API that consumes whatever is needed to infer the new data, which it outputs.
+- **Model Trainer**: A Python service that can consume records from the catalogue index in bulk in order to train a model, and outputs/stores a persistent representation of this model for the *inferrer* to use.
+- **Inference Manager**: A Scala service that lives in the pipeline and contains "the usual" Wellcome message-passing, Akka, etc logic & libraries, which sychronously calls the *inferrer* and attaches the new data to the work/image before passing it along.
+
+The usage of these services would look like this:
+
+![An architecture diagram for data science services](https://user-images.githubusercontent.com/4429247/76087593-631cc700-5fae-11ea-819f-3545e6dc7c65.png)
+
+#### Implementation details
+
+- The inferrer and inference manager exist in one task definition (and therefore on one host). The inferrer is not marked as essential: crashes are tolerated. Timeouts and other failures are naturally handled by the queue redrive policies.
+- The model trainer runs automatically as soon as it starts: it is triggered by an autoscaling policy. Scaling up could be based on regular time intervals, messages in queues, lambdas etc, and scaling down would be done by the trainer itself.
+- The inferrer loads the model from S3 when it starts. Inferrer instances will be short-lived (as they'll scale to zero when not in use) so triggering a restart or loading a new model by other means is not necessary.
+
+### Questions & drawbacks
+
+> Are the network requests for (eg) image files going to be very expensive and/or slow? How could we mitigate this?
+
+> How do we deal with wanting to infer several separate things? A chain of inferrer-manager pairs would be simplest and cleanest, but this would require `n` times as many network requests for images as one inferrer doing all the work.


rfcs/021-data_science_in_the_pipeline/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+The usage of these services would look like this:
+
+![An architecture diagram for data science services](https://user-images.githubusercontent.com/4429247/76087593-631cc700-5fae-11ea-819f-3545e6dc7c65.png)
+
+#### Implementation details
+
+- The inferrer and inference manager exist in one task definition (and therefore on one host). The inferrer is not marked as essential: crashes are tolerated. Timeouts and other failures are naturally handled by the queue redrive policies.
+- The model trainer runs automatically as soon as it starts: it is triggered by an autoscaling policy. Scaling up could be based on regular time intervals, messages in queues, lambdas etc, and scaling down would be done by the trainer itself.
+- The inferrer loads the model from S3 when it starts. Inferrer instances will be short-lived (as they'll scale to zero when not in use) so triggering a restart or loading a new model by other means is not necessary.
```
harrisonpim: How are you imagining this scaling being done? 
Does this overlap with the work to move reindexing etc to spot instances?

jamieparkinson: This would be the same as things currently are in the pipeline (as in, we'd get it for free): if a service has nothing in its input queue, then it scales to zero. Isn't really affected by the spot instances work.

warrd:
Looks good. Got some questions:

1) Are you imaging the Inference Manager interacts with the Inferer using an http API? Not quite clear from the docs.

2) Is it a seperate index that you are writing these inferences to, and if so how does that interact with current system? Did you also consider adding inferences about a work to the standard works index (i.e. in a new or existing stage of the existing works pipeline)?

3) Does @harrisonpim envision any occasions where the ML models might need to be queried dynamically during in response to an API request, rather than pre computed and stored?

jamieparkinson:
> Are you imaging the Inference Manager interacts with the Inferer using an http API? Not quite clear from the docs.

Yes (although really it just needs to be TCP), will clarify that.

> Is it a seperate index that you are writing these inferences to?

No, the inferred data will just be a field on the work/image - the inference manager would just live in the normal pipeline. I'll clarify that too.

harrisonpim:
> 3\. Does @harrisonpim envision any occasions where the ML models might need to be queried dynamically during an API request, rather than pre computed and stored?

Possibly - something like a reverse image search / reverse palette search etc might require inference on a new image, but there are no hard plans for those things yet. Talking about pipeline services only here for now

rfcs/021-data_science_in_the_pipeline/README.md
```diff
@@ -0,0 +1,54 @@
...
```
kenoir: I do not think this is possible if the inference-manager and inferrer share a task definition.

rfcs/021-data_science_in_the_pipeline/README.md
```diff
@@ -0,0 +1,54 @@
...
+- **Inference Manager**: A Scala service that lives in the existing pipeline and contains "the usual" Wellcome message-passing, Akka, etc logic & libraries, which sychronously calls the `inferrer` and attaches the new data to the work/image (by populating a field) before passing it along.
+
+The usage of these services would look like this:
+
+![An architecture diagram for data science services](https://user-images.githubusercontent.com/4429247/76087593-631cc700-5fae-11ea-819f-3545e6dc7c65.png)
+
+#### Implementation details
+
+- The inferrer and inference manager exist in one task definition (and therefore on one host). The inferrer is not marked as essential: crashes are tolerated. Timeouts and other failures are naturally handled by the queue redrive policies.
+- The model trainer runs automatically as soon as it starts: it is triggered by an autoscaling policy. Scaling up could be based on regular time intervals, messages in queues, lambdas etc, and scaling down would be done by the trainer itself.
```
kenoir: The pattern elsewhere is to scale services on queue length - those services then poll for work and go away when scaled down. 

Implicit here is that the model trainer does not need to be informed about work to do (other than by its existence) and that the service itself knows when to scale down.

rfcs/021-data_science_in_the_pipeline/README.md
```diff
@@ -0,0 +1,54 @@
...
+- **Model Trainer**: A Python service that can consume records from the catalogue index in bulk in order to train a model, and outputs/stores a persistent representation of this model for the *inferrer* to use.
+- **Inference Manager**: A Scala service that lives in the existing pipeline and contains "the usual" Wellcome message-passing, Akka, etc logic & libraries, which sychronously calls the `inferrer` and attaches the new data to the work/image (by populating a field) before passing it along.
+
+The usage of these services would look like this:
+
+![An architecture diagram for data science services](https://user-images.githubusercontent.com/4429247/76087593-631cc700-5fae-11ea-819f-3545e6dc7c65.png)
+
+#### Implementation details
+
+- The inferrer and inference manager exist in one task definition (and therefore on one host). The inferrer is not marked as essential: crashes are tolerated. Timeouts and other failures are naturally handled by the queue redrive policies.
```
kenoir: This makes sense for 2 closely linked services talking synchronously to one another.

rfcs/021-data_science_in_the_pipeline/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+These suggest 3 distinct, but loosely coupled, services:
+
+- **Inferrer**: A Python service that provides a synchronous API (most likely RESTful HTTP) that consumes whatever is needed to infer the new data, which it outputs.
+- **Model Trainer**: A Python service that can consume records from the catalogue index in bulk in order to train a model, and outputs/stores a persistent representation of this model for the *inferrer* to use.
+- **Inference Manager**: A Scala service that lives in the existing pipeline and contains "the usual" Wellcome message-passing, Akka, etc logic & libraries, which sychronously calls the `inferrer` and attaches the new data to the work/image (by populating a field) before passing it along.
+
+The usage of these services would look like this:
+
+![An architecture diagram for data science services](https://user-images.githubusercontent.com/4429247/76087593-631cc700-5fae-11ea-819f-3545e6dc7c65.png)
```
kenoir: Where does this fit into the existing pipeline? What are things & augmented things?

rfcs/021-data_science_in_the_pipeline/README.md
```diff
@@ -0,0 +1,54 @@
...
+- DS services do not know about our data model(s)
+- Pipeline services do not know about data science
+- DS services do not know about SQS, message passing, the Actor model, etc
+- DS models are persistent and separate to DS inferrers
+- DS models can be retrained on demand
+
+These suggest 3 distinct, but loosely coupled, services:
+
+- **Inferrer**: A Python service that provides a synchronous API (most likely RESTful HTTP) that consumes whatever is needed to infer the new data, which it outputs.
+- **Model Trainer**: A Python service that can consume records from the catalogue index in bulk in order to train a model, and outputs/stores a persistent representation of this model for the *inferrer* to use.
```
kenoir: Will there ever be multiple models? How would you account for multiple model trainers?

rfcs/021-data_science_in_the_pipeline/README.md
kenoir: I had previously imagined that augmented data would be fed into the pipeline as another source, matched & merged as other data sources are.

rfcs/021-data_science_in_the_pipeline/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+- The inferrer and inference manager exist in one task definition (and therefore on one host). The inferrer is not marked as essential: crashes are tolerated. Timeouts and other failures are naturally handled by the queue redrive policies.
+- The model trainer runs automatically as soon as it starts: it is triggered by an autoscaling policy. Scaling up could be based on regular time intervals, messages in queues, lambdas etc, and scaling down would be done by the trainer itself.
+- The inferrer loads the model from S3 when it starts. Inferrer instances will be short-lived (as they'll scale to zero when not in use) so triggering a restart or loading a new model by other means is not necessary.
+
+### Questions & drawbacks
+
+> Are the network requests for (eg) image files going to be very expensive and/or slow? How could we mitigate this?
+
+> How do we deal with wanting to infer several separate things for one piece of content (eg a feature vector _and_ a palette vector for a single image)? A chain of inferrer-manager pairs would be simplest and cleanest, but this would require `n` times as many network requests for images as one inferrer doing all the work.
```
kenoir: I think this is resolved by treating augmented data as another source.

rfcs/021-data_science_in_the_pipeline/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+> Should DS APIs be open to the public?
+
+> Should model artefacts be open to the public?
+
+> Should DS inferrers work only for things we know (ie feature vectors can be inferred only for images within the WC ecosystem, by passing an image ID) or with any content (ie feature vectors can be inferred for any image, by passing a publicly accessible URI)
+
+> It seems like we'll have to reinfer on every existing work/image when a model is retrained, which will significantly add to the expense of retraining based on the above considerations about network requests etc. Is this necessarily the case?
```
kenoir: Aye. This is a good question - i'd like to understand how to handle this. Can we infer data about a new work from an existing model?

kenoir:
I think if we treat augmented data as another data source the inferrer implementation defined here might look a lot more like an existing pipeline service rather than as a sidecar of a pipeline service.

I'm guessing here that the inference manager drops into the pipeline as another step. I can imagine a situation where it is moved to the side (out of the critical path), or indeed removed entirely with an inferrer in place of an adapter.

rfcs/021-data_science_in_the_pipeline/README.md
jamieparkinson: To be clear - both would scale to zero

rfcs/021-data_science_in_the_pipeline/README.md
jamieparkinson: Yes - the scaling down part needs some more thought but that idea was intended to be explicit.

We discussed having something similar to a reindexer to feed the model trainer works but that would introduce a requirement for it to understand and work through the queues etc, which we didn't want.

rfcs/021-data_science_in_the_pipeline/README.md
jamieparkinson: An inference task definition could live anywhere in the pipeline after transformation. A thing has optional fields that contain data that the inference task definition might populate, after which the thing is augmented. I'll clarify that in the RFC. 

I don't really see the advantages of having augmented data as a new source - it would introduce a cycle into the pipeline (in that the source would need to have a source of existing works) and the merger would just be "undoing" this cycle.

We want the data science stuff to be source-agnostic, merely adding extra searchable information to works - in the absence of something like change feeds in ES it seems to me like that should live in the pipeline.

rfcs/021-data_science_in_the_pipeline/README.md
jamieparkinson: See later in the document. This mostly imagines that we have one of this trio of services per model / piece of functionality.

rfcs/021-data_science_in_the_pipeline/README.md
jamieparkinson: Would it? We'd still have to make the same sort of decisions in that source "generator", I think?

rfcs/021-data_science_in_the_pipeline/README.md
jamieparkinson: We can and should infer data from a new work using an existing model, which is why we can confidently say that the model trainer will run infrequently. The problem identified here is that a new model would likely invalidate the existing inferred data in the case of (eg) vectors.

kenoir:
Had a chat with @harrisonpim regarding this idea and had some new impressions:

- Treating augmented (ML) data as another data source creates a big loop (with many services to delay update), whereas the model described here has a tight loop to update a work.
- Using the merger & matcher to "patch" data onto an existing work moves the logic about how to do that away from the inference step where it would be more easily understood. 
- Feeding the data into the pipeline as another source would force us to create a "work" for that augmented data which feels superfluous.  
- Another adapter would mean a transformer, again unnecessary as we already know where the created data fits at the inferrer step.

rfcs/021-data_science_in_the_pipeline/README.md
jamieparkinson: Exactly the intention - not sure if you're saying they shouldn't be like this?

jamieparkinson:
Ha, I think I've just said almost exactly that in different words in my replies! Good that we're independently on the same page

rfcs/021-data_science_in_the_pipeline/README.md
kenoir: But that's ok isn't it? The data in the catalogue API will be eventually consistent?

rfcs/021-data_science_in_the_pipeline/README.md
kenoir: They should be like this.

rfcs/021-data_science_in_the_pipeline/README.md
jamieparkinson: Eventually but maybe with a period of being universally nonsensical in between, I think? @harrisonpim knows better

rfcs/021-data_science_in_the_pipeline/README.md
kenoir: Before the ingestor perhaps?

rfcs/021-data_science_in_the_pipeline/README.md
kenoir: > Ha, I think I've just said almost exactly that in different words in my replies! Good that we're independently on the same page

I needed to clear it up as the conversations i've had with @jtweed before you or @harrisonpim applied thought to this presumed the ML data as new data source model. Re reading this and talking to Harrison makes me feel the approach described here fits better.

rfcs/021-data_science_in_the_pipeline/README.md
jamieparkinson: Yeah, that's where I imagined it. Makes sense for doing this for images as well.

kenoir:
> Ha, I think I've just said almost exactly that in different words in my replies! Good that we're independently on the same page

I needed to clear it up as the conversations i've had with @jtweed before you or @harrisonpim applied thought to this presumed the ML data as new data source model. Re reading this and talking to Harrison makes me feel the approach described here fits better.

rfcs/021-data_science_in_the_pipeline/README.md
jamieparkinson: > Can we parallel mapreduce the process?

I did wonder about something like this - is it insanity to download images in one place (map) and to send the inferrers big old byte streams (or form multipart uploads) for them to work on (reduce)?

rfcs/021-data_science_in_the_pipeline/README.md
harrisonpim: We can definitely infer on new works with existing models - that's the whole point!

The (potential) issue is that we might add a bunch of new data to our catalogue which deviates significantly from what already exists, so would be poorly described by a model trained on the old data. In that case we'd like to retrain the model to describe the new data as well as the old stuff.

If the model aims to label things with eg plain text subjects (`cat`, `dog`, `house`), the model can be retrained to output new classes without invalidating the old inferred data - no issues there.

However, if the model outputs a _feature vector_ as its description of the image, text, or whatever, there's no guarantee that the space described by that feature vector has any correspondence with the one from the old model. Oversimplifying, it would be like one model learning to describe things in English, and a subsequent model learning to describe them in French. 
If we then want to compare descriptions of works (for image similarity, etc), we need them all to be described by the same model for that comparison to make sense. Unless we can figure out a way of updating models without invalidating the old descriptions, new descriptions would need to be inferred for every existing work, as well as the new ones.

kenoir: So feels like retraining necessitates a reindex (in some cases) - which we could certainly do, but is a motivation for ensuring that reindex cost is low. 

Another mitigation might be removing the "one message per work * number of services" overhead of a reindex and either:

- reindexing further along (i.e. after matching/merging)
- optionally have a way of batch updating inferred data

rfcs/021-data_science_in_the_pipeline/README.md
kenoir: My impression is there are a cascade of options for holding image data closer to the inferers depending on how expensive in time & money the process is.

To address parallelising generating feature vectors we could have multiple inferrer services, one for each feature type we need to compute, an inferrer-manager sends messages to multiple inferrers that do their work and report back to the manager which handles completing on success/failure over multiple requests. I think I prefer this to chaining as we can parallelise compute - though perhaps this is not necessary to start with as we'll only have one feature vector to compute?

The inferrers (either as you describe above or as I suggest) could be given access to the image data in a few ways, (in rough order of speed of access):

- Downloading the images from S3 on demand
- Utilising the IIIF API to download smaller images
- Utilising some network accessible cache of those smaller images (possibly utilise the DLCS cache or have our own).
- Utilising some in-memory cache of those smaller image.

The inferrers could be given more access to host hardware by moving away from Fargate and doing something a bit more exotic with an EC2 cluster host too.

I'd start with the easy solution (getting little images from DLCS and utilising its cache), and move on to more interesting infrastructure as required.

kenoir: Oop, and I forgot batch processing stuff here is an option too.

PR #21, opened 2020-03-20
Adds logging RFC

kenoir:
Aiming to fulfil some of: https://github.com/wellcomecollection/docs/blob/master/rfcs/019-platform_reliability/observability.md

[A better view of this PR](https://github.com/wellcomecollection/docs/tree/71fa55bd077a37f574e7d0f6d437a08c98cac0c0/rfcs/021-logging)


.gitignore
```diff
@@ -0,0 +1,2 @@
+.idea*
+.DS_Store
```
jamesgorrie: I put these in my global git ignore as they feel environment specific, rather than project specific.

rfcs/021-logging/README.md
```diff
@@ -0,0 +1,113 @@
...
+  - (con) Multiple `???-transit` services would still be required.
+- Provide a single collector service across the platform utilising [AWS PrivateLink](https://docs.aws.amazon.com/vpc/latest/userguide/endpoint-service.html).
+  
+  VPC Endpoint Services (AWS PrivateLink) allows you to share an [NLB (Network Load Balancer)](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html) fronted service between VPCs and between AWS accounts. We have made use of AWS PrivateLink to share on-premises VPN accessible services (e.g. our Workflow service "Goobi") between accounts with success in the past.
+  
+  Endpoint services would be provisioned alongside network and account configuration in the [platform-infrastructure](https://github.com/wellcomecollection/platform-infrastructure) terraform stack. 
+  
+  - (pro) One service with central configuration for Elasticsearch authentication.
+  - (pro) We have experience with successfully using PrivateLink.
+  - (con) Requires some log collector infrastructure and network configuration though centralised and with a single instance of the collector service.
```
jamesgorrie: Not sure if it's a pro or a con here, but the fact that if it goes down, all our logging goes down.

Would there still need to be a sidecar that talks to this service from each of the services?

jamesgorrie: Just saw the diagram - makes more sense.

.gitignore
kenoir: I think they are common enough to justify being per project, 2 lines in a .gitignore is pretty cheap. Also having these in a global .gitignore is a good idea though.

rfcs/021-logging/README.md
jamesgorrie: So the "log collector endpoint" in the service account (not platform) is an ALB? Not sure I grasp what those _are_, but get what they're _doing_.

kenoir: > Not sure if it's a pro or a con here, but the fact that if it goes down, all our logging goes down.

Some truth to that, though I think we can mitigate that risk as we do in our other services (multiple tasks/cross AZ distribution).

> Would there still need to be a sidecar that talks to this service from each of the services?

This is described in the PR - the "Firelens" logging adapter feeds logs to a Fluent Bit sidecar, though AWS takes care of providing that image and we can package it so that it is transparent to consumers. I read somewhere that this is described as a "hidecar"!

kenoir:
> Nice set of options, shall the team run through them move to phase 2?

If we're happy to this approach is sound, next step is to spike a proof a concept, it shouldn't take longer than a day to build something. We should take that into planning.

rfcs/021-logging/README.md
jamesgorrie: Because of the wordplay, I approve whole-heartedly.

.gitignore
jamesgorrie: Yip, don't mind, just mentioning.

rfcs/021-logging/README.md
```diff
@@ -0,0 +1,113 @@
...
+At present we route logs through a cluster-local Logstash service that is configured to authenticate with a Elastic Cloud hosted Elasticsearch instance and forward logs there with several `logstash-transit` services across the estate.
+
+There are a few options for routing to Elasticsearch.
+
+- Directly to Elasticsearch from the Fluent Bit sidecar.
+  - (pro) No other log forwarding infrastructure.
+  - (con) Elasticsearch config would be required in each service.
+- Forwarding logs to a cluster-local collector as we do now.
+  - (???) We can use the Fluent Bit [`forward`](https://docs.fluentbit.io/manual/input/forward) input and so replace Logstash with Fluent Bit which _might_ be more reliable.
+  - (con) Multiple `???-transit` services would still be required.
```
jamesgorrie: This is still the case, but just mode hidden though?

.gitignore
jamieparkinson: Some useless input: I like to use [this gitignore](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) for personal projects or ones on which I know other devs will be using JetBrains IDEs

rfcs/021-logging/README.md
kenoir: We still need to route logs from there to ElasticSearch. What we are discussing in this section is how to do that. Lots of collectors (logstash-transit services) or one collector service made available via PrivateLink.

PR #25, opened 2020-04-20
Write-up of the Preservica verification

alexwlchan:


kenoir:
This all seems sensible to me. Might it be valuable for @amme2 to look over this?

projects/2020-q1-preservica-verification.md
```diff
@@ -0,0 +1,78 @@
...
+
+
+
+## The verification process: a how-to
+
+*   **Look for files that were copied byte-for-byte to the storage service.**
+
+    The Preservica metadata lets us work out where a file would have been copied to in the storage service, if it was copied -- space, external identifier, and filename within the bag.
+
+    If that bag exists, and it contains a file with matching filename, file size, and SHA-1 checksum, this file has been copied successfully.
```
amme2, line 40: How do you know that the file within the bag has copied successfully and has not corrupted? Doesn't this only test that the filenames and sizes match, rather than using the SHA-1 hash to verify the continued integrity of the transferred file?

projects/2020-q1-preservica-verification.md
```diff
@@ -0,0 +1,78 @@
...
+*   **Ignore files that had been replaced by newer versions.**
+
+    Some b numbers have multiple versions -- for example, if a book was digitised and then one of the images was replaced.
+    This would create two deliverable units.
+
+    If the same catalogue reference has multiple DUs, ignore all files associated with all but the newest version of that DU.
+
+*   **Replace files that were corrupted during the migration process.**
+
+    The matching process found some JP2 files that were corrupted when they were copied to the storage service.
```
amme2, line 67: Ah, see above... I don't see where the test for corruption is documented here.

amme2:
Looks fine to me from a born digital perspective, with exception of 2 minor comments added.

projects/2020-q1-preservica-verification.md
alexwlchan: Possibly this sentence isn't clear enough? How about:

> If that bag exists, look inside the bag for a file with a matching filename. If such a file exists, and the size and SHA-1 checksum of the file in the storage service match those of the file in Preservica, then this file has been copied successfully.

amme2: Better. What wasn't clear to me was the checksum supplied by Preservica is verified against a checksum generated using the same algorithm as part of the bagging process (I think that's what you're saying?). I read it as the checksum supplied by Preservica had been copied and pasted - so to speak - into the bag.

PR #26, opened 2020-04-27
RFC: Images endpoint

jamieparkinson:
https://github.com/wellcomecollection/docs/blob/images-endpoint/rfcs/023-images-endpoint/README.md

### DIFF
--- a/rfcs/023-images-endpoint/README.md
+++ b/rfcs/023-images-endpoint/README.md
@@ -19,10 +19,10 @@ The initial images endpoint will be in **beta**, and therefore will be liable to
 ```
 {
   "id": String,
-  "location": DigitalLocation,
+  "locations": List[DigitalLocation],
   "source": {
     "id": String,
-    "sourceType": "Work"
+    "type": "Work"
   },
   "type": "Image"
 }
@@ -47,6 +47,6 @@ This is intentionally minimal at this point as we decide what fields an image sh
 
 - `query` - text search query
 - `include` - optional list of extra fields _not including `visuallySimilar`_
-- `items.locations.license` - license to filter for
+- `locations.license` - license to filter for
 - `page`
 - `pageSize`
--- a/rfcs/023-images-endpoint/README.md
+++ b/rfcs/023-images-endpoint/README.md
@@ -6,9 +6,9 @@
 
 ## Context and requirements
 
-- Images are now entities in the catalogue API
-- We want to be able to search for images, and to fetch specified images
-- The images endpoint should behave similarly to the works endpoint
+- There is a new entity in the catalogue API: `Image`. These are conceptual, like works, and consist of images from the visual collections and interesting images from inside books and other materials.
+- We want to be able to search for images, and to fetch specified images.
+- The images endpoint should be consistent with the works endpoint.
 
 ## Proposal
 
@@ -20,11 +20,15 @@ The initial images endpoint will be in **beta**, and therefore will be liable to
 {
   "id": String,
   "location": DigitalLocation,
-  "parentWork": String
+  "source": {
+    "id": String,
+    "sourceType": "Work"
+  },
+  "type": "Image"
 }
 ```
 
-This is intentionally minimal at this point as we decide what fields an image should have.
+This is intentionally minimal at this point as we decide what fields an image should have; it's what is needed to render the panel on search results.
 
 ### Single image
 
@@ -33,6 +37,7 @@ This is intentionally minimal at this point as we decide what fields an image sh
 *Query params*
 
 - `include` - optional list of fields, as with works endpoint
+  - `visuallySimilar` - includes a list of visually similar works
 
 ### Multiple images
 
@@ -41,8 +46,7 @@ This is intentionally minimal at this point as we decide what fields an image sh
 *Query params*
 
 - `query` - text search query
-- `similarTo` - an image ID; returned images are sorted by visual similarity to this.
-- `include` - as with the single image endpoint
-- `license` - license to filter for
+- `include` - optional list of extra fields _not including `visuallySimilar`_
+- `items.locations.license` - license to filter for
 - `page`
 - `pageSize`
--- a/rfcs/023-images-endpoint/README.md
+++ b/rfcs/023-images-endpoint/README.md
@@ -28,7 +28,7 @@ This is intentionally minimal at this point as we decide what fields an image sh
 
 ### Single image
 
-**`GET` **`/catalogue/v2/images/:image_id`
+**`GET`** `/catalogue/v2/images/:image_id`
 
 *Query params*
 
--- a/rfcs/023-images-endpoint/README.md
+++ b/rfcs/023-images-endpoint/README.md
@@ -19,8 +19,8 @@ The initial images endpoint will be in **beta**, and therefore will be liable to
 ```
 {
   "id": String,
-	"location": DigitalLocation,
-	"parentWork": String
+  "location": DigitalLocation,
+  "parentWork": String
 }
 ```
 
--- a/rfcs/023-images-endpoint/README.md
+++ b/rfcs/023-images-endpoint/README.md
@@ -1,0 +1,48 @@
+# RFC 023: Images endpoint
+
+**Status:** :construction: Draft :building_construction:
+
+**Last updated:** 27/04/2020
+
+## Context and requirements
+
+- Images are now entities in the catalogue API
+- We want to be able to search for images, and to fetch specified images
+- The images endpoint should behave similarly to the works endpoint
+
+## Proposal
+
+The initial images endpoint will be in **beta**, and therefore will be liable to have breaking changes made without a version bump.
+
+### Image display model
+
+```
+{
+  "id": String,
+	"location": DigitalLocation,
+	"parentWork": String
+}
+```
+
+This is intentionally minimal at this point as we decide what fields an image should have.
+
+### Single image
+
+**`GET` **`/catalogue/v2/images/:image_id`
+
+*Query params*
+
+- `include` - optional list of fields, as with works endpoint
+
+### Multiple images
+
+**`GET`** `/catalogue/v2/images`
+
+*Query params*
+
+- `query` - text search query
+- `similarTo` - an image ID; returned images are sorted by visual similarity to this.
+- `include` - as with the single image endpoint
+- `license` - license to filter for
+- `page`
+- `pageSize`


kenoir:
Should there be a single parameter for relating image similarity? i.e. `similarTo` Could we express this is similarity by some metric? 

`similarBy=magicMetric&similarTo=imageId`

Do we intend to ever have multiple measures of similarity?

jamieparkinson:
> Should there be a single parameter for relating image similarity? i.e. `similarTo` Could we express this as similarity by some metric?
> 
> `similarBy=magicMetric&similarTo=imageId`
> 
> edit: The names and possibly the structuring of params like this is maybe not right, but hopefully conveys the idea.
> 
> Do we intend to ever have multiple measures of similarity?

Yes, good point - cc @harrisonpim

rfcs/023-images-endpoint/README.md
```diff
@@ -0,0 +1,48 @@
...
+### Multiple images
+
+**`GET`** `/catalogue/v2/images`
+
+*Query params*
+
+- `query` - text search query
+- `similarTo` - an image ID; returned images are sorted by visual similarity to this.
+- `include` - as with the single image endpoint
+- `license` - license to filter for
```
alexwlchan, line 46: Does the DigitalLocation field include the license? If not, it feels a little odd that we'd filter by this but you'd not see it.

jamieparkinson: It does, yes - I did wonder about extracting it but thought it was a better idea to keep it exactly the same as the works display model. Open to discussion/change though.

harrisonpim:
I'm sure we'll want to make more measures of similarity available, or at least test different implementations of the same feature (eg AB-testing different ways of assessing the similarity of images' colour palettes).
Don't have strong opinions on whether that's specified in one parameter (eg `visuallySimilarTo=imageId`) or two (`similarBy=features&similarTo=imageId`)

harrisonpim:
I think this is a stretch of the RFC's current scope, but eventually I'd like to let users combine those similarity measures/filters eg
"give me images which look like `x` but have a colour palette like `y`"
"give me images which have `blue` in them and look a bit like `x` and `y`"

jamieparkinson:
Thinking about it more, I reckon "similar" as a general concept might be a bit too constrictive and having the 2 dimensions of `similarBy` and `similarTo` may get confusing and complex - so I reckon `visuallySimilarTo=imageId` might be better.

Are you/we happy that "visually similar" would always refer to the _current_ feature vectors technique (ie it wouldn't later get confused with colour palettes or other models for extracting feature vectors)?

harrisonpim:
Yeah, I think that's clear enough (at least to me)... again, might be worth getting @jtweed's opinion on that

gestchild:
Will the image display model eventually contain information to tie it to a specific canvas within a iiif-manifest location of the parent work?

jamieparkinson:
> Will the image display model eventually contain information to tie it to a specific canvas within a iiif-manifest location of the parent work?

I _think_ so but my understanding of IIIF is not great, again one for @jtweed

rfcs/023-images-endpoint/README.md
warrd: In the works API we often use the path of the field to perform filtering, i.e. `location.license=...`. Don't know how consistent we are about that though.

jamieparkinson: Looks like we're very inconsistent 🙃 - the license filter for works is filtering on (any) item's location's license https://github.com/wellcomecollection/catalogue/blob/master/api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchRequestBuilder.scala#L143

rfcs/023-images-endpoint/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+The initial images endpoint will be in **beta**, and therefore will be liable to have breaking changes made without a version bump.
+
+### Image display model
+
+```
+{
+  "id": String,
+  "location": DigitalLocation,
+  "parentWork": String
```
jamesgorrie, line 23: nit: language

Similar to the language [being discussed here](https://github.com/wellcomecollection/storage-service/pull/486#discussion_r416408216), should this just be a `workLink` or similar, and in this case the way of expressing this link is a string.

This implies that the work has to exist for the image to exist, which, while is practically true as these images are generated from works, is that what we're trying to imply?

rfcs/023-images-endpoint/README.md
jamesgorrie: The field we try to use as the parameter is in the display JSON not Elastic path.

[So license as a queryparam](https://github.com/wellcomecollection/catalogue/blob/master/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/WorksParams.scala#L134) should probably have been `items.locations.license=bloop`. 🙃 ❗

rfcs/023-images-endpoint/README.md
jamieparkinson: Good point, I think perhaps we actually _do_ want to imply this? I remember it being discussed before but can't remember much about that...

jamesgorrie: Would we say that the images are children of the work?

Either way - could we type this as a `case class WorkLink(id: String)` as I think this will start to be useful in things like the concepts work?

---

Sideline:
As a point, a nice piece of language spotted through our ontologies [is `partOf`](https://github.com/wellcomecollection/platform/search?q=part+of&unscoped_q=part+of).
Not sure it actually applies here.

jtweed: I don't think this link should be expressed as a string, it should be an object that contain potentially include metadata from the work. The property name should also express the relationship, which I'm not sure should be parent, but still pondering that.

rfcs/023-images-endpoint/README.md
jtweed: Indeed - if we're including the license here (and I think we should and support filtering by license), then we should follow the pattern from works:

- license goes on the location
- the query parameter is `locations.license`

rfcs/023-images-endpoint/README.md
```diff
@@ -0,0 +1,48 @@
+# RFC 023: Images endpoint
+
+**Status:** :construction: Draft :building_construction:
+
+**Last updated:** 27/04/2020
+
+## Context and requirements
+
+- Images are now entities in the catalogue API
```
jtweed, line 9: Might be worth being more specific about what "images" are in this context, to be clear that this is not about digitised content a la IIIF. That they are more conceptual, like works, and consist of images from the visual collections and interesting images from inside books and other materials.

rfcs/023-images-endpoint/README.md
```diff
@@ -0,0 +1,48 @@
...
+- We want to be able to search for images, and to fetch specified images
+- The images endpoint should behave similarly to the works endpoint
```
jtweed, line 11: Language nit pick, I'd say should be consistent with, not similar

rfcs/023-images-endpoint/README.md
jamieparkinson: The discussion above is ambiguous - to clarify, license does go on location, but for the works endpoint the query parameter is just `license`, although as @jamesgorrie says it should've been `items.locations.license`.

What do you think we should go for here?

rfcs/023-images-endpoint/README.md
```diff
@@ -0,0 +1,48 @@
...
+The initial images endpoint will be in **beta**, and therefore will be liable to have breaking changes made without a version bump.
+
+### Image display model
+
+```
+{
+  "id": String,
+  "location": DigitalLocation,
+  "parentWork": String
+}
```
jtweed, line 24: I know this is there, but it should be documented here. This should have `"type": "Image"` too.

rfcs/023-images-endpoint/README.md
```diff
@@ -0,0 +1,48 @@
...
+## Proposal
+
+The initial images endpoint will be in **beta**, and therefore will be liable to have breaking changes made without a version bump.
+
+### Image display model
+
+```
+{
+  "id": String,
+  "location": DigitalLocation,
```
jtweed, line 22: I think this should potentially be `locations` and a list. That way we can support eg a IIIF image location and a location that is a deep link into the IIIF presentation API.

jtweed: The locations should also copy across license, credit, access conditions etc I think. You should be able to do the right thing here without having to refer to the works endpoint.

rfcs/023-images-endpoint/README.md
jtweed: I wonder whether something like `sourceWork` is better? It's not really a parent/child relationship. That or `relatedWork` but that feels not strong enough and a bit generic. Doesn't say how it's related.

rfcs/023-images-endpoint/README.md
jamieparkinson: iirc we discussed this before and decided on 1 location per image? I'm wary of this as it's not just an API change but also a model change with huge consequences (eg, you can't give an `Image` that contains multiple images a feature vector).

I think that last time we said that stuff like deeplinks could be in a separate field to avoid this problem (which also wouldn't be a breaking API change to add).

jamieparkinson: > The locations should also copy across license, credit, access conditions etc I think. You should be able to do the right thing here without having to refer to the works endpoint.

Yes, this already happens, `DigitalLocation` is a shorthand for the exact same schema as in the works API.

rfcs/023-images-endpoint/README.md
```diff
@@ -0,0 +1,48 @@
...
+  "parentWork": String
+}
+```
+
+This is intentionally minimal at this point as we decide what fields an image should have.
```
jtweed, line 27: So on the assumption that for search purposes, we're copying all relevant data into a hidden full text field in the index and this is just for display. I'd say it's what is needed to render the panel on search results, because otherwise you'll need to make another API request for that data, which seems suboptimal.

rfcs/023-images-endpoint/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Single image
+
+**`GET`** `/catalogue/v2/images/:image_id`
+
+*Query params*
+
+- `include` - optional list of fields, as with works endpoint
```
jtweed, line 35: What are the available includes? I can think of two:

- visuallySimilar - includes the similar images
- sourceWork - which hydrates the source work - is that useful?

rfcs/023-images-endpoint/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Multiple images
+
+**`GET`** `/catalogue/v2/images`
+
+*Query params*
+
+- `query` - text search query
+- `similarTo` - an image ID; returned images are sorted by visual similarity to this.
```
jtweed, line 44: This shouldn't be a separate query param, it should be an include. For naming I'd vote for `visuallySimilar`.

jtweed:
Two things from the comments above:

- Yes, I think there should be multiple locations, one of which is the IIIF image, one that deep links into the IIIF manifest.
- I think that the similarity/relationships more generally should be entity-centric, not functionality-centric. Ie, you should ask for an image and its visuallySimilar images, not images that are similarTo an id.

rfcs/023-images-endpoint/README.md
jtweed: The path version, whatever the path ends up being for `/images` - we should really fix this on `/works`

rfcs/023-images-endpoint/README.md
jtweed: I don't mean that an image is multiple images, I mean that an image can potentially exist in multiple locations. Difference between an image as something conceptual and some concrete (eg an actual jpg).

rfcs/023-images-endpoint/README.md
jamieparkinson: I am not sure what you mean here / if my guess is right I don't think it's technically feasible:

- do you mean that if you `include` `visuallySimilar` then you get a list of visually similar images alongside each returned image? if so, how many / how do you specify how many?

- @harrisonpim would it be possible to include such a list or would getting the list of visually similar images for each image require 1 query per result?

rfcs/023-images-endpoint/README.md
jamieparkinson: Basically I think multiple locations is good, having that expressed in a flat list is bad.

rfcs/023-images-endpoint/README.md
jamieparkinson: Possibly none at the moment, just thinking about how it will look as it grows

rfcs/023-images-endpoint/README.md
jamieparkinson: > I don't mean that an image is multiple images, I mean that an image can potentially exist in multiple locations.

Right, but you still need to have one of these as canonical for inference etc

rfcs/023-images-endpoint/README.md
jamieparkinson: I like `sourceWork`

rfcs/023-images-endpoint/README.md
jtweed: You are correct that better describing is better, we haven't done that on `/works` though. Though there are more kinds of locations there.

jtweed: You should be able to tell from the locationType what kind of location it is, though you are correct that doesn't describe the relationship.

rfcs/023-images-endpoint/README.md
jtweed: Yes, that's what I meant. It might not be feasible or desirable.

jtweed: You're talking me round to it being a query parameter, I think I was wrong before.

rfcs/023-images-endpoint/README.md
jamieparkinson: I'm saying that there is a concrete problem with having a `locations` list: which one of those does the inferrer fetch to extract feature vectors from?

jtweed: The iiif image, if you want the actual physical image

jamieparkinson: Does that not mean that `locations` a list of 2+ different types of things, with an implicit constraint that there is only 1 thing of 1 of the types? In which case we should make that explicit eg:
```
  "location": blah,
  "presentation": blah
```

jtweed: I'm not sure I want to tie the model to location types so strongly, but would definitely like input from others here esp with respect to the same situation on works.

rfcs/023-images-endpoint/README.md
warrd: If it is an include (meaning when we ask for it, we get a list of visually similarly images on the given work) it might make sense only including it on the single work endpoint? If visually similar images are retrieved dynamically (I'm guessing via the feature vector stuff) intuitively it would be pretty slow for say a list of 10 works each getting 4 visually similar images.

warrd: Btw I hadn't refreshed my browser before I wrote my comment so I missed everything after the initial comment

jamieparkinson: > only including it on the single image endpoint

That's a good suggestion which I am kind of convinced by. The difference is between searching for images (sorted by similarity to a given one) and between asking for a specific image, as well as those most similar to it - and I'm not sure which of those I prefer. Technically there's basically no difference.

What do other people think?

warrd: The suggestion I made of only having visually similar images on the single image endpoint would be a retrieval by ID and a single query. I don't reckon you could do the list of images and visually similar images too in a single query very easily.

warrd: > The difference is between searching for images (sorted by similarity to a given one) and between asking for a specific image, as well as those most similar to it - and I'm not sure which of those I prefer. Technically there's basically no difference.

If you were always going to display a number of visually similar images on the single image endpoint the include makes sense, if it is something you might want todo async in response to some user input (after having received the image itself) then the original query parameter suggestion probably makes more sense

jamieparkinson: > I don't reckon you could do the list of images and visually similar images too in a single query very easily.

Yep agreed. The difference I'm talking about is only in terms of how you make the single-id search, either:

- `/images/[id]?includes=visuallySimilar` or
- `/images/?visuallySimilarTo=[id]`

(obviously they'd have different return types too)

rfcs/023-images-endpoint/README.md
kenoir: > I'm saying that there is a concrete problem with having a `locations` list: which one of those does the inferrer fetch to extract feature vectors from?

Can we not have differing internal & internal models where the canonical location for the inferrer is explicit internally but hidden from the API consumer?

rfcs/023-images-endpoint/README.md
jamieparkinson: I re-read @jtweed's comment below, have been convinced by the former of these options

harrisonpim: `/images/[id]?includes=visuallySimilar` feels more appropriate to me, and fits with @jtweed's "entity-centric" comment below

rfcs/023-images-endpoint/README.md
jamesgorrie: > You are correct that better describing is better, we haven't done that on /works though. Though there are more kinds of locations there.

Personally I would have really preferred more explicit relationships on `works`, as a client and someone writing code in the pipeline.

---

I like @kenoir's suggestion.

A similar example would be thumbnails on `Work`s.

It is just a digital location, but special as we have given it a context or way for it to be used.

Could this just be a case of having `locations` which works the same as in `works`.

We would have the `vectorsFrom: DigitalLocation` as internal as I can't imagine a client finding this useful.

Having a `thumbnail` type param, or better suggested `presentation` or `display` (preferably with a URL template that can be used for the client to create it's own sizes) would also be super useful.

I can't see a problem with there being duplication in the more explicitly related locations and the list.

If we just have the more explicit locations, adding the the more generic `locations` list later should be easy?

rfcs/023-images-endpoint/README.md
jamesgorrie: And all images will come from a work and no other sources?

jamieparkinson: Good point, how about something like
```scala
trait ImageSource {
  val id: Id
}

case class SourceWork(
  id: Id,
  sourceType: String = "Work"
) extends ImageSource

case class DisplayImage(
  ...,
  source: ImageSource
)
```

rfcs/023-images-endpoint/README.md
jtweed: Ok let's look at being explicit then, we actually need to change locations on works as part of the locations and requesting work anyway, so this might lead us towards a pattern that we can reuse there.

rfcs/023-images-endpoint/README.md
jtweed: Though I was actually becoming more convinced by the latter as it's more flexible and lets us decide how many to return. At the expense of requiring another API call to populate that component, but I think that's ok and arguably preferable as we only need visually similar images for those that are clicked on.

rfcs/023-images-endpoint/README.md
jamesgorrie: Looks ✨ 

`SourceWork` will definitely come in handy elsewhere (if we abstract `ImageSource` to `Source` or something).

jamieparkinson:
**Regarding location(s):**

The discussion above was useful but we didn't come out of it with anything concrete, and whilst we're all agreed on the value of being explicit I think it would be unwise to thrash around on that at the expense of delaying the shipping of an images endpoint and ripping apart the current internal model, especially when at the moment we don't actually have multiple locations.

I think that the least-effort, most flexible change is to follow @kenoir's suggestion and expose `locations: List[Location]` in the API, _but to keep the internal model the same as a single location for now_. We can always add more locations in either/both the internal and display models later.

Does anyone object to this?

jamieparkinson:
Alternatively: let's be explicit by renaming `location` -> `displayLocation` (eg) and leaving everything else the same?

rfcs/023-images-endpoint/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Image display model
+
+```
+{
+  "id": String,
+  "location": DigitalLocation,
+  "source": {
+    "id": String,
+    "sourceType": "Work"
```
jtweed, line 25: This should be `type` not `sourceType`.

jtweed:
I think I still favour the more generic list until we've thought through the semantics a bit more and seen how locations on works evolve. But if do want to do it differently, I think `imageLocation` is better than `displayLocation`.

rfcs/023-images-endpoint/README.md
jamieparkinson: Are you sure? I think that might conflict with the ontology types / discriminators in the existing model - isn't the pattern that here we'd have `"type": "Source"`? `sourceType` is more like an enum of possible sources

jamieparkinson:
I'm going to go with the list so that we retain consistency across the endpoints and because I think the "explicit" version needs some more thought to be scalable across entities and situations, **but** I would like to discuss more with @jamesgorrie so that we can come up with a solution that's better all round.

rfcs/023-images-endpoint/README.md
jtweed: That's my point. It should be the ontology type. I don't think we need a `Source` type that just wraps the possible sources. The list of possible sources is other types in our model, ie `Work` mainly but possibly `Exhibition`, `Story`.

PR #30, opened 2020-05-07
RFC 025 defines rules for deriving identifier from filename

tomcrane:
This needs:

1) The case-insensitive requirement for deciding whether the filename starts with a b number could be more precise (case-sensitivity is culture-specific). 
2) Further transformation of the derived ID after the first two rules are applied. Remaining reserved characters or chars not defined in https://tools.ietf.org/html/rfc3986 - what to do with them? Remove, replace with underscore, convert chars with diacritics to closest ascii equivalent...

### DIFF
--- a/rfcs/025-identifiers_from_filenames/README.md
+++ b/rfcs/025-identifiers_from_filenames/README.md
@@ -1,0 +1,67 @@
+# RFC 025: Minting asset identifiers from filenames
+
+**Last updated: 7 May 2020.**
+
+## Background
+
+Whether generated through photography during digitisation, or acquired in digital form, all assets enter workflows with a filename on disk:
+
+* b12345678_0001.jp2
+* bob-field-recording-june-1978.wav
+
+We want to preserve this filename so that it is reflected in the URL of an asset on the web, which will usually be some sort of derivative. A IIIF Image service, or an mp3 created from a master audio file.
+
+The filename is used to generate a meaningful identifier, that ends up in a URL path segment:
+
+```
+https://..../some-prefix/b28047345_0035.jp2/info.json
+                         |--- asset id ---|
+```
+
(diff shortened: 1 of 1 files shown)


kenoir:
I think @jtweed needs to look at this.

jtweed:
For non-alpha characters, fwiw Archivematica uses pretty blunt rules that do at least have the benefit of ensuring ascii file names:

https://www.archivematica.org/en/docs/archivematica-1.12/user-manual/transfer/change-filenames/

tomcrane:
One consideration is that the code that transforms a METS file path into a DLCS identifier needs to produce the same results in the .NET DDS as it does in the Scala API. Rules for removing diacritics and transforming non ascii chars to their nearest equivalent (e.g., `и́` => `ii`, `Č` => `C`) are going to be more complex than "prepend bnumber if not present, replace space with underscores".

What's the most robust way of implementing `mystring.ToAscii()` across platforms?

tomcrane:
(if this RFC ends up with 50 lines of pseudocode it's probably wrong!)

tomcrane:
From the Archivematica link above:

> To prevent these kinds of failures from happening, Archivematica implements a script that changes file and directory names to conform to the requirements of more restrictive tools. Valid characters are defined as the following:

> `-_.()abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789`

> The script tries to replace any character that is not included in the above list with its nearest equivalent - for example, the character Č is replaced by C.

The crucial thing here is _how_ does it "replace any character that is not included in the above list with its nearest equivalent"; how does it decide that Č => C?

These rules are important for DDS/IIIF-builder when it sends assets to DLCS, but they also need to be compatible with this: https://github.com/dlcs/protagonist/issues/340

kenoir:
Closing as outdated.

PR #32, opened 2020-05-29
Add an RFC about our tagging policy

alexwlchan:


rfcs/025-tagging_our_resources.md
```diff
@@ -0,0 +1,49 @@
...
+
+
+
+## Solution
+
+We are going to tag resources in a Terraform configuration like so:
+
+```hcl
+tags = {
+  TerraformConfigurationURL = "https://github.com/wellcomecollection/:repo_name/:configuration_path"
```
jamesgorrie, line 32: Given configuration might mean the service calling a module - what do we mean here?
If I was provisioning one of our modules, would I expect this to the the provisioner, or the module URL?

If it's the module, might `TerraformRepo` or similar be more explicit?

alexwlchan: I'm using "configuration" in the [strict Terraform sense](https://www.terraform.io/docs/configuration/index.html#resources-and-modules):

> A **Terraform configuration** consists of a root module, where evaluation begins, along with a tree of child modules created when one module calls another.

PR #34, opened 2020-06-29
Add ADRs and context

kenoir:
In order that we record and share information about the current state of the platform we should use "Architecture Decision Records (ADRs)".

This PR adds:
- Rationale for ADRs
- Rationale for RFCs
- ADRs for:
  - logging
  - secrets

### DIFF
--- a/adr/README.md
+++ b/adr/README.md
@@ -1,0 +1,9 @@
+# Architecture Decision Records (ADR)
+
+Documents here are intended to record and describe particular architectural decisions. 
+
+---
+
+An ADR records a a particular decision and contains the key information to understand how the mechanism it describes work.
+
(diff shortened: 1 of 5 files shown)


adr/README.md
```diff
@@ -0,0 +1,9 @@
+# Architecture Decision Records (ADR)
+
+Documents here are intended to record and describe particular architectural decisions. 
+
+---
+
+An ADR records a a particular decision and contains the key information to understand how the mechanism it describes work.
+
+**An ADR is not an RFC**. An RFC is speculative and provides a place to discuss multiple options with pros and cons, which may or may not be implemented. An ADR relates only to the current implementation and will be replaced as new decisions are made.
```
jamesgorrie, line 9: This is great!

PR #35, opened 2020-07-06
Relevance reporting service

jamesgorrie:
We need a way of testing and reporting on how us changing the elastic queries will affect the way we expect results to be returned.


kenoir:
Great idea.

Some questions:
- How do we determine which queries to use?
- Can you provide a specific example of what is happening in the lambda, perhaps with pseodo-code?

An observation: when we have a test that can pass/fail search relevance in a gross way, we can use that in multiple scenarios, e.g.  manual "smoke testing" of a staging API before promotion to prod, local testing of new search algorithms, as a "gate" in a CI/CD environment to automate deployments. 

Determining what the test looks like, and how the criteria are set is crucial. I imagine you have plans to include input from library & archive staff when creating the test criteria - making that clear in this RFC would be good.

rfcs/026-relevance_reporting_service/README.md
```diff
@@ -0,0 +1,56 @@
+# RFC 026: Relevance reporting service
+
+**Last updated: 06 July 2020.**
+
+## Glossary
+
+* **elastic-query:** A blob of JSON we send to Elasticsearch
+* **search-terms:** A set of terms sent as a string from a client to a search service
```
harrisonpim, line 8: 👍  for taking some of the load off "query"

rfcs/026-relevance_reporting_service/README.md
```diff
@@ -0,0 +1,56 @@
...
+* Allowing internal teams to have a greater understanding and more input as to how the search works, thus being able to share this with external researchers
+* External researchers interested in how their results are being returned
+
+## Proposal
+
+![architecture](./relevance_service.png)
+
+1. Creating a public API where you can send elastic-queries
+1. This query will be run against a set of predefined search-terms, which will also have a preset set of relevant documents which should match
+1. This will be run against Elastic's [ `rank_eval` ](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-rank-eval.html) api
```
harrisonpim, line 32: do we want to say any more here about how we'll use `rank_eval`? eg the appropriate ways that records might be allowed / required to match the search-terms

rfcs/026-relevance_reporting_service/README.md
```diff
@@ -0,0 +1,56 @@
...
+1. We will report on the response of `rank_eval` response
+
+## Technical implementation
+
+The relevance service will be a lambda that can be triggered by services that have IAM permission to do so.
+
+The service can return a fail / pass as a response to the call, being useful for CI etc, and can store that report in S3 for dashboards.
```
harrisonpim, line 39: could also include some more internal detail about how a failure might be induced, eg 
- the distinction between a soft failure of a single test and a hard failure which is reported to CI etc
- hard failure above a threshold number of soft failures
- hard failure caused by a critical test failing

jamesgorrie:
@kenoir 

> How do we determine which queries to use?

Depends on what you mean - see the glossary, I will try to answer both:

### Search-terms
@alicerichmond is going through some rounds of qualitative research with internal people first, and working with them to go through some searches that they know should return certain results. There are a few documents about which we will distil into this<sup>`*`</sup>.

@taceybadgerbrook is also working on analysing the quantitative search queries to help come up with this. [e.g.](https://github.com/wellcomecollection/catalogue/pull/634/files)

### Elastic-queries
These will be created on the basis that they work with the above research.

<sup>`*`</sup> Documents to distil
* https://docs.google.com/document/d/1aIbE4IAOZb1Cbyp9ei81HtZ5w98AVZHIrudP9RUG1Is/edit#heading=h.byr3zbqes7qg
* https://wellcomecloud-my.sharepoint.com/:w:/g/personal/a_richmond_wellcome_ac_uk/EZ-5bvoQ76NCkIBDok_L0c0BaqsaFKFLN-J9KtnEprUsKw?e=lyzSmO
* https://wellcomecloud-my.sharepoint.com/:w:/g/personal/a_richmond_wellcome_ac_uk/EZ-5bvoQ76NCkIBDok_L0c0BaqsaFKFLN-J9KtnEprUsKw?e=lyzSmO

jamesgorrie:
@kenoir actually adding the answers to the questions to the RFC.

jamesgorrie:
@kenoir added some pseudo code and some notes on how we;re working with collections info et al.

rfcs/026-relevance_reporting_service/README.md
jamesgorrie: Yip, maybe let's run some experiments together.

rfcs/026-relevance_reporting_service/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+@taceybadgerbrook is also working on analysing the quantitative search queries to help come up with this. [e.g.](https://github.com/wellcomecollection/catalogue/pull/634/files)
+
+<sup> `*` </sup> Documents to distil
+
+* https://docs.google.com/document/d/1aIbE4IAOZb1Cbyp9ei81HtZ5w98AVZHIrudP9RUG1Is/edit#heading=h.byr3zbqes7qg
+* https://wellcomecloud-my.sharepoint.com/:w:/g/personal/a_richmond_wellcome_ac_uk/EZ-5bvoQ76NCkIBDok_L0c0BaqsaFKFLN-J9KtnEprUsKw?e=lyzSmO
+* https://app.gitbook.com/@wellcomecollection/s/catalogue/search/intentions-and-expectations
+
+## Technical implementation
```
kenoir, line 47: Just "Implementation"?

rfcs/026-relevance_reporting_service/README.md
kenoir: I wonder if storing the result of the runs in ES might allow the results to be visualised in Kibana?

rfcs/026-relevance_reporting_service/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+The relevance service will be a lambda that can be triggered by services that have IAM permission to do so.
+
+The service can return a fail / pass as a response to the call, being useful for CI etc, and can store that report in S3 for dashboards.
+
+### Pseudo λ
+
+``` js
+const http = require('http');
```
kenoir, line 56: 👍

rfcs/026-relevance_reporting_service/README.md
jamesgorrie:
```suggestion
## Implementation
```

rfcs/026-relevance_reporting_service/README.md
jamesgorrie: As discussed this will come out while doing the work rather than thinking about it.

rfcs/026-relevance_reporting_service/README.md
```diff
@@ -0,0 +1,123 @@
...
+* **elastic-query:** A blob of JSON we send to Elasticsearch
+* **search-terms:** A set of terms sent as a string from a client to a search service
+
+## Summary
+
+A way for us to be able to test and report on the efficacy of elastic-queries by being able to compare a set of search-terms and their respective expected results and ordering.
+
+## Motivation
+
+We need to have and instil transparency and confidence in the way that our search relevancy works.
```
kenoir, line 16:
```suggestion
We need to have confidence and instil transparency in the way that our search relevancy works.
```

rfcs/026-relevance_reporting_service/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+Any test to help with this will need to be against our production data, as that is exactly what people will be seeing.
```
kenoir, line 18:
```suggestion
We should test using the same data people are seeing.
```

rfcs/026-relevance_reporting_service/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+* Developers and data-scientists being able to write new search queries with  confidence in improvement and without fear of regression
+* Allowing internal teams to have a greater understanding and more input as to how the search works, thus being able to share this with external researchers
+* External researchers interested in how their results are being returned
+
+## Proposal
+
+![architecture](./relevance_service.png)
+
+1. Creating a public API where you can send elastic-queries
```
kenoir, line 30: Why should the API be public?

rfcs/026-relevance_reporting_service/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Use cases
+
+* Developers and data-scientists being able to write new search queries with  confidence in improvement and without fear of regression
+* Allowing internal teams to have a greater understanding and more input as to how the search works, thus being able to share this with external researchers
+* External researchers interested in how their results are being returned
+
+## Proposal
+
+![architecture](./relevance_service.png)
```
kenoir, line 28: I find the architecture diagram & the explanation below quite confusing (I found I needed the context of the Elasticsearch rank eval docs to understand better).

Some explanation of the terms used and some examples of what they are ("query", "elastic-query", "search terms") would be useful for me.

rfcs/026-relevance_reporting_service/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+<sup> `*` </sup> Documents to distil
+
+* https://docs.google.com/document/d/1aIbE4IAOZb1Cbyp9ei81HtZ5w98AVZHIrudP9RUG1Is/edit#heading=h.byr3zbqes7qg
+* https://wellcomecloud-my.sharepoint.com/:w:/g/personal/a_richmond_wellcome_ac_uk/EZ-5bvoQ76NCkIBDok_L0c0BaqsaFKFLN-J9KtnEprUsKw?e=lyzSmO
+* https://app.gitbook.com/@wellcomecollection/s/catalogue/search/intentions-and-expectations
+
+## Implementation
+
+The relevance service will be a lambda that can be triggered by services that have IAM permission to do so.
```
kenoir, line 56: Why a lambda? What language are you proposing to write it in?

rfcs/026-relevance_reporting_service/README.md
jamesgorrie:
```suggestion
1. Creating an API where you can send elastic-queries
```

Was more about being accessible, but thinking about it - that's the nature of an API, public or not.

rfcs/026-relevance_reporting_service/README.md
jamesgorrie: There is a glossary at the top of the doc outlining those queries - does that need to be made clearer?

kenoir: For me, yes: "A blob of JSON we send to Elasticsearch" doesn't tell me what it is for.

Here's my understanding: 

1. The service accepts a query with some expected results. 
2. The service compares the results of that query on real data to the expected results. 
3. The services responds with an evaluation of how well the expected results match reality. 

The service is a layer on top of the rank-eval API in Elasticsearch allowing you to record interactions and take actions based on responses.

PR #37, opened 2020-08-03
Relation Embedder

warrd:
https://github.com/wellcomecollection/platform/issues/4672


rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+* The fact that the relations are not indexed up front means we can not use related data within search queries (such as if we wanted to include the title of the root of the collection for any descendent works, for example).
+
+* The use of paths for encoding the structure of the tree couples relations tightly to Calm. There is upcoming use cases for relations beyond just archives, and on a conceptual level the goal of the pipeline should be to generate data which is abstracted away from any particular source system.
+
+We therefore need a method to generate relation data on works within the pipeline, in advance of indexing for the API.
+
+## Enricher
+
+In order to include relation data on the work model itself it is clear there needs to be an additional stage of the pipeline which is able to query relationships between works, and include data from related works onto the work itself. This proposed name for this stage is the enricher.
```
jamieparkinson, line 44: I wonder if the name for this design should indicate that this is a kind of "relational enricher", as otherwise it's not clear that (whilst the task performed is still enrichment) it differs from eg an image enricher that doesn't analyse relationships but instead performs ML inference.

rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+Generally, for the insertion, update or removal of any given node, we need to update its siblings, its parent, and the whole subtree containing its descendants. The worst case for this process is an update of the root node, in which case the whole archive needs to be denormalised, with the largest archives containing in the order of 1000 works.
+
+## Commonalities with Other Pipeline Stages
+
+The matcher and merger stages have similar underlying functionality to the enricher. Abstractly these stages involve:
+
+1) Receiving a work as input and applying domain knowledge to build a graph of connections to other works
+2) Using the input work and its graph to emit a new work at the output, including any required data from the relations
```
jamieparkinson, line 78: The matcher/merger can emit many works for one input work. The matcher outputs the whole graph and the merger *may* merge them, and even those which are merged into a target work are still emitted as redirected works.

rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+This points to a potential universal solution within the pipeline, such as using a common service, library or data store to assist with some of these tasks. Additionally, there are potential similarities with upcoming concepts work, where there has been some investigation into how concepts would fit within a graph model.
+
+There are a few issues to bear in mind when considering how a shared solution would work:
+
+* Whilst stage 1 (currently the matcher) could potentially also involve analysis of archive relations, it is important as stated above that the denormalisation in the enricher takes place on a merged works to ensure data consistency. This indicates the merger and enricher stages being separate and accessing works from two separate works stores (unmerged and merged).
+
+* The matcher currently uses a custom DynamoDB graph implementation which was built specifically for the purpose. This was never meant to be a general solution and a rewrite would likely need to be involved in the building of some common system.
```
jamieparkinson, line 86: Possibly add the silver lining that the dynamo implementation is struggling even now - it wasn't built for dealing with such big trees

rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+* The graph element of both the matcher and relations work are relatively similar, but the denormalisation stages (the enricher and the merger) have less parallels. For example, the input to the merger is currently a group of work IDs to be merged with the stage having no knowledge of the graph representation, unlike the enricher which is likely to receive single works and needs access to the graph.
```
jamieparkinson, line 88: Merger input is a set of a set of IDs - iirc this is necessary for the _un_-merging of works. Good point though.

rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
...
+    │
+(relates to)
+    │
+    ├── [ Concept A ]
+    └── [ Concept B ]
+```
+
+Here square brackets indicate entities and parentheses indicate (directed) relationships.
+
+Having a single graph representation would enable richer queries and analytics of our data, and would likely be especially useful for machine learning work. However this may introduce complexity by breaking assumptions about data flow in the pipeline, such as if there were multiple stages which had the ability to write to the graph. This should therefore should be considered with caution. Whether as part of a single graph representation or a more specialised store solely for inter work relationships in the context of archives, using such a database to store relationships would aid implementation of the denormalisation stage, especially with the traversal and emitting of related works.
```
jamieparkinson, line 113: I'm interested in how much can be done with pure graph DB queries. The `query` -> `process results` -> `update graph` flow is so natural for graph DBs that I _think_ most of them can do this kind of work within the query language, which obviously has massive advantages not least in terms of consistency.

[This](https://neo4j.com/docs/cypher-manual/current/introduction/quering-updating-administering/) is one thing I could find, I'd be interested to see if we could write a pseudo-query doing the sort of thing we're going to want.

rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
...
+    // Path query implementation
+  }
+}
+```
+
+This has the advantage of requiring more gradual changes to the existing pipeline architecture rather than introducing lots of new services and logic all at once, and allows us to start writing code straight away, giving us more time to consider how a graph implementation would work with both the current matcher / merger and upcoming concepts work.
+
+## Potential problems
+
+The extra pipeline stage discussed introduces more work to be performed in the pipeline, with at least one additional service and extra Elasticsearch index. This could potentially be costly, especially due to the fact that updates to one work have implications on others. It may make sense in the longer term to introduce more complexity into the pipeline regarding knowing when an update is required (this could take the form of knowing what fields are included in related works for example, and only emitting related updates when one of these changes).
```
jamieparkinson, line 160: +1 to smarter indexing

rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+Rather than storing the full work within the graph database, if only the IDs and relationships are stored this allows us more flexibility about where the graph analysis is performed in the pipeline and prevent data inconsistency errors in cases where the analysis was performed on pre-merged works. The current matcher database also stores solely the IDs, with the merger using these IDs to fetch the works themselves from the recorder VHS.
+
+There are a number of graph and non graph stores we could use for storing relationships:
+
+* [Neo4j](https://neo4j.com) is the most well known graph store and probably the most advanced. The version 4 release from this year introduces sharding and thus provides horizontal scalability. We could run this hosted with [Neo4j Aura](https://neo4j.com/aura) or deploy to AWS ourselves. Scala supports seems OK, with a [DSL library](https://github.com/manishkkatoch/scala-cypher-dsl) which seems semi-official but not exactly widely used.
+
+* [AWS Neptune](https://aws.amazon.com/neptune) is a fully managed graph database solution from Amazon. Querying is done via Gremlin or SPARQL.
```
jamieparkinson, line 121: I think I'm inclined towards Neptune just because of how much of a pain DB management can be... Neo4j docs have some quite opaque stuff that implies scalability is a feature of the Enterprise Edition (🙄 ) only - is that true?

rfcs/026-work-enricher/README.md
harrisonpim: fwiw this distinction gets a bit trickier with the concepts enricher where we're planning to do both

rfcs/026-work-enricher/README.md
jamieparkinson: weco-bot.... have you become sentient?

jamieparkinson: but I agree about Aura anyway

rfcs/026-work-enricher/README.md
jamieparkinson: Perhaps we don't have to include the name in the services then - it would just be good to have one term for this design which uniquely identifies it eg "the works enricher is a relational enricher", "the concepts enricher is a relational and inference enricher"

rfcs/026-work-enricher/README.md
kenoir: I think that this implementation will make it necessary to split reindexing into multiple steps - we will need all merged works in a store to proceed with relational enrichment as nodes in the graph may be missing unless the reindex up to that point is complete.

rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
...
+}
+
+class ElasticsearchRelatedWorksService(es: ElasticClient) {
+  def retrieveRelatedWorks(work: IdentifiedBaseWork): Future[RelatedWorks] = {
+    // Path query implementation
+  }
+}
+```
+
+This has the advantage of requiring more gradual changes to the existing pipeline architecture rather than introducing lots of new services and logic all at once, and allows us to start writing code straight away, giving us more time to consider how a graph implementation would work with both the current matcher / merger and upcoming concepts work.
```
kenoir, line 156: +1 to this.

rfcs/026-work-enricher/README.md
warrd: The enricher stage would only have read access to the graph store, with messages at its output being the way it affects the outside world, so don't think these sort of queries are going to be useful.

rfcs/026-work-enricher/README.md
jamieparkinson: Ignoring performance and (in)efficiency, I don't think that's true - as with the matcher/merger, the graph will be incrementally updated as each related work comes through.

rfcs/026-work-enricher/README.md
jamieparkinson: But if we have a graph store we will have something that writes to it - which will need to do something like this (like the matcher does now), no?

rfcs/026-work-enricher/README.md
warrd: I think I agree with @jamieparkinson here, if you are suggesting in some way that the enriching blocks on any update until all messages earlier in the pipeline have been processed (which sounds like a load of extra complexity). The enriching will eventually be correct by the time all messages are processed, even if the graph is incomplete at first.

rfcs/026-work-enricher/README.md
warrd: Yes true. It is just adding single connections and does not need to do full graph traversal, but something like that could still be useful.

rfcs/026-work-enricher/README.md
harrisonpim: general point about db management is fair, but afaict scalability is one of aura's headline features - would be very surprised if it was an add-on

kenoir:
Noting that Neo4J can be queried using Gremlin also, and that it might be useful to ensure portability of some code between Neptune/Neo4J. 

Not forgetting the suggestion in this RFC that we use Elasticsearch as an intermediary letting us take smaller steps towards a solution.

rfcs/026-work-enricher/README.md
jamieparkinson: @harrisonpim assuming you mean neo4j rather than aura, I agree but got that idea from [here](https://neo4j.com/business-edge/whats-in-neo4j-enterprise-edition/)

> Scaling Neo4j horizontally over multiple machines is one of the greatest benefits of using the Enterprise Edition

Perhaps that's what v4 brings to the community edition?

kenoir:
We should understands how this will impact https://github.com/wellcomecollection/docs/pull/29 and take input from @jamesgorrie

rfcs/026-work-enricher/README.md
harrisonpim: i was referring to aura as a one of [three neo4j editions](https://neo4j.com/subscriptions/) (community, enterprise, aura). from what i can see, enterprise and aura both have scaling as a prominent feature

jamieparkinson: Ah right OK - so I think that mean that we couldn't run horizontally scalable Neo4j ourselves

alicefuzier: FWIW, as part of the original investigation for the matcher merger way back when I di look into Neptune or whatever it was called back then. Gremlin is horrible and everyone knows my opinions about SPARQL, so I would lean towards neo4j moslty because it's well established as a graph database and the language makes a lot more sense. Even if Aura is new and it definitely will have teething problems, I would expect them to be less long-lasting than having to deal with a horrible query language forever

rfcs/026-work-enricher/README.md
alicefuzier: Not really important but I think just enricher is a bit vague in that almost every step of the pipeline is enriching. In fact, the whole pipeline could be called enricher. I would prefer a name that hints in what way it's enriching

rfcs/026-work-enricher/README.md
alicefuzier: To be honest I don't think the input format is the most important difference between the merger and the proposed enricher. We could modify the merger to receive single works and read the graph but the business logic of merging and the logic of enriching are completely different and probably wouldn't need to be base on field by field rules

rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+* The graph element of both the matcher and relations work are relatively similar, but the denormalisation stages (the enricher and the merger) have less parallels. For example, the input to the merger is currently a group of work IDs to be merged with the stage having no knowledge of the graph representation, unlike the enricher which is likely to receive single works and needs access to the graph.
+
+Besides from the use of graphs and similarities to the matcher and merger, another aspect worth mentioning is the fact that we will need another store for denormalised works in addition to a store for merged works. Having two Elasticsearch indices coupled to corresponding ingestors is a possible solution (one for the current index of merged works and adding an additional one for denormalised works). There have been some previous discussions about replacing the recorder VHS with an Elasticsearch index, so it might be a worthwhile time to consider how this would work in this context, as here we would introducing an Elasticsearch index which was not the sink of the pipeline but used as an intermediate store.
```
alicefuzier, line 90: Not saying that we should address this as part of this RFC but this might also sit within the more general problem of "how do we store changes to works in the pipeline" which we currently solve in one way in the recorder and in another way in big messaging

harrisonpim:
> Noting that Neo4J can be queried using Gremlin also, and that it might be useful to ensure portability of some code between Neptune/Neo4J.

Wrt query languages, this reminded me that one of neo4j's biggest advantages is [APOC](https://neo4j.com/developer/neo4j-apoc/), a library of cypher functions which massively reduce the complexity of queries you need to write to do anything interesting. I found that doing the equivalent work in gremlin or sparql got scratchy very quickly. [It looks like aura also supports APOC](https://aura.support.neo4j.com/hc/en-us/articles/360038743233-Stored-Procedures-in-Neo4j-Aura), which is reassuring.

rfcs/026-work-enricher/README.md
warrd: Yes completely, I think it would be good to get it right for this work so up for having these discussions before proceeding if people think it worthwhile.

harrisonpim: i think this, and how it affects denormalising changes between works/concepts, is something that it would be especially useful to get @jamesgorrie's opinion on. we went back and forth on a lot of options while fleshing out the concepts branch, but a solution which works for the whole pipeline would be amazing

kenoir: IMO we should switch out the VHS for a faster cheaper store in the recorders. The question is when. Do we do it now to make future reindexes faster? Or do we wait until we've met the library switch-off objective? Will making the change now enable us to reach that objective faster (given the dev time to switch the recorder datastore, risk of dealing with smoothing out bugs with the "new" way).

rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
+# RFC 026: Work Enricher
+
+## Background
+
+Recently we have started ingesting archive data from Calm. An archive according to our model is a collection of works arranged hierarchically in a tree structure according to some path. For example, the works with paths `A`, `A/1`, `A/1/X`, `A/1/Y` and `A/2` would for the following tree:
```
alexwlchan, line 5:
```suggestion
Recently we have started ingesting archive data from Calm. An archive according to our model is a collection of works arranged hierarchically in a tree structure according to some path. For example, the works with paths `A`, `A/1`, `A/1/X`, `A/1/Y` and `A/2` would form the following tree:
```

typo?

rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+Our API allows us to display relations for some particular work using a number of fields: 
+
+* `parts` is an array containing the children of the work (i.e. `parts` for work `A` would contain `A/1` and `A/2`)
+
+* `partOf` is an array containing the ancestors of the work (i.e. `partOf` for work `A/1/X` would contain `A` and `A/1`)
+
+* `precededBy` is an array containing the siblings of the work earlier in sort order (i.e. `precededBy` for work `A/2` would contain `A/1`)
+
+* `succeededBy` is an array containing the siblings of the work later in sort order (i.e. `succeededBy` for work `A/1/X` would contain `A/1/Y`)
```
alexwlchan, line 24: Two notes:

* You probably want "e.g." (for example) instead of "i.e." (that is)
* Double-checking I've understood correctly, and maybe expanding the example would benefit: `precededBy` doesn't show the siblings that come later, right? And vice versa for `succeededBy`?

So if the tree was

```
A
 ├── 1
 │   ├── X
 │   ├── Y
 │   └── Z
 │
 └── 2
```

Then the `precededBy` of A/1/Y would be {A/1/X}, and the `succeededBy` would be {A/1/Z}?

rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+* The use of paths for encoding the structure of the tree couples relations tightly to Calm. There is upcoming use cases for relations beyond just archives, and on a conceptual level the goal of the pipeline should be to generate data which is abstracted away from any particular source system.
+
+We therefore need a method to generate relation data on works within the pipeline, in advance of indexing for the API.
+
+## Enricher
+
+In order to include relation data on the work model itself it is clear there needs to be an additional stage of the pipeline which is able to query relationships between works, and include data from related works onto the work itself. This proposed name for this stage is the enricher.
+
+It is important that this stage takes place after the merger: if denormalisation happens before this stage the related data may be inconsistent with what exists on the work post merging. For this reason the enricher should exist near the end of the pipeline immediately preceding the ID minter.
```
alexwlchan, line 46: I'm a bit confused in this paragraph: what is "this stage"? Is that the enricher or the merger?

An example or explanation of how we'd end up with inconsistent data would help: I think it's because `precededBy` is a sorted list, and you can't combine that if you're merging two works together, but maybe I'm missing something.

It has to come after the merger, but is there any requirement for where it goes before/after the ID minter? I think it's position relative to the ID minter is mostly irrelevant (and maybe mention of that should be omitted).

rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+There are a few possible methods by which the enricher can query relationships between works:
+
+1) Assuming the merged works coming into the enricher are stored within some Elasticsearch index, the existing API code for querying on the archive path could be repurposed for this.
```
alexwlchan, line 50: "Assuming the works are in an index" – how?

> Another service could index all the pre-enricher works in an Elasticsearch index. The enricher could then use the existing API code for finding relationships.

rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
...
+```
+A
+├── 1
+│   ├── X
+│   └── Y
+│
+└── 2
+```
+
+Generally, for the insertion, update or removal of any given node, we need to update its siblings, its parent, and the whole subtree containing its descendants. The worst case for this process is an update of the root node, in which case the whole archive needs to be denormalised, with the largest archives containing in the order of 1000 works.
```
alexwlchan, line 71: I can see why insertion or removal triggers a cascade of related works – why is that true for updates? Doesn't it depend on what part of the work was updated? e.g. if a field that's not on the partial works get updated, none of the related works would actually change

rfcs/026-work-enricher/README.md
alexwlchan: "It was never built to be a general solution"

I'm not sure that's true – it was never built to *scale*, but that's a bit different. IIRC, the code was written in the idea that we might eventually reused it.

Having not looked recently; is it the graph database or the locking that doesn't scale?

rfcs/026-work-enricher/README.md
alexwlchan: > Will making the change now enable us to reach that objective faster?

Depends on how many reindexes we expect to do between now and the library switch-off, right? If the index is fairly stable and we're not reindexing very often, it can wait. If we expect to make a lot of changes between now and the switch off, doing this work now might save us time later.

rfcs/026-work-enricher/README.md
warrd: OK good to know, I actually don't know too much about the internals but I think @jamieparkinson is under the impression that it is the locking code that doesn't scale well. Amazon introduced transactions to DynamoDB in November 2018, which might have rendered the custom locking unnecessary?

From my point of view there are a few further issues with the use of DynamoDB for this purpose, they are not necessarily all show stoppers but things to be aware of:

* You can't specify schema / constraints in Dynamo in the same way you could in a dedicated graph database or relational. For example relationships pointing to non existent objects can easily be prevented in Postgres / Neo4J., Neo4J easily allows specifying different types of relationships (so can represent matches and parents easily within the same store), etc

* Graph operations (traversal etc) and querying of relationships will be built in the graph databases instead of requiring custom solutions

rfcs/026-work-enricher/README.md
warrd: I agree we should switch out the recorder VHS only when we have proved the concepts elsewhere and there is more time available for optimising the pipeline. But it makes sense to come up with a solution for this work which we think will be applicable to that too IMO.

rfcs/026-work-enricher/README.md
warrd: Yep exactly :+1:

rfcs/026-work-enricher/README.md
warrd: Assuming the enricher has no knowledge of what happens inside the merger, and the merger can mangle multiple work data in any way it wants (say it takes works with titles `A` and `B` and produces a work with title `A + B`). Say the enricher receives work `C` where one of its children is `A`, then if it is before the merging the work would contain the following data:

```
{
  "parts": [
    {
      "title": "A",
      ...
    },
  ],
  ...
}
```

where what we really want is:

```
{
  "parts": [
    {
      "title": "A + B",
      ...
    },
  ],
  ...

File: rfcs/026-work-enricher/README.md (hunk 14 above)
Comment (warrd) [reply rfcs/026-work-enricher/README.md:71]: Yes that is true, I mention that in the problems section at the bottom

File: rfcs/026-work-enricher/README.md (hunk 3 above)
Comment (jamieparkinson) [reply rfcs/026-work-enricher/README.md:86]: I spent a few days working on the matcher/Dynamo issues a couple of weeks ago - it appears to be mostly the locking that doesn't scale, and in a way that transactions don't help with unfortunately

Comment (jamieparkinson) [reply rfcs/026-work-enricher/README.md:86]: actually to elaborate on this, I don't think the matcher graph store would scale to this sort of thing. It is a special-purpose graph store designed for storing a large number of small graphs which are not connected to each other, and so traversing/querying/updating larger graphs would be very unwieldy

Comment (alicefuzier) [reply rfcs/026-work-enricher/README.md:86]: yes, that's true. It wasn't built with the idea that it couldn't be replaced but definitely the model was a lot of small disconnected graphs. It would not scale for this problem

File: rfcs/026-work-enricher/README.md (hunk 4 above)
Comment (jamieparkinson) [inline rfcs/026-work-enricher/README.md:88]:
```suggestion
* The graph element of both the matcher and relations work are relatively similar, but the denormalisation stages (the enricher and the merger) have fewer parallels. For example, the input to the merger is currently a group of work IDs to be merged with the stage having no knowledge of the graph representation, unlike the enricher which is likely to receive single works and needs access to the graph.
```

rfcs/026-work-enricher/README.md
jamesgorrie: Just a thought experiment here, what if we have 1 works index that has multiple services writing to it?

Some concerns:
* debugging
* re-running from different steps in the pipeline
* ?...

1indextorulethemall(1)

rfcs/026-work-enricher/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+Having a single graph representation would enable richer queries and analytics of our data, and would likely be especially useful for machine learning work. However this may introduce complexity by breaking assumptions about data flow in the pipeline, such as if there were multiple stages which had the ability to write to the graph. This should therefore should be considered with caution. Whether as part of a single graph representation or a more specialised store solely for inter work relationships in the context of archives, using such a database to store relationships would aid implementation of the denormalisation stage, especially with the traversal and emitting of related works.
+
+Rather than storing the full work within the graph database, if only the IDs and relationships are stored this allows us more flexibility about where the graph analysis is performed in the pipeline and prevent data inconsistency errors in cases where the analysis was performed on pre-merged works. The current matcher database also stores solely the IDs, with the merger using these IDs to fetch the works themselves from the recorder VHS.
```
jamesgorrie, line 115: Where would we store the work data then for retrieval?

e.g. I get a graph back, but need to send over the list of updated works, would I then look those IDs up from an index somewhere?

Would this mean extending `BigMessaging` to not only lookup from VHS< but potentially another index e.g. Elasticsearch?

rfcs/026-work-enricher/README.md
jamesgorrie: The more I look at this ☝️ and race conditions and a whole bunch of other things, it feels like a bad idea.

kenoir:
@jamesgorrie 

> The more I look at this ☝️ and race conditions and a whole bunch of other things, it feels like a bad idea.

Can you be more specific about what you are referencing and what concerns you have in more detail?

rfcs/026-work-enricher/README.md
kenoir: > Would this mean extending BigMessaging to not only lookup from VHS< but potentially another index e.g. Elasticsearch?

I do not understand why we would do that? I think we should work to remove the need for `BigMessaging` entirely.

warrd: I think conflating things here, big messaging does not currently look up VHS (looks up S3). Definitely would not need to look up elasticsearch.

The application code would need to look up Elasticsearch to fetch the work data after receiving IDs from the graph store, as mentioned this is what we already do with the existing graph store implementation, where we look up the recorder VHS within the merger.

jamesgorrie: My question is if we're passing IDs about and will need the whole works at stages in the pipeline, where does that lookup happen? It's not clear to me in this step.

jamesgorrie:
@kenoir the review thing took my comment out of context, so I just [updated the comment in place](https://github.com/wellcomecollection/docs/pull/37/files#r465758785).

rfcs/026-work-enricher/README.md
warrd: I don't think within this RFC I have suggested passing IDs around between services (which I don't think is a bad idea, just that not what this RFC focused on). The piece of text this comment refers to is about the graph store only containing IDs, but having a ES index to look up related works. When you look up the graph store in the enricher you will also need to look up the ES index, but that is within a single service (it will still receive works and emit works as messages as far as this RFC concerned).

rfcs/026-work-enricher/README.md
jamesgorrie: @jamieparkinson theoretically how do you think this applies to images?

Should they be being created at this stage, or after the merging has happened?

rfcs/026-work-enricher/README.md
jamesgorrie: Super sorry if my question was unclear - but this answers it:

> the graph store only containing IDs, but having a ES index to look up related works

I do like the idea of passing around IDs in messaging, and having the receiving service do a lookup.

rfcs/026-work-enricher/README.md
harrisonpim: We're deep in the semantics here and I don't want to distract from the actual work at hand, but I agree with @jamesgorrie.

I see enrichers as distinct from other pipeline services in that works/images/concepts/whatever should come out of an enricher with _new_ data which doesn't appear anywhere in the original sources.

That might be because we're fetching information from third parties (eg the concepts enricher), or because we're doing some internal analysis (eg extracting image features and colour palettes), but generally, an enrichment step should be something that adds new information.

jamesgorrie:
Just a thought that I had overnight - and sorry if this has been discussed above.

Are we at risk of generalising what an enricher (name TBD) does? Is it not just a service that is part of the pipeline that receives a work and gives back a work(s) with extra data?

rfcs/026-work-enricher/README.md
warrd: The description of what enriching from @harrisonpim makes complete sense to me. But that makes me question whether this is really doing any enriching :thinking:, its more copying existing data from elsewhere. "Denormaliser" doesn't really have the same ring to it though.

harrisonpim: ☝️ fits with @jamesgorrie's comment below about the risk of over-generalising what an enricher is. I don't see anything wrong with calling this a 'denormaliser' tbh, especially if its functionality is going to appear in a few places across the pipeline.

rfcs/026-work-enricher/README.md
alicefuzier: Why would we want to update just one index though? I don't think I understand what problem would we solve this way... Or maybe I don't understand the proposal

jamesgorrie: What I suggested was definitely a bad idea, thinking aloud.

PR #38, opened 2020-08-06
Pipeline Intermediate Storage

warrd:


rfcs/027-pipeline-intermediate-storage/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+The main issue with this approach is that there is potentially less delineation of concerns, with stages of the pipeline performing work additional to solely data transformation. Any failures during the storage would result in the whole incoming message being marked as a failure, with any previous work on transforming the data having to be redone. Having a separate service solely for storage thus means the point of failure for a particular message making its way through the pipeline is more granular.
+
+However, it is worth mentioning that the waters are muddied somewhat due to the fact that the ability to receive whole works as input to some service requires the use of `big_messaging`. In the case of works over 256KB these works will be stored twice, once in S3 and subsequently in the intermediate store itself. Beyond just reducing costs related to this repeated storage, it may make the whole architecture more easily understandable and improve visibility if we just used intermediate stores and the passing of keys between services, rather than requiring both systems.
```
kenoir, line 147: >if we just used intermediate stores and the passing of keys between services, rather than requiring both systems.
👍

rfcs/027-pipeline-intermediate-storage/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+* Can access via AWS network address to reduce NAT gateway costs (see [here](https://www.elastic.co/guide/en/cloud/current/ec-getting-started-private.html)).
+
+* Back of the envelope calculations show likely to be much cheaper than current solution.
+
+* Horizontal scalability means quite capable of supporting data at sizes much beyond what we have now.
+
+Cons:
+
+* Requires use of external cloud platform or maintenance of our own Elasticsearch cluster, rather than being a service within AWS.
```
kenoir, line 65: I wonder if we would consider the AWS Elasticsearch service to mitigate this.

warrd: Oh yeh I did know they offered elasticsearch service but slipped my mind.. Wondering whether it would still be easy to set up with Kibana etc? What were the original reasons for going with Elastic Cloud rather than this?

rfcs/027-pipeline-intermediate-storage/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+* Terraform does not control the creation and removal of Elasticsearch indices so might require manual clean up after removing a pipeline.
+
+* No transactions which might prove useful for certain use cases.
+
+Questions:
+
+* Elasticsearch supports [versioning](https://www.elastic.co/blog/elasticsearch-versioning-support), including the ability to index with our own external version numbers. However when using external versioning it only indexes data when the version number is greater than currently stored, whereas our business logic currently expects updates to be processed when versions are greater or equal. Some thought will need to be given whether this might work for us, and if not we would need to implement our own system on top of the Elasticsearch APIs (which may not be particularly resilient to race conditions due to lack of transactions).
```
jamesgorrie, line 73: You can use [`external_gte`](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#index-version-types), which we use in a few places.

warrd: Oooo exactly what I was looking for but didn't find.

rfcs/027-pipeline-intermediate-storage/README.md
```diff
@@ -0,0 +1,147 @@
...
+3. Happily stores data in our current range of around 8GB, and scaling beyond this for future proofing.
+
+4. Some form of disk persistence so that we do not need to perform a reindex from scratch in the case of the database going down.
+
+5. Cheaper overall cost than currently, including read / write requests, associated networks costs, and storage costs.
+
+6. Ideally improved performance.
+
+7. Ideally better visibility into the data.
+
```
jamesgorrie, line 44: As we talk about consistency and ACID compliance, what do we see as the requirements in this regard for this store?

rfcs/027-pipeline-intermediate-storage/README.md
kenoir, line 73: There are read/write consistency guarantees (is this what you mean by transactions?) that may be sufficient for our purposes.

rfcs/027-pipeline-intermediate-storage/README.md
```diff
@@ -0,0 +1,147 @@
...
+### Service
+
+The current `recorder` service is a wrapper round VHS which stores items received at its input, emitting a message contained the ID of the stored item for downstream pipeline stages to use themselves for access in future.
+
+ We could mirror this design using a different store to VHS for upcoming intermediate storage needs. As the behaviour would be essentially the same regardless of what state the works were in (unmerged, merged, enriched etc) the exact same service can be inserted in different parts of the pipeline differing only in configuration of what store the service has access to.
+
+### Module
+
+An alternative to using a separate service would be to have the services performing the data transformation writing to the store directly themselves, such as through a common module (`scala-storage` may already serve this purpose sufficiently). This would clearly result in less infrastructure, and so would reduce costs and possibly maintenance burden.
```
jamesgorrie, line 143: I'm a little lost on this - are we saying above we have 1 intermediate store per service, and this is a shared store?

rfcs/027-pipeline-intermediate-storage/README.md
warrd: By transactions I was referring here to a sequence of queries that can be rolled back upon failure, more the atomicity part of ACID than the consistency.

rfcs/027-pipeline-intermediate-storage/README.md
warrd: Under this design not every service would need an intermediate store, but every service making a change to a work (so far that would be a single transformered work store, a merged work store, and a minted work store).

warrd: At the moment this already exists for the transformed works (`recorder` store), but not for the others (unless we consider the final index the minted work store)

rfcs/027-pipeline-intermediate-storage/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Implementation
+
+Orthogonal to the choice of store, the way we write data into the store for usage later in the pipeline has two main possibilities.
+
+### Service
+
+The current `recorder` service is a wrapper round VHS which stores items received at its input, emitting a message contained the ID of the stored item for downstream pipeline stages to use themselves for access in future.
+
+ We could mirror this design using a different store to VHS for upcoming intermediate storage needs. As the behaviour would be essentially the same regardless of what state the works were in (unmerged, merged, enriched etc) the exact same service can be inserted in different parts of the pipeline differing only in configuration of what store the service has access to.
```
jamesgorrie, line 139: Another example of this would be the `ingestor`s?

rfcs/027-pipeline-intermediate-storage/README.md
warrd: Should probably elaborate on this. The proper meaning of a transaction is a single unit of work performed on the database which is ACID compliant. What I (perhaps incorrectly) was using the word transaction to mean was short hand for a group of operations that all form a single unit of work. This is especially useful when making changes to multiple tables concurrently in a relational database etc.

Elasticsearch does not provide this capability, which I thought perhaps would be useful if we needed to implement our own versioning manually on top of Elasticsearch (such as by making a query to check the version, then making a subsequent insert request). I'm not sure grouped transactions would actually be useful in this case, I haven't given it enough thought, but as @jamesgorrie points out there is a way to get Elasticsearch versioning to behave how we want so it's not actually relevant to this particular use case.

rfcs/027-pipeline-intermediate-storage/README.md
warrd: Yes but I think the ingestors do more than what we require for intermediate storage (creating mappings etc), so might not be helpful to mix this up together.

rfcs/027-pipeline-intermediate-storage/README.md
warrd: Well I think we could work around not having any of these properties (it may be desirable in some cases as its all to do with trade offs of course, see e.g. [CAP theorem](https://en.wikipedia.org/wiki/CAP_theorem)). Elasticsearch is eventually consistent rather than atomically consistent for example, Redis not completely durable depending on persistence model used etc

rfcs/027-pipeline-intermediate-storage/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Requirements for Intermediate Store
+
+Taking into account the business logic of our services, and given the issues with the current VHS highlighted in [Problem](#problem), there are a number of requirements and desirable features for any alternative store:
+
+1. The ability to insert items into the store with some unique string ID and some specific integer version (set externally to the pipeline from our adapters), with a method to only store data when the version given is greater or equal to what is currently in the store.
+
+2. The ability to retrieve data from the store, where the latest data is always retrieved and the version is returned with the data.
+
+3. Happily stores data in our current range of around 8GB, and scaling beyond this for future proofing.
```
alexwlchan, line 35: Is 8GB a total, or for a single item?

rfcs/027-pipeline-intermediate-storage/README.md
alexwlchan: AIUI, we use hosted Elasticsearch because we don't want to deal with running our own cluster – an ELK stack is a pain to maintain and gets fiddly in all sorts of ways. Paying Elastic means they deal with weird cluster issues for us.

Might be fine for internal stuff; these clusters would be more transient and if we had any issues, we could delete the cluster and restart.

rfcs/027-pipeline-intermediate-storage/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+Pros:
+
+* In memory so likely to be fastest reads / writes out of all the options.
+
+* AWS offers it via ElastiCache.
+
+* Improved visibility over VHS.
+
+Cons:
```
alexwlchan, line 87: Big unmentioned con is that we don't use Redis anywhere else, so we'd have to write Scala implementations for it from scratch.

rfcs/027-pipeline-intermediate-storage/README.md
warrd: 8GB is the size of the current works index (I dont know if this includes just the raw data, probably includes the internal indexes ES maintains too for performing search)

alexwlchan: [nod] I suspect the bigger constraint if the size of a single item – DynamoDB can happily swallow 8GB of data, it's the 400KB per-item limit that's more of an issue.

warrd: Yep definitely should have mentioned that, will add it

rfcs/027-pipeline-intermediate-storage/README.md
jamesgorrie: I'd agree that eventually consistant is fine for this. We have a 10 loose guideline as to when updates will be seen in the pipeline. Maybe worth us having a performance budget in place to work to in general.

This is pretty interesting https://research.google/pubs/pub39966/

rfcs/027-pipeline-intermediate-storage/README.md
alicefuzier: I think we already use external_gte in the ingestor

rfcs/027-pipeline-intermediate-storage/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+* Replicas can help with scalability when large number of reads.
+
+* Transaction support might be useful for some of our current logic or future use cases.
+
+Cons:
+
+* Performance may not match other options due to stronger consistency guarantees (though likely to be much better than VHS due to only hitting a single system and making use of a cache).
+
+* Writing raw SQL is not fun, and I have not really been satisfied with any of the Scala wrappers I have tried.
```
alicefuzier, line 121: A big +1 to this

rfcs/027-pipeline-intermediate-storage/README.md
alicefuzier: mmmh I think the `ingestor_common` library wants you to define an `IndexConfig `for your ingestor, but you can create an `IndexConfig` that is jus dynamic mapping. It might be simpler to use the ingestor for new storage needs rather than the recorder..

rfcs/027-pipeline-intermediate-storage/README.md
alicefuzier: well, it does exist for the others too IMO, but it's hidden in bigmessaging

rfcs/027-pipeline-intermediate-storage/README.md
jamieparkinson: I'm not sure about this idea - if we don't always have intermediate stores between services then I think it will get complicated: a shared database across many microservices is generally regarded as a Bad Idea.

Another way of saying this: not sure that bigmessaging and intermediate storage are the same problem / are solved by the same solution. Intermediate storage is required for presenting a consistent view of data to multiple services, whereas bigmessaging is about pushing large bits of transformed data between services.

rfcs/027-pipeline-intermediate-storage/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+* Improved visibility over VHS, and ability to query data.
+
+* Replicas can help with scalability when large number of reads.
+
+* Transaction support might be useful for some of our current logic or future use cases.
+
+Cons:
+
+* Performance may not match other options due to stronger consistency guarantees (though likely to be much better than VHS due to only hitting a single system and making use of a cache).
```
jamieparkinson, line 119: I expect given ID minter experience that there might also be throughput issues (ie all the general problems people have with scaling a RDBMS)

rfcs/027-pipeline-intermediate-storage/README.md
jamieparkinson: +1 to self-managed transient stores

rfcs/027-pipeline-intermediate-storage/README.md
warrd: The anti pattern IMO is when there is multiple services writing into the same database, I do not think having single write with one or two places reading out (like we do with `recorder` and `matcher` / `merger`) is going to be problematic. I am not suggesting one single store, but a few stores at different points in the pipeline (e.g. the relation embedder store does not know anything about any other stores apart from the merged works store, so still encapsulation in that sense).

`big_messaging` and intermediate storage shouldn't be the same problem I agree, but due to limitations in SQS it complicates things. Worth mentioning also that SQS could also be viewed as a form of intermediate storage (a queue, rather than key / value store).

warrd: Also worth mentioning that we already pass keys into databases between services, this is not something new. The choice isn't really between big bits of data or passing keys or, it's between passing big bits of data and then keys, or just passing the keys.

rfcs/027-pipeline-intermediate-storage/README.md
warrd: I know there is no desire to use Postgres / RDS, and I don't personally think we should for this purpose, but just to stick up for them: I don't know about the ID minter in particular but in the postgres through bulk inserts you can write 100K+ rows per second, and have no doubt we could make it much faster that DynamoDB + S3 (where each item is written individually with a call to two separate services)

rfcs/027-pipeline-intermediate-storage/README.md
jamieparkinson: Sorry, I was only referring to the idea of bigmessaging using an intermediate store like this - I agree that there is no problem with the proposed pattern of single write / few read stores, but I don't think we should be extending that to bigmessaging.

warrd: Ah ok, makes sense :+1:

jamesgorrie: I still _think_ BigMessaging could be made redundant to simplify the pipeline with the intermediary storage solution, if we were happy that every step in the pipeline emitted IDs and proceeding step did a lookup.

I am pretty certain we would hit bottlenecks while reindexing, which is something to think about irrespective of ☝️ - if these stores are going to written to what is the throughput we need to support, although the ingestors are a good example and _seem_ to deal with the load.

rfcs/027-pipeline-intermediate-storage/README.md
jamesgorrie: > I think we already use external_gte in the ingestor

[Yarp](https://github.com/wellcomecollection/catalogue/blob/master/pipeline/ingestor/ingestor_common/src/main/scala/uk/ac/wellcome/platform/ingestor/common/Indexer.scala#L28)

rfcs/027-pipeline-intermediate-storage/README.md
warrd: Yes I agree big messaging should be removed where not necessary. I don't see how we would introduce new bottlenecks into the pipeline that don't exist already, assuming any new store we use is at least as fast as VHS. The relation embedder could introduce a new bottleneck but that will more be to do with the requirements of that service rather than using intermediate storage vs big messaging.

rfcs/027-pipeline-intermediate-storage/README.md
kenoir: From Slack: 
> On reflection introducing AWS Elasticsearch may introduce a whole new bunch of failure modes we are unfamiliar with.

I am for starting with Elasticcloud ES and then considering other options if we run into issues.

PR #39, opened 2020-09-04
Work state modelling RFC

warrd:


rfcs/029-work_state_modelling/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+4) The type hierarchy from `BaseWork` downwards is quite complex and it can be hard to understand the meaning of particular types of work within the code. The naming could also do with changes in places (e.g. `TransformedBaseWork` does not make much sense as a name, given that the concept of a work does not exist before the transformer).
+
+## Proposed Solution
+
+This RFC proposes having a cleaner separation of the pipeline state of a work with the sort of work it is, namely with different forms of polymorphism:
+
+1) Sub-typing is used for differentiating different sorts of work, with a `Work` being a sum type ADT consisting of the finite number of possible case classes. Note this proposed sub-typing is much simpler than the current implementation, having only a single level of depth and a single parent type.
```
kenoir, line 27: This is a succinct and clear explanation of how we should represent works 👏 .

rfcs/029-work_state_modelling/README.md
```diff
@@ -0,0 +1,177 @@
...
+  title: Option[String] = None,
+  genres: List[Genre[State#Id]] = Nil,
+  images: List[Image[State#ImageId, State#Id]] = Nil,
+  ...
+)
+
+/** Relations contain the related works for some particular work, and only exist
+  * at the Denormalised and Identified stages of the pipeline.
+  */
+case class Relations[State <: WorkState](
```
kenoir, line 163: How will we represent different kinds of relation (i.e. relationships within a collection may not be the only kind of relation we wish to represent).

rfcs/029-work_state_modelling/README.md
```diff
@@ -0,0 +1,177 @@
...
+  parts: Option[List[Work[State]]],
+  partOf: Option[List[Work[State]]],
+  precededBy: Option[List[Work[State]]],
+  succeededBy: Option[List[Work[State]]]
+)
+```
+
+## Potential Issues
+
+* There may be some problems with the implementation of this with regards to the `id_minter` pipeline stage, which works on raw JSON data rather than Scala types. For example, in the `WorkState.Denormalised` case class above `sourceIdentifier` has been included again (it already exists in the work itself) as the `id_minter` will need to generate the `canonicalId` here for the output to be correct. Personally I think the `id_minter` could be implemented pretty simply on our types rather than raw JSON, which would make things a bit safer any avoid some of these idiosyncrasies, although a disadvantage of this is it would require a separate `id_minter` for `Work`, `Image` and any future types we wanted to attach IDs to.
```
kenoir, line 173: It might be time to take the leap with the `id_minter` and start dealing in types - the idea was to allow it to deal easily with new "things" that need identifiers but that has turned out not to be the case.

kenoir:
Impressive write up 👍

rfcs/029-work_state_modelling/README.md
warrd: As far as I understand, this is intended for different sorts of relations beyond just collections, with @jtweed designing it with other use cases in mind. All relations which form a directed acyclic graph of works can be represented this way: we will need to make sure we prevent cycles, which I think should be fine in practice with collections and the other individual use cases being of tree form.

rfcs/029-work_state_modelling/README.md
jamesgorrie: From @alicefuzier in slack
> I disagree on the  line about making the id minter understand types. In fact I would go entirely the orther way and write the id minter in a dynamic language cause doing dynamic stuff (which is what the id minter is doing) in scala is quite painful

rfcs/029-work_state_modelling/README.md
kenoir: How do we describe the type of relation though? Would that be another `ADT`? i.e. 

```scala
sealed trait Relationship {
   parts: Option[List[Work[State]]]
  // etc
}

case class CollectionRelationship(parts: Option[List[Work[State]]]) extends Relationship
```

warrd: Maybe @jtweed can chime in but I think the idea is to make relations agnostic to what sort of thing they are in the source system (so from the perspective of the consumer of the API they do not know / care that it is an archive they are looking at, rather than a hand picked bunch of photos of a certain theme or something else).

Fwiw if I was to encode the relation type in the model I would probably do it more like this than using inheritance:

```scala
case class Relations[State](
  parts: Option[List[Relation[State]]],
  ...
)

case class Relation[State](
  relationType: RelationType,
  work: Work[State]
)

sealed trait RelationType

object RelationType {
  case object Collection extends RelationType
  ...
}
```

kenoir:
```scala
case class Relation[State](
  relationType: RelationType,
  work: Work[State]
)
``` 
👍  

If we don't care about the kind of relationship then we shouldn't encode it. Are we certain we don't care @jtweed ?

rfcs/029-work_state_modelling/README.md
kenoir: Quoting myself from slack:

>  ^ I defer to @ali about the id_minter  she’s seen more examples of where type agnosticism has been useful there.

jamieparkinson: I agree with @alicefuzier about this - not having to keep the ID minter up-to-date with our types is its great advantage imo. Could you explain this a bit more - 

> For example, in the `WorkState.Denormalised` case class above `sourceIdentifier` has been included again (it already exists in the work itself) as the `id_minter` will need to generate the `canonicalId` here for the output to be correct

The ID minter already handles cases where it must mint the same `sourceIdentifier` (to the same `canonicalId`) multiple times in a work - is this case different?

warrd: So there seems to me 3 options:

1. The `id_minter` works on JSON and any models we use are made uniform to fit the schema to what the `id_minter` expects (i.e. always have a `sourceIdentifier` / `otherIdentifiers` at the same level of JSON as the `canonicalId` is generated)

2. The `id_minter` works on JSON but has some knowledge built in about different models and where to generate the `canonicalId`

3. The `id_minter` works on Scala types, with different ID minters being a function of the form:

```scala
def mint[T[_], Unminted, Minted](item: T[Unminted]): T[Minted]
```

I think I prefer 3 as:

**a)** It feels wrong to me changing a model to fit the `id_minter` service as in 1

**b)** Option 2 seems like the worst of both worlds (domain specific knowledge rather than general, but weakly typed)

**c)** It does not seem to me like it would involve much to implement 3 for `Work`, `Image`, `Concept` etc (just a bit of boiler plate), and we get type safety on top of that (being sure that the minter always correctly transitions data from one state to another valid state)

rfcs/029-work_state_modelling/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+4) The type hierarchy from `BaseWork` downwards is quite complex and it can be hard to understand the meaning of particular types of work within the code. The naming could also do with changes in places (e.g. `TransformedBaseWork` does not make much sense as a name, given that the concept of a work does not exist before the transformer).
+
+## Proposed Solution
+
+This RFC proposes having a cleaner separation of the pipeline state of a work with the sort of work it is, namely with different forms of polymorphism:
+
+1) Sub-typing is used for differentiating different sorts of work, with a `Work` being a sum type ADT consisting of the finite number of possible case classes. Note this proposed sub-typing is much simpler than the current implementation, having only a single level of depth and a single parent type.
+
+2) Parameterisation of the `Work` by a state parameter indicates what stage of the pipeline it is in, and which can hold specific data depending on stage. By constraining the parameter to one of a set of known types, we can consider the `Work` model as a finite state machine with the stages in our pipeline each containing a particular transition.
```
jamieparkinson, line 29: Nice application of theory to reality. It will be possible to implement this in such a way that the allowed transitions (and hence the progression of the Work type) are represented in the code, which will be _great_

rfcs/029-work_state_modelling/README.md
alicefuzier: I agree option 2 is the worst of both worlds. I think I prefer option 1 instead of option 3 because i can't think of an easy way to write a type based idminter that we don't have to change every time we add/move/rename a field that needs minting an id. The main advantage of the agnostic id minter to me is that it just works and we hardly had to change it when the model changes. I don't think adjusting the model slightly to keep that beingh the case is that bad. Certainly if there is a way to have a type based id minter where we don't have to explicitly list all the fields that need identifying then that would be the best of both worlds

jamieparkinson (approved):
I like this! Will be interested to see how it looks in practice; I hope there aren't any cases where type erasure (etc) gets in our way

rfcs/029-work_state_modelling/README.md
warrd: If we did do a typed `id_minter` in the future, personally I would prefer the explicit (but slightly more verbose) method of defining different functions for different types, rather than getting clever with cats or whatever and doing advanced type gymnastics to make it generic. And I think it would only ever need a line or two change each time, and any required change would always be caught by the compiler.

----

However one big advantage of option 1 not mentioned is that it doesn't require any new code in the `id_minter`, and as doesn't seem like much desire for change for anyone else, I will try and make that option work.

jamieparkinson: So is the problem that you want `canonicalId`s without adjacent `sourceIdentifiers`* (or vice versa) in some places? I agree with https://github.com/wellcomecollection/docs/pull/39#discussion_r484339055 but more than that I think the modelling of identifiers as being 1-to-1 and predictable on either side of the minter has been hugely successful in how it removes the need to reason very hard about identifier types, and I wouldn't want to lose that.

tl;dr imo the current agnostic approach is not just beneficial for the ID minter but is also beneficial for our brains

_*`otherIdentifiers` is not required or understood by the ID minter_

jamesgorrie: 2¢  - Option `1` seems to have everything the minter needs to do it's job and remain as simple as possible.

warrd: @jamieparkinson I guess the problem is that it forces the `sourceIdentifier` to be moved from the model itself to the `state`, when it isn't really a state dependent thing and remains constant throughout the pipeline.

I don't really understand how having it typed would change the 1-to-1 ness / predictability of the generated identifiers. Also in some ways the agnostic approach is beneficial to our brains, but on the other hand not having the compiler telling us we are doing something wrong does fill me with a latent sense of danger when changing the modelling of stuff.

alexwlchan:
👀

jamesgorrie (approved):
Very much in favour of this!

rfcs/029-work_state_modelling/README.md
alicefuzier: it's true that we risk sending invalid works (or images or whatever) from the id minter. But 
a) if the id minter has run long enough with the same code we can be confident that it doesn't and the agnostic model means it doesn't change so that confidence isn't hard to achieve 
b) tests in the idminter are around works (maybe they shouldn't be though?) 
c) even if that happens it would be caught in the next step of the pipeline, we don't risk exposing invalid json to the API, say.
It's true that the 1-1 predictability would stay the same, but the ability to completely forget the id minter, as we do now most of the time, would be lost. As you say, that would mean a one-line change in most cases, but I think the current situation where we don't have to think about it at all is preferable and the downside of having to tweak the model to adjust that isn't that bad. (BTW sourceIdentifiers alongside canonicalId isn't the only thing idminter tied in the model. identifiedType is a field used by the idminter to change the type of things when they are Minted, like UnidenfiedWork -> IdentifiedWork. Which is a bit ugly)

alicefuzier (commented):
This is really great! Much more flexible and nicely separates the data from the state (which is a bit muddled at the moment).

rfcs/029-work_state_modelling/README.md
warrd: Yep I am very confident the `id_minter` is correct currently, it's more for when you make changes it is something which could be caught compile time. So yeh, when you make changes to the model the ability to forget the ID minter isn't really lost, as you have to think whether something you are doing is going to make errors at runtime. The situations where you should be thinking about rather than forgetting the ID minter are the same in both cases.

alicefuzier: mmh yes and no. this long discussion proves that you can't just forget the idminter when you change the model but also, this is a big model overhaul that we are not going to do often. When you just add/remove/rename a field, which is what we do in normal cases, you don't have to think about the id minter. 

Anyway, I think I see your point. I still think the cost of/risk is worth the simplicity of having an agnostic id minter but I agree there are drawbacks and see why we could prefer things being explicit and more verbose in this case.

rfcs/029-work_state_modelling/README.md
```diff
@@ -0,0 +1,177 @@
...
+  type Id <: IdentifierState
+  type ImageId <: IdentifierState
+}
+
+object WorkState {
+
+  case class Unmerged() extends WorkState {
+    type Id = IdentifierState.Unminted
+    type ImageId = IdentifierState.Unidentified
+  }
```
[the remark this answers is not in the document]
warrd: This is how things are currently. `Unminted` means it can be `Unidentified` or `Unidentifiable`. Images are always unidentifiable (unlike concepts etc), so it doesn't make sense to use `Unminted` for that. Naming is slightly weird, I agree, but that's the best we could come up with at the time.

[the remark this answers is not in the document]
jamieparkinson: @warrd assuming you mean "images are always _identifiable_ here"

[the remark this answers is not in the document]
warrd: Yep!

rfcs/029-work_state_modelling/README.md
```diff
@@ -0,0 +1,177 @@
...
+  *      ▼
+  *   Merged
+  *      |
+  *      | (relation embedder)
+  *      ▼
+  * Denormalised
+  *      |
+  *      | (id minter)
+  *      ▼
+  *  Identified
```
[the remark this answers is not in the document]
warrd: Hmmm yeh maybe a better name than `Work[Unmerged]` but not into `Work[Transformed]` personally as it's not really the work that is transformed, its the data that is transformed (with the work being the result of that transformation),

Can't think of any better ideas really tho.. `Work[InitialState]` / `Work[Init]` / `Work[Vanilla]`?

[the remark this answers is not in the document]
jamieparkinson: Something like `Work[Source]` / `Work[Origin]` / `Work[SingleOrigin]`?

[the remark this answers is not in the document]
jamieparkinson: Single-origin ethically sourced Works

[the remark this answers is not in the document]
warrd: I like `Work[Source]`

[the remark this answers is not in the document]
warrd: Whilst we're on this subject, any alternative names to `StandardWork[State]`?

rfcs/029-work_state_modelling/README.md
```diff
@@ -0,0 +1,177 @@
...
+  case class Unidentifiable() extends Minted with Unminted
+}
+
+/** Work contains the work itself. It is parameterised by its state, meaning
+  * the same type of Work can be in a number of possible states depending on
+  * where in the pipeline it is. This allows us to easily add new types of work
+  * (such as if Collection is decided to be a separate type to StandardWork),
+  * with the state of the work in the pipeline being an orthogonal concern.
+  */
+sealed trait Work[State <: WorkState] {
```
alexwlchan, line 123: Possibly question, but why is `State` a type parameter on `Work` if it's not used?

warrd: The subtypes of `Work` all use it, so it enables us to do things like this:

```scala
def getRelations(work: Work[Denormalised]): Relations[Denormalised] =
  work match {
    case StandardWork(_, _, state) => state.relations
    ...
  }
```

Otherwise if we just had a `sealed trait Work {}` every time we wanted to look in the state of one of it's cases we would just have an `Any` which would be pretty useless.

Regardless of access to particular data, it would still be useful for us to have it parameterised by some state due to the FSM part: we can ensure only transitions from one sort of work to another (which wouldn't be possible at the work level if only the sub types were parameterised).

alexwlchan: For that use case, I’d prefer

```
sealed trait Work[State <: WorkState] {
  …
  val state: State
}

def getRelations(work: Work[Denormalised]): Relations[Denormalised] =
  work.state.relations
```

at which point you delete the method as unnecessary.

warrd: Yep definitely agree we should do that too

PR #41, opened 2020-10-08
Add pipeline merging RFC

alicefuzier:
Rendered version https://github.com/wellcomecollection/docs/tree/pipeline_merging/rfcs/030-pipeline_merging

### DIFF
--- a/rfcs/030-pipeline_merging/README.md
+++ b/rfcs/030-pipeline_merging/README.md
@@ -1,0 +1,125 @@
+## Pipeline merging
+### What should the pipeline be doing?
+
+A few cases to remind ourselves what the behaviour of the whole pipeline should be when merging things:
+
+- A:1 -> B:1 if A:1 is the canonical results in:
+    - A:1 merged with B:1
+    - B:1 redirected to A:1
+ 
+- A:1 -> B:1 -> C:1 if A:1 is the canonical results in:
+    - A:1 merged with B:1 and C:1
+    - B1: redirected to A:1
+    - C:1 redirected to A:1
+ 
+ - assuming the previous case has happened, an update comes for B:2 which no longer links to C:1:
+    - A:1 merged to B:2
+    - B:2 redirected to A:1
+    - C:1 no longer redirected 
+
+### How this currently works
+
+#### Matcher
+The matcher receives individual works and updates stores the state of the graph.
+
+- It receives A:1 which says that A:1 -> B:1
+    - it stores the link A:1 -> B1
+    - it stores A:1 and B:1 as belonging to the same component AB
+    - it sends `[[A:1, B:1]]` to the merger
+- It receives B:1 -> C:1
+    - it stores the link B:1 -> C:1
+    - Updates A:1, B:1 And C:1 as belonging to ABC
+    - it sends `[[A:1, B:1, C:1]]` to the merger
+ - It receives B:2 breaking the link to C:1
+    - it stores B:2 with no links removing the link to C:1
+    - Updates A:1, B:2 ad belonging to AB
+    - it sends `[[A:1,B:2],[C:1]]` to the merger
+
+#### Merger
+The merger reads the ids and versions received by the matcher from the recorder store 
+and decides if and how to merge based on internal rules. If it decides to merge it updates the works as:
+- Fills `numberOfSources` field on the work selected as target with the number of works merged into it
+- Modifies all the other works as redirected poiting to the target
+
+if it decides not to merge, it sends the works unchanged
+
+##### Previous behaviour
+At some point (ie probably a few weeks ago) the merger used to send a `merged` boolean flag
+ to the ingestor that basically indicated if the work had been tampered with by the merger.
+
+#### Ingestor
+Because updates to works can get to the ingestor out of order, the ingestor assigns a version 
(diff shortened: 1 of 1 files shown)


kenoir:
You might want to add a brief bit of context up front in this RFC e.g. what is merging, why are we doing it.

rfcs/030-pipeline_merging/README.md
```diff
@@ -0,0 +1,124 @@
...
+that was the result of a previous unlink operation
+
+
+### Solution proposed
+
+The solution suggested by Nic is that we pass through in the work a field with the timestamp 
+of the last update.
+The merger should then use the maximum of the timestamps of all works sent by the matcher 
+as the version for each of the works.
+The ingestor should use greater than versioning instead of greater or equal.
```
warrd, line 109: Not sure I understand the reason for `gt` rather than `gte` :thinking:... I can't think of any data consistency issues that exist only with `gte` rather `gt`. One potential upside of `gte` would be that we could e.g. just redeploy the Calm transformer in an existing pipeline and run a Calm reindex where the data is retransformed (with `gt` any changes would be dropped).

Is there an issue I'm not thinking of? Sorry if we went over this in the meeting earlier.

rfcs/030-pipeline_merging/README.md
```diff
@@ -0,0 +1,124 @@
...
+ 
+The updated timestamp needs to be recorded in the adapter in different ways depending on the source:
+- Sierra: the sierra API already provides a timestamp with down to the second granularity 
+for each update and the Sierra adapter already stores it in the Sierra VHS. The sierra transfomer can 
+read this value and copy it into the work. 
+- Miro: Miro data doesn't change so can just set it to zero in the transformer
+- Calm: Calm data has a `retrievedAt` field in the Calm VHS. It comes from the HTTP header 
+of the response from the Calm API and it just represents when the response was sent from the server. 
+Because of this a lot of Calm records have the same timestamp of when the first harvest was done.
```
warrd, line 118: Looking at the code it looks to me like the `retrievedAt` field is initialised from the HTTP request for the individual Calm record, which would mean they would have different timestamps?

warrd:
From the meeting early it seemed like using the `retrievedAt` field etc is just for initialising the existing data in the adapter VHS, but going forward we would use `Instant.now()`. Maybe worth mentioning this if so?

rfcs/030-pipeline_merging/README.md
alicefuzier: If I undesrand the code, it comes from the response, not the request. But yeah, it's not exactly the same so probably no reason to worry and will update

warrd: Yep it's the response :+1:

rfcs/030-pipeline_merging/README.md
alicefuzier: Mh yeah, good point about transformer changes, didn't think about that. I thought that there was consensus in the meeting that the way we use gte at the moment is dodgy and the new method wouldn't need it but I don't see how it would hurt, so yeah let's leave it

alicefuzier:
Yeah, regarding the `Instant.now()` thing actually I'd like some more input. In the meeting, we said to take the timestamp with `Instant.now()` in all the adapters when we see an update for a record and use that in the work. 

Thinking about it a bit more, I have reservations about this approach:
- if we do a reharvest we end up with new times very close together which don't reflect when the actual update happened
- Sierra and METS already give us a granular enough and reliable time of the actual update, not when we first saw it in the catalogue, which means it won't change with a reharvest. We can just set the timestamp to zero for Miro cause it doesn't change

Form Calm, we could take the time with `Instant.now()` in the adapter instead of taking it from the response if we don't trust the response from the Calm server, but then I would also use that to populate the `retrievedAt` field. If instead, we trust the server enough to populate the `retrievedAt` then I think we should just use that and change only the Calm transformer, not the adapter. 

What do others think?

rfcs/030-pipeline_merging/README.md
```diff
@@ -0,0 +1,137 @@
+## Pipeline merging
+
+### What is merging and why do we need it
+
+The pipeline feeds from different sources. Records from different sources may contain information relating
```
kenoir, line 5: 👏

warrd:
Sounds good to me, I am happy if you are with the approach of not using `Instant.now()`

rfcs/030-pipeline_merging/README.md
```diff
@@ -0,0 +1,137 @@
...
+as the version for each of the works.
+The ingestor doesn't need to use greater than or equal to versioning anymore, but we will leave it so that 
+we don't necessarily have to reindex in a new index if we change transformer rules.
+ 
+The updated timestamp needs to be recorded in the adapter in different ways depending on the source:
+- Sierra: the sierra API already provides a timestamp with down to the second granularity 
+for each update. The Sierra adapter already stores it in the Sierra VHS. The sierra transfomer can 
+read this value and copy it into the work. 
+- Miro: Miro data doesn't change so we can just set it to zero in the transformer
+- Calm: Calm data has a `retrievedAt` field in the Calm VHS. It comes from the HTTP header
```
kenoir, line 131: It makes sense, but it is really a shame that Calm doesn't provide sub-day granularity!

PR #42, opened 2020-11-09
RFC: Relation Batcher

warrd:
Readable version: https://github.com/wellcomecollection/docs/blob/relation-batcher/rfcs/031-relation_batcher/README.md


jamieparkinson:
Could you explain why the router and the batcher need to be separate services rather than part of the relation embedder?

alicefuzier:
Mostly, it has to do with the fact that the batcher cannot scale. It can only be one task, so if it's pared/merged with the router or the relation_embedder they can't scale either.
The reason to separate the router is again because we want it to scale, but it needs to sit before the batcher

warrd:
Assuming the `batcher` is a separate service, the `router` could not be part of the `relation_embedder` (as then the messages will flow from `relation_embedder` -> `batcher` -> `relation_embedder`, with it doing very different things two times round).

In terms of the `batcher`, there are a number of issues with combining it with the `relation_embedder`. The `batcher` will probably be implemented with Akka via [`groupedWithin`](https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/groupedWithin.html) and setting a large `n` and interval (e.g. 30 minutes). We could just do this in the `relation_embedder`, which would work fine with a single task. However once we start scaling the `relation_embedder` this would result in the work being suboptimal (i.e. different concurrent tasks will end up processing the same archives and doing duplicate work). So we need the `batcher` to be a single task which does all the batching, allowing the `relation_embedder` to scale independently.

rfcs/031-relation_batcher/README.md
```diff
@@ -0,0 +1,112 @@
...
+In order to prevent the large amount of duplicate work performed within the `relation_embedder`, we propose adding two more small services to the pipeline before the `relation_embedder`: a `router` which extracts archive works and passes through non-archive works, and a `batcher` which groups works contained within the same archives over a given time period, preventing the `relation_embedder` from performing the excessive levels of duplicate work we are currently seeing.
+
+### Router
+
+The `router` takes a work id, looks it up in the `pipeline_storage` and checks if it's part of an 
+archive (i.e. has a `collectionPath` field).  If it is, it sends the value of `collectionPath` to the `batcher`, otherwise sending the work straight to the `id_minter_works`. 
+
+### Batcher
+
+The `batcher` groups work received over a given time period and works out the optimal set of paths to send towards
```
jamieparkinson, line 62: How is the number of works here bounded?

warrd: In practice we will (as I mentioned above) use [`groupedWithin`](https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/groupedWithin.html) and set a very large number (e.g. 100,000).

warrd: The messages consist just of paths. Say we have 100,000 paths which are 50 characters long, that is only 5MB, so memory is not going to be an issue with number of these sorts of sizes.

warrd:
> The reason to separate the router is again because we want it to scale, but it needs to sit before the batcher

This is true but also the method in how they run is quite different: for low latency of standard works the `router` should process every message as it comes from the `merger`, whereas to improve performance for the `relation_embedder` we need the `batcher` to run over larger time frames.

jamieparkinson:
2 related questions:

- Having the batching function (whether in its own service or as part of another) scale is suboptimal, but how suboptimal is it? How does the loss of performance there compare to the loss due to a bottleneck single-task, fixed-time process? I think you can estimate (assuming random distribution of nodes in the queue) that splitting the work across `n` workers would be equivalent to having a batch of `1/n`th the size.
- How sure are we that this will work? What sort of speedup could we expect and is there a way to verify without doing so much work up front?

alicefuzier:
regarding question number 1, I think the more you scale the batcher, the more unnecessary work the relation_embedder ends up doing, because as you say, we slice the size of the queue, so it's less able to filter out redundant work. redundant work we can see already is pretty bad for performance, so I don't think you would gain much performance wise if you make the batcher scale

jamieparkinson:
agreed - just wondering if that performance hit is bad enough to justify the extra complexity

alicefuzier:
I would say so yeah. The way the number of denormalisations it has to do grows seems to have a much worse impact

rfcs/031-relation_batcher/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Potential Issues
+
+* The `batcher` necessarily introduces some latency into the processing of archive works that is not present in the current pipeline. We are able to tweak the time processing interval for this to provide a compromise between `relation_embedder` performance, delay, and costs. One potential solution would be to make the interval duration change dynamically in response to pipeline load, so that it's big while there are many messages  in flight, but it's small if there are not many messages, decreasing latency for updates while retaining efficiency during a reindex.
+
+* Another disadvantage is the proposal introduces 2 new services, and some additional architectural complexity due to the fact that the messages split at the `router` before joining again at the `id_minter`. The exact number of services and modes of communication could probably be modified, but the particular design we came up with was the simplest we could think of in regards to not introducing complicated logic and additional storage beyond using queues with SQS.
```
warrd, line 112: Copying relevant parts of discussion from slack here:

kenoir:
I think trying to catch works that are related by slicing loading/processing messages into half an hour chunks, we might be better off running over the entire set of works after they've been reindexed and embedding relations then (after we've built the intermediate works store).

That could also be performed with a filter to works only with a `collectionPath` as suggested to be performed by the `router` in this RFC.

rfcs/031-relation_batcher/README.md
jamieparkinson: I think that this RFC proposes an implicit split between the tree-building and the denormalisation stage; the batcher builds a tree in memory in which only parent->child references require updates in order to avoid our need to denormalise entire lineages on every update.

I would be interested in exploring solutions which make this 2-step process explicit, hopefully avoiding complexity in the process.

warrd: > @kenoir I think we need to talk about this, it feels like we might need to do something different with relationships to get this to work rather than trying to shoehorn it into the existing pipeline
> @jamesgorrie A lot of the work around concepts in the pipeline pointed in that direction too.
> @jamieparkinson It feels like tree building should be completed before denormalisation happens - as all you need to represent a tree is to contain child references in parents. I think this is really what the proposed batcher is doing (just doing so implicitly and - intentionally - suboptimally)
> @kenoir so something like reindex entirely without denormalisation then iterate over all works denormalising data onto them as appropriate?
> @jamieparkinson something like that - maybe with some intermediate storage, but then again I can’t actually think of why that would be necessary… trees are really simple data structures, we don’t need anything fancy to represent them
> @jamieparkinson maybe we end up with 2 indices, one denormalised and one not
> @kenoir run the denormalisation in process on a schedule for all works (or for just those that have changed in a given time period)
> @jamesgorrie this is where we got to with concepts. Trying to imagine the pipeline as: `source data (calm etc) => transforms => works index` and then `works index => transforms => denormalised works`

warrd: To address the tree building issue: as me and @alicefuzier discussed on a call with @kenoir, building the trees is really not the issue. Having a seperate graph data store or what have you would not change the nature of the work, we already have a representation of trees that is queryable (i.e. as a `collectionPath`), and having a tree built before denormalisation would still require the same amount of work.

warrd: > so something like reindex entirely without denormalisation then iterate over all works denormalising data onto them as appropriate?

> run the denormalisation in process on a schedule for all works (or for just those that have changed in a given time period)

I would argue that this is kind of what the the proposal in this RFC is doing. If we only wanted the denormalisation process to happen daily or whatever we could set the interval to 24 hours.

> Trying to imagine the pipeline as: source data (calm etc) => transforms => works index and then works index => transforms => denormalised works

I also think this is what the pipeline currently does. For me having an explicit split between these two parts of the pipeline does not really give anything: maybe you could deploy them separately maybe, but not sure whether that would be helpful or a hindrance.

warrd: > I think that this RFC proposes an implicit split between the tree-building and the denormalisation stage; the batcher builds a tree in memory in which only parent->child references require updates in order to avoid our need to denormalise entire lineages on every update.
> I would be interested in exploring solutions which make this 2-step process explicit, hopefully avoiding complexity in the process.

Making this more explicit maybe be useful in future (as I discussed in the original [Relation Embedder RFC](https://github.com/wellcomecollection/docs/tree/master/rfcs/027-relation-embedder#graph-store)).

However imo it does not avoid any complexity and would definitely add some: having a extra store introduces more state into the pipeline, requires more infra, code for interacting with new databases etc.

warrd: > I think trying to catch works that are related by slicing loading/processing messages into half an hour chunks, we might be better off running over the entire set of works after they've been reindexed and embedding relations then (after we've built the intermediate works store).

What I think this proposes is modifying the existing index (i.e. the one the API is reading from) at regular intervals. @kenoir can you explain why you think this would be better? To me this has lots of downsides:

* No longer a uni directional flow of data from inputs to outputs
* Separate to the main pipeline
* Puts more strain on the index that we are serving the API from

warrd: Sorry for so many seperate messages. I am not being critical of all the above due to being wedded to the current design, just need to be convinced an alternative is an improvement / more complexity is worth it.

alicefuzier:
> I think trying to catch works that are related by slicing loading/processing messages into half an hour chunks, we might be better off running over the entire set of works after they've been reindexed and embedding relations then (after we've built the intermediate works store).

The batcher could do that as @warrd said, you just make the time frame bigger, but it's more flexible than just that. The batches as defined in the RFC could have a dynamic timeframe to allow updates to go through in near real time. Also, I don't really see the advantage of doing that as relations won't be queriable in the api, so the index will (presumably?) not be usable by the api. It seems to me this only decreases performance/incresase delay?

rfcs/031-relation_batcher/README.md
kenoir: >we do this on an intermediate store. 

I do mean on an intermediate store, yes.

kenoir:
> The batcher could do that as @warrd said, you just make the time frame bigger, but it's more flexible than just that. The batches as defined in the RFC could have a dynamic timeframe to allow updates to go through in near real time.

This made me reconsider a bit. I can see that it's really the same thing as saying, "stop, build the graph, then carry on" but with variable performance and a non-world stopping approach ("build a hopefully connected chunk of the graph, then carry on").

> It seems to me this only decreases performance/incresase delay?

Yes. In which case it does add weight to the approach suggested here.

rfcs/031-relation_batcher/README.md
jamieparkinson: > having a tree built before denormalisation would still require the same amount of work.

The same amount of work as the solution proposed in this RFC? I think it would be slightly less due to the fact that there would be guaranteed no duplicated effort, right?

Regardless I'm not sure what you mean by "the tree building issue" - as you say this part is easy, that's why I was suggesting separating the denormalisation (the hard part) from it entirely.

warrd: Maybe we are talking about different things here. Building a tree up front does not affect the number of opera the relation embedder has no to perform: you still query for other affected works (i.e. the subtree from its parent downwards), and then for every one of those works you query for its relations.  It just affects the way you make those two queries (whether using paths or another graph representation). The performance issue currently is not the query times but the fact that works are denormoralised multiple times, doing that upfront does not affect it.

jamieparkinson:
> The batches as defined in the RFC could have a dynamic timeframe to allow updates to go through in near real time.

Yeah, this makes me reconsider too - rather than being about timeframes, is this perhaps more about backpressure? Could we use [conflate](https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/conflate.html) or [batch](https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/batch.html) to perform the aggregation as and when that optimisation is required?

rfcs/031-relation_batcher/README.md
alicefuzier: > I was suggesting separating the denormalisation (the hard part) from it entirely.

I don't think I follow what you mean. Is it something like @kenoir suggestion to do denormalisation only after having build an intermediate storage of all works without denormalisation?

jamieparkinson: What I'm saying is effectively the same as @kenoir's suggestion, yes.

I am saying that building a tree up front means that you have enough information afterwards to perform an optimal denormalisation whereas denormalising as you go causes duplicate work as the tree is built - is that wrong?

warrd: There is no concept of up front though really, as works keeping coming in. So the best you can do is within a time interval (although maybe can get more clever with respect to reindexing etc)

alicefuzier:
>  is this perhaps more about backpressure? Could we use conflate or batch to perform the aggregation as and when that optimisation is required?

Mhh i'm not really sure about that. I agree that the problem the batcher is solving is closer to backpressure than tree building. How I think about it is it's a filter to reduce redundant work. Do you mean using `conflate` and `batch` in the relation_embedder to avoid having a separate `batcher` service?

jamieparkinson:
I see what you mean - I was thinking of handling those would be in the batcher, but I can see that actually the backpressuring operation doesn't live there. To me that's another indication that something about the way these services are split up isn't quite right, rather than an indication that backpressure is the wrong model for the problem - I might add to your description that it's a filter to reduce redundant work *when the capacity to do that work isn't available*...

rfcs/031-relation_batcher/README.md
alicefuzier: In a reindex, with no dynamic updates coming through, the optimal would be to send the root of each archive after its descendants have all gone through. So if you just stopped at an intermediate store, you could just do that and you would have the relation_embedder building each tree exactly once (correct me if I'm wrong @warrd )

But we have dynamic updates coming from all sources which could potentially affect any denormalisation, so as @warrd is saying, it's a bit harder in that scenario to figure out when/how to stop cause it's a constant stream of updates.

Even for the reindex scenario, I think stopping at an intermediate stage only adds delay. It is what the batcher would be doing anyway and if you made the timeframe long enough you would hit the best case scenario. But then you would make the time that it takes to have a usable index bigger I think

jamieparkinson:
What aspect of the relation embedder work is slow? The ingesting or the querying?

alicefuzier:
The problem (again @warrd correct me if I'm wrong) is that it's a one to many operation. So, for a work coming in, multiple works might end up denormalised and updated. Because it's one to many, updates to different nodes in the tree can cause overlapping sets of works to be denormalised and often these sets are not just overlapping, but exactly the same and with _a lot_ of works. So the query time and ingest time are fine, but the volume of operations is not. And a lot of these operation are just repeated over and over for no reason.

jamieparkinson:
I wonder if we could have a non-scaling batcher that has worker (actor) embedders upon whose mailboxes it backpressures? Rather than a linear one-directional flow

warrd:
> What aspect of the relation embedder work is slow? The ingesting or the querying?

Those two things take time, and there is other CPU bound things (deserialising responses from ES, sorting relations before denormalisation)

But I think framing the issue like that is missing the point, it's more that the number times we are doing those operations scales with `O(n^2)` when we could achieve it in the ideal case with `O(n)`.

jamieparkinson:
Wasn't saying that's the issue, just curious as to whether it's IO-, CPU-, etc- bound to try to work out how the work might get split up and scaled

alicefuzier:
> I wonder if we could have a non-scaling batcher that has worker (actor) embedders upon whose mailboxes it backpressures? Rather than a linear one-directional flow

I don't think I understand how that would look like

rfcs/031-relation_batcher/README.md
jamieparkinson: Yeah, I think I see now that separating the denormalisation step doesn't really offer anything better than an in-pipeline batching solution. Still want to discuss further whether a non-scalable batcher working on a time period is the best way to do that though.

alicefuzier:
>  I might add to your description that it's a filter to reduce redundant work when the capacity to do that work isn't available...

Mhh not sure I agree. The batcher doesn't know if there's available capacity. What it does is just filter out redundant work _within its timeframe_

jamieparkinson:
Basically I was trying to think of how you would implement the "dynamic timeframe" - I know the batcher here doesn't know the state of the relation embedder, but I feel like it _should_ otherwise it can't know how to set the timeframe. I was thinking of how you could get separate embedders to exert backpressure on the batcher.

warrd:
I think it's a cool idea but not sure how we would implement it in practice. Could kind of simulate it by having the `batcher` reading the number of messages in the `relation_embedder` queue and reacting to that.

kenoir:
> know the batcher here doesn't know the state of the relation embedder,

You can perhaps make some guesses by the length of the SQS queue that feeds it.

alicefuzier:
> Basically I was trying to think of how you would implement the "dynamic timeframe" 
We didn't put it in the RFC and perhaps we should, but a simple idea that @warrd came up with is make the batcher remember the number of works it read from the queue at rthe previous run. If it's increased => increase the timeframe; if it's decreased => decrease the timeframe

warrd:
> We didn't put it in the RFC and perhaps we should, but a simple idea that @warrd came up with is make the batcher remember the number of works it read from the queue at rthe previous run.

Yeh this would work OK, but on further consideration its more the number of messages being queued further down the pipeline that should inform how often we batch.

alicefuzier:
Yeah, that would make the batcher a backpressurer for the relation_embedder effectively, but across all instances of the relation_embedder. I think either way the point is there's many ways of doing that and it's not very complicated to do

jamieparkinson:
Length of SQS queue + batch/conflate operators sounds like a reasonable heuristic way to do that to me. That said, I still don't like the fact that we'll have a non-scalable service in the pipeline... do we know of any ways around that?

warrd:
What in particular don't you like about it being non-scalable? It is extremely lightweight compared to other services (only IO is reading SQS, writing to SNS), and only runs once every 30 mins or so.

alicefuzier:
The thing is the batcher is a bit working opposite to the rest of the pipeline, cause really when we have loads of messages, you want the batcher to have visibility to the biggest portion of the queue that you can. So if we have many batcher instances, and each one reads a bit of that queue, you significantly reduce the performance gains, rather than increasing them.

kenoir:
I wonder if only time is actually the right way to split things ... rather than by number of messages.

kenoir:
> I wonder if only time is actually the right way to split things ... rather than by number of messages.

You'd need some aspect of time too obviously so you don't get stuck waiting for a batch to finish filling up.

jamieparkinson:
I can see why it needs to be non-scaling - it's solely that it doesn't feel good to have a service which will actually _slow down_ the pipeline if it scales up.

kenoir:
> it's solely that it doesn't feel good to have a service which will actually slow down the pipeline if it scales up.
I agree with the sentiment (I think even if you batch by number of messages you still need to process them serially?).

warrd:
> I wonder if only time is actually the right way to split things ... rather than by number of messages.

Yep there is some discussion above [about this](https://github.com/wellcomecollection/docs/pull/42#discussion_r520496786), sorry easy to miss.

alicefuzier:
> I can see why it needs to be non-scaling - it's solely that it doesn't feel good to have a service which will actually slow down the pipeline if it scales up.

Yeah, I agree it's a bit counterintuitive. But I think the advantages that we gain in performance are worth it. And it would result in a very simple and lightweight code

alicefuzier:
> Yep there is some discussion above about this, sorry easy to miss.

Yes, but that discussion is more about capping messages to avoid them being too many. If I understood correctly, the focus is on time first and number of messages second and the idea is we would make the number of messages big enough to not have memory issues. 
if I understood @kenoir comment, he's instead suggesting to put the focus on number of messages first rather than time. Which is an interesting point actually cause it is the important thing, but it's implicit in the time based solution. I wonder whether it would be as easy to adapt dynamically

warrd:
Not sure if people are still hung up on this or not, but if it actually came to point where we needed to scale the batcher (which I just think is never going to need to happen), you could have multiple containers which process paths depending on a range in the alphabet e.g. 5 seperate nodes each with an individual queue for paths starting with ranges `A-F`, `G-K`, `L-P`, `Q-U`, `V-Z`

jamieparkinson:
Could you add something to the RFC about how you might dynamically change the batching timeframe (eg to be lower when a reindex isn't taking place) as mentioned above?

I'm still not totally convinced by this approach, to be honest, and I don't know how applicable it will be to other denormalisation applications that we might encounter, but I can't think of anything better / nor do I think any of the misgivings I have are enough to say it isn't the best thing to do right now. I'll be interested to see how it goes in practice.

PR #43, opened 2020-11-24
Incident response protocol

kenoir:
Describing our process for dealing with incidents on the Wellcome Collection digital platform.

### DIFF
--- a/INCIDENTS.md
+++ b/INCIDENTS.md
@@ -26,7 +26,7 @@ There are some situations where a person decides that it is best to initiate the
 
 ### Incident lead
 
-When it's clear an incident has occured the person responsible is the most senior developer who is available to respond. Who that is can be worked out via Slack in the #wc-platform channel, this person is now the "incident lead" until they pass that role on explicitly to someone else. It is not expected that the most senior developer lead on all incidents, and that role should be handed to the person best placed to lead on a solution.
+When it's clear an incident has occured the person responsible is the most senior developer who is available to respond. Who that is can be worked out via Slack in the #wc-platform channel. This person is now the "incident lead" until they pass that role on explicitly to someone else. It is not expected that the most senior developer lead on all incidents, but rather the role should be handed to the person best placed to lead on a solution.
 
 The incident lead should announce in #wc-platform that an incident has occured, ensuring that the relevant delivery manager is aware of the situation   (and agrees to continue where appropriate). A new Slack channel should be created for that incident, started with description of the problem and what is known at that time.
 
--- a/INCIDENTS.md
+++ b/INCIDENTS.md
@@ -17,7 +17,10 @@ An incident must be worked on as soon as possible to mitigate loss of service to
 
 An potential incident might be reported in lots of ways (tweet from a user, report from a developer) but should be funneled to a delivery manager to assess the damage and decide whether to call the issue an "incident" or prioritise work via other channels.
 
-There will be situations where it is obvious that a situation needs to be delat with immediately and doesn't need "approval" by a delivery manager to be called an incident, in that situation ensure a delivery manager is informed as part of inititiating an incident.
+There are some situations where a person decides that it is best to initiate the protocol themselves without a delivery manager at first. When this happens, ensure that the delivery manager is informed that this has happened. Some examples of where this might happen
+
+* The incident is small enough to get a rapid fix out
+* A delivery manager is unavailable
 
 ## Roles
 
--- a/INCIDENTS.md
+++ b/INCIDENTS.md
@@ -3,13 +3,20 @@
 
 ## What's an incident?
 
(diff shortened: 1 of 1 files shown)


INCIDENTS.md
```diff
@@ -0,0 +1,48 @@
...
+- Search APIs being unavailable or serving incorrect data
+- A security compromise such as credentials being exposed on the public internet.
+
+An incident must be worked on as soon as possible to mitigate loss of service to users and minimise reputational damage.
+
+## Reporting an incident
+
+An potential incident might be reported in lots of ways (tweet from a user, report from a developer) but should be funneled to a delivery manager to assess the damage and decide whether to call the issue an "incident" or prioritise work via other channels.
+
+There will be situations where it is obvious that a situation needs to be delat with immediately and doesn't need "approval" by a delivery manager to be called an incident, in that situation ensure a delivery manager is informed as part of inititiating an incident.
```
jamesgorrie, line 20: I am unclear about this - assuming this is saying that anyone can initiate the protocol? If that's right would something like this work?

```suggestion
There are some situations where a person decides that it is best to initiate the protocol themselves without a delivery manager at first. When this happens, ensure that the delivery manager is informed that this has happened. Some examples of where this might happen

* The incident is small enough to get a rapid fix out
* A delivery manager is unavailable
```

INCIDENTS.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Roles
+
+### Incident lead
+
+When it's clear an incident has occured the person responsible is the most senior developer who is available to respond. Who that is can be worked out via Slack in the #wc-platform channel, this person is now the "incident lead" until they pass that role on explicitly to someone else. It is not expected that the most senior developer lead on all incidents, and that role should be handed to the person best placed to lead on a solution.
```
jamesgorrie, line 26:
```suggestion
When it's clear an incident has occured the person responsible is the most senior developer who is available to respond. Who that is can be worked out via Slack in the #wc-platform channel. This person is now the "incident lead" until they pass that role on explicitly to someone else. It is not expected that the most senior developer lead on all incidents, but rather the role should be handed to the person best placed to lead on a solution.
```

INCIDENTS.md
```diff
@@ -0,0 +1,51 @@
+# Incident Response Protocol
+
+
+## What's an incident?
+
+An incident is defined here as a severe immediate or short-term potential degredation in the user-facing performance of our services.
```
jamieparkinson, line 6: I think that the definition should remain light-touch, but is there anything we can put here that will help us take appropriate actions earlier - thinking of the fact that the incident that prompted this doc took several hours to notice because we all assumed that it was a minor issue only affecting users behind a particular toggle.

Automated testing helps, but there will always be something we haven't tested for - maybe we could say something about bare minimum investigations of bugs before we decide if they are/aren't an incident? Not sure

INCIDENTS.md
```diff
@@ -0,0 +1,51 @@
...
+
+The incident lead should announce in #wc-platform that an incident has occured, ensuring that the relevant delivery manager is aware of the situation   (and agrees to continue where appropriate). A new Slack channel should be created for that incident, started with description of the problem and what is known at that time.
+
+The incident lead should pull in whoever they need to understand and resolve the problem. All those involved in resolving the incident should report their progress in the shared incident channel. When someone is no longer able to assist or has stopped helping that should also be made clear.
+
+### Supporting colleague
+
+If an incident lead requests your help to resolve an issue, please assist as required reporting your progress in the shared Slack channel. If you are unable to assist or are no longer working on the problem please also make that clear.
```
jamieparkinson, line 37: Anything about whether people should or shouldn't support if they're not asked to? Sometimes we end up with a "too many cooks" situation where useful information gets lost or progress is slowed by discussion, but equivalently I'm not sure we want to all sit there ignoring it until told to do something.

INCIDENTS.md
jamesgorrie: Could it be turning this into a set of questions?

* Who is affected?
* How has it affected their service needs?
* Are their workarounds we could communicate?
* What comms would help mitigate interference?

After answering these questions you could do a quick review with someone to assess the damage?

INCIDENTS.md
jamesgorrie: The leads discretion to put a team together if they need it and ask relevant questions to The Right People™ when they need to.

It feels like there might be a two different roles:
* Lead and designated people solving the problem
* Everyone offering any information they know

I've found the too many cooks is generally in people trying to solve the problem, assuming mileage may differ.

kenoir: >The leads discretion to put a team together if they need it

I think so.

>Anything about whether people should or shouldn't support if they're not asked to?

Something like "offers of help should be first directed to the incident lead"?

INCIDENTS.md
kenoir: Added some questions to ask when a potential incident arises.

PR #44, opened 2021-01-28
RFC: Calm deletion watcher

jamieparkinson:
Still a bit of a wip, thought I'd get it out for discussion before refining further.

Formatted RFC: https://github.com/wellcomecollection/docs/blob/calm-deletions-rfc/rfcs/032-calm-deletions/README.md

### DIFF
--- a/rfcs/032-calm-deletions/README.md
+++ b/rfcs/032-calm-deletions/README.md
@@ -20,7 +20,7 @@ The deletion watcher will consist of:
 
 The lambda will place a message on the reindexer queue requesting a "reindex" of Calm records to be sent to the worker. It will be triggered by a scheduled Cloudwatch event or manually. By default it will request the full source scan but in the case of a manual trigger it will be possible to request specific IDs, which will be useful in the case of expediting known deletions.
 
-The worker will consume records and check for their existence in the Calm API. If the record has been deleted (is no longer present) then it will update the VHS entry to flag the deletion. 
+The worker will consume records, filter out those that are already flagged as deleted, and check for the existence of the rest in the Calm API. If the record has been deleted (is no longer present) then it will update the VHS entry to flag the deletion. The deletion flag will live in the DynamoDB record so that the flag can be checked without fetching the object from S3.
 
 The calm transformer will check for the presence of the deletion flag in the source data and create `Deleted` works as appropriate.
 
--- a/rfcs/032-calm-deletions/README.md
+++ b/rfcs/032-calm-deletions/README.md
@@ -12,19 +12,24 @@ So far as we can tell, there is no way to find deleted records other than checki
 
 ## Proposal
 
-- The deletion watcher will be an ECS task within the existing calm adapter cluster.
-- It will run as a scheduled task at a frequency to be agreed with the collections information team.
-- It will also be possible to run the task manually via the `RunTask` API. A script will be provided for this.
-- The task will scan the source data VHS table and check for the existence of each record in the Calm API. If the record has been deleted (is no longer present) then it will update the VHS entry to flag the deletion. 
-- The calm transformer will check for the presence of the deletion flag in the source data and create `Deleted` works as appropriate.
+The deletion watcher will consist of:
+
+- A worker task in the calm adapter cluster
+- A lambda
+- The existing reindexer
+
+The lambda will place a message on the reindexer queue requesting a "reindex" of Calm records to be sent to the worker. It will be triggered by a scheduled Cloudwatch event or manually. By default it will request the full source scan but in the case of a manual trigger it will be possible to request specific IDs, which will be useful in the case of expediting known deletions.
+
+The worker will consume records and check for their existence in the Calm API. If the record has been deleted (is no longer present) then it will update the VHS entry to flag the deletion. 
+
+The calm transformer will check for the presence of the deletion flag in the source data and create `Deleted` works as appropriate.
 
 This architecture is illustrated below (arrows indicate the direction of data flow):
 
-![image](https://user-images.githubusercontent.com/4429247/106134857-08103780-615f-11eb-845f-112109ead907.png)
+![image](https://user-images.githubusercontent.com/4429247/106171539-8da8dd00-6189-11eb-9aca-2577f1ab6ed7.png)
 
 ## Questions & potential issues
 
 - Are we really 100% sure that polling for deletions is the only way to detect them?
-- Can we reuse any existing things for this (eg the reindexer)?
 - Might we want to be continually checking (more like crawling) for deletions rather than doing it all at once?
 
--- a/rfcs/032-calm-deletions/README.md
+++ b/rfcs/032-calm-deletions/README.md
@@ -1,0 +1,30 @@
+# RFC 032: Calm deletion watcher
+
+**Status:** Draft
+
+**Last updated:** 28/01/2021
+
+## Background
+
+We've realised that the pipeline does not pick up deleted Calm records - once they're deleted there is no trace of them in the Calm API, but they remain in the source VHS tables. Especially because a cleanup project at the end of last year was responsible for a very large number of deletions, this needs to be resolved.
+
+So far as we can tell, there is no way to find deleted records other than checking for their absence. This means that our solution needs to take some sort of polling-based approach: going through the source table and checking for the continued presence of each record in the Calm API.
+
+## Proposal
+
+- The deletion watcher will be an ECS task within the existing calm adapter cluster.
+- It will run as a scheduled task at a frequency to be agreed with the collections information team.
+- It will also be possible to run the task manually via the `RunTask` API. A script will be provided for this.
+- The task will scan the source data VHS table and check for the existence of each record in the Calm API. If the record has been deleted (is no longer present) then it will update the VHS entry to flag the deletion. 
+- The calm transformer will check for the presence of the deletion flag in the source data and create `Deleted` works as appropriate.
+
+This architecture is illustrated below (arrows indicate the direction of data flow):
+
+![image](https://user-images.githubusercontent.com/4429247/106134857-08103780-615f-11eb-845f-112109ead907.png)
+
+## Questions & potential issues
+
+- Are we really 100% sure that polling for deletions is the only way to detect them?
+- Can we reuse any existing things for this (eg the reindexer)?
+- Might we want to be continually checking (more like crawling) for deletions rather than doing it all at once?
+


rfcs/032-calm-deletions/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+We've realised that the pipeline does not pick up deleted Calm records - once they're deleted there is no trace of them in the Calm API, but they remain in the source VHS tables. Especially because a cleanup project at the end of last year was responsible for a very large number of deletions, this needs to be resolved.
+
+So far as we can tell, there is no way to find deleted records other than checking for their absence. This means that our solution needs to take some sort of polling-based approach: going through the source table and checking for the continued presence of each record in the Calm API.
+
+## Proposal
+
+- The deletion watcher will be an ECS task within the existing calm adapter cluster.
+- It will run as a scheduled task at a frequency to be agreed with the collections information team.
+- It will also be possible to run the task manually via the `RunTask` API. A script will be provided for this.
```
alexwlchan, line 17: I feel like specifying the exact API here is overkill, and also not quite what I had in mind.

I imagine this would be an SQS-backed autoscaling worker (like everything else we have), and triggered by a Lambda posting a message to an SNS topic. If you want to run it manually, you run the Lambda through the console. Similar to the snapshot generator.

rfcs/032-calm-deletions/README.md
```diff
@@ -0,0 +1,30 @@
...
+- The calm transformer will check for the presence of the deletion flag in the source data and create `Deleted` works as appropriate.
+
+This architecture is illustrated below (arrows indicate the direction of data flow):
+
+![image](https://user-images.githubusercontent.com/4429247/106134857-08103780-615f-11eb-845f-112109ead907.png)
+
+## Questions & potential issues
+
+- Are we really 100% sure that polling for deletions is the only way to detect them?
+- Can we reuse any existing things for this (eg the reindexer)?
```
alexwlchan, line 28: We can reuse some of the reindexer code, in particular the parallel scan. That allows you to distribute work among many workers, rather than having a single worker that does the whole table.

Or just have the reindexer fire records directly at whatever's polling Calm.

rfcs/032-calm-deletions/README.md
```diff
@@ -0,0 +1,30 @@
...
+- The task will scan the source data VHS table and check for the existence of each record in the Calm API. If the record has been deleted (is no longer present) then it will update the VHS entry to flag the deletion. 
+- The calm transformer will check for the presence of the deletion flag in the source data and create `Deleted` works as appropriate.
+
+This architecture is illustrated below (arrows indicate the direction of data flow):
+
+![image](https://user-images.githubusercontent.com/4429247/106134857-08103780-615f-11eb-845f-112109ead907.png)
+
+## Questions & potential issues
+
+- Are we really 100% sure that polling for deletions is the only way to detect them?
```
alexwlchan, line 27: It's one of (at least) two approaches, in my mind.

You could also reharvest the entire Calm catalogue into a fresh VHS, compare the IDs, and work out what's missing.

There are ~370k items in the current Calm VHS. Call it 400k to keep the numbers simple. Now suppose you have two copies of it, in memory, that's 800k. If each UUID takes a kilobyte of memory, that's 800MB, which should fit into the memory of one of our apps. Diff those two sets, voila, you have your deleted items.

The optimal approach would probably be a hybrid of both:

* Automatically and continually polling records that haven't changed in a while on a fixed schedule, and pruning them that way
* A button that lets us do a complete reharvest and mark any missing records as deleted

rfcs/032-calm-deletions/README.md
jamieparkinson: I don't think it needs to be that complex, though - ECS [provides](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/scheduled_tasks.html) task scheduling and we use `RunTask` elsewhere (eg for the training tasks for inference services, and previously for the reindexer).

SQS is useful when dynamic input is required, but that's not the case here. What are the advantages of Lambda + SQS + SNS in this case?

rfcs/032-calm-deletions/README.md
jamieparkinson: I think this task should stay within calm adapter infra rather than using the reindexer. I'm not sure if you mean multiple workers in the sense of multiple tasks but I'm not sure this needs to scale to > 1 task, if it's only running once a week or something like that it doesn't matter if it takes a little while...

rfcs/032-calm-deletions/README.md
jamieparkinson: Not sure I understand - what's the advantage of having an intermediate VHS for comparison? It seems like it would be much slower and more expensive.

I agree that we need both of those bullet points, I think this RFC _mostly_ covers them but seeing whether they've recently been changed needs a whole extra trip to the S3 portion of the VHS - is it worth it?

rfcs/032-calm-deletions/README.md
alexwlchan: > SQS is useful when dynamic input is required, but that's not the case here

Hmm. Are we sure about that?

If you make this an SQS-backed worker that takes a (CalmSourceRecord) and checks the freshness of that, then:

1. You can reuse the reindexer pretty much unmodified for scanning the table
2. It's not an issue if you get interrupted halfway through a scan
3. We can yank individual records with a specific reindex outside the usual schedule, without having to rescan the whole table (which might take hours or days)

I'm sure there's a bias towards doing another SQS-backed worker because we already run a bunch of those and have plenty of tooling for running them. It's a much more well-understood process than using RunTask.

(The Lambda is probably wrong for this, you're right.)

rfcs/032-calm-deletions/README.md
alexwlchan: Yeah, you don't really need the intermediate VHS unless you plan to drop the existing one (which we don't, but we should write down why it's not enough to reharvest and move on).

That depends on what data we record in the Calm DynamoDB table, and whether we want to record freshness checks somewhere. I idly wondered whether we'd put a `lastFetchedFromCalm` field in DynamoDB which the freshness checker would update, but I think that's probably wrong – a whole pile of conditional writes we mostly don't care about.

alexwlchan: It occurs to me as well that a reharvest may also be a little tricksy, because it takes a while.

If there's a record in the Calm VHS that you didn't see during the reharvest, is it because that record got deleted before you started the reharvest, or added afterwards?

rfcs/032-calm-deletions/README.md
alexwlchan: > I think this task should stay within calm adapter infra rather than using the reindexer

Why? We already cross the cluster boundary elsewhere.

> I'm not sure if you mean multiple workers in the sense of multiple tasks but I'm not sure this needs to scale to > 1 task, if it's only running once a week or something like that it doesn't matter if it takes a little while...

True, and we don't want to hammer the Calm API too hard, but see comment above about using SQS to handle being interrupted halfway through.

rfcs/032-calm-deletions/README.md
jamieparkinson: I don't really follow this idea of "freshness" - could you elaborate on that? All we're checking for is whether things have been deleted.

alexwlchan: Ah sorry.

For me, the notion of "freshness" is about _"when did I last _**know**_ this thing was in Calm"_?

If a record was updated yesterday, it's very "fresh" – it seems unlikely that it's been deleted.<br/>
If a record was updated last year and we haven't polled for it in a month, it's more likely that it's been deleted.

And I was wondering if that might affect how the deletion checker works, e.g. _"I don't re-poll for a record that was updated less than a week ago"_ (depending on how C&I feel).

jamieparkinson: Yeah, I did wonder about these optimisations but I think they're really quite complicated for minimal gain - record freshness is going to follow a power law distribution where the _vast_ majority of records haven't been updated in a long time, so the reduction in requests that doing this would offer is likely to be a fairly small percentage of the total.

rfcs/032-calm-deletions/README.md
jamieparkinson: Given that the job of the adapters is to reflect the true source data inside our walls, I think the deletion watcher is very clearly a part of the calm adapter - so if we were to do this I think we'd want to put the reindex worker task inside the adapter cluster (which is perfectly achievable) so it would all be deployed into one place.

alexwlchan: I’m not sure I agree – wouldn’t that mean we should put the reindex in the catalogue pipeline clusters as well? Or indeed, smush everything into one super cluster?

rfcs/032-calm-deletions/README.md
jamieparkinson: Hmmm, yeah, I can see the advantages - the ability to update specific records is probably the most useful aspect of this approach.

At the same time 2 tasks + scaffolding for something this simple seems a bit like overkill! It would also _double_ the number of table reads required as the reindexer sends (and needs to send, because of SQS message sizes) only references to source data rows, rather than the whole records. Not sure (as in, genuinely can't decide) if we'd want to incur that penalty for these advantages...

kenoir: Using only the `RunTask` API provides limited guarantees (and auditing) that the task that is launched completes successfully.

rfcs/032-calm-deletions/README.md
jamieparkinson: I don't think so - you have to draw the boundaries somewhere and that's where I believe they should be for this work. As I say, it doesn't actually make much difference here as we can run the reindexer task wherever we want.

rfcs/032-calm-deletions/README.md
alexwlchan: > At the same time 2 tasks + scaffolding for something this simple seems a bit like overkill! It would also double the number of table reads required as the reindexer sends (and needs to send, because of SQS message sizes) only references to source data rows, rather than the whole records. Not sure (as in, genuinely can't decide) if we'd want to incur that penalty for these advantages...

Aha! So this may be a source of the difference of opinion here – the reindexer _used_ to do that, but no longer!

This is an optimisation Alice suggested – the reindexer now sends the complete record, to save doing the double DynamoDB lookup in the transformers. If the service can tap that feed (which can be added with a config change in the Terraform), it gets a copy of every row in DynamoDB.

rfcs/032-calm-deletions/README.md
alexwlchan: So I guess I'm just wary of having multiple instances of the reindex worker floating around, because it adds complexity for deployment and triggering a reindex.

It's been acceptable to have the reindex worker living separately from the tasks before, and inter-cluster communication over queues. I'm still struggling to see the benefit, and what makes this process different from the rest of the pipeline.

kenoir:
Another approach may be to "take over" the deletion process from Calm, i.e. we provide a mechanism to do deletions that tells both Calm and Catalogue that a deletion has taken place (this seems simpler in some ways to me than the process described in this RFC). We would need to engage product folk to really understand if it's a viable approach though.

rfcs/032-calm-deletions/README.md
```diff
@@ -0,0 +1,30 @@
...
+## Background
+
+We've realised that the pipeline does not pick up deleted Calm records - once they're deleted there is no trace of them in the Calm API, but they remain in the source VHS tables. Especially because a cleanup project at the end of last year was responsible for a very large number of deletions, this needs to be resolved.
+
+So far as we can tell, there is no way to find deleted records other than checking for their absence. This means that our solution needs to take some sort of polling-based approach: going through the source table and checking for the continued presence of each record in the Calm API.
+
+## Proposal
+
+- The deletion watcher will be an ECS task within the existing calm adapter cluster.
+- It will run as a scheduled task at a frequency to be agreed with the collections information team.
```
warrd, line 16: We wouldn't necessarily have to update all records at the same time, it could be staggered over a week. For example one day we update all records IDs starting with `0-1`, the next day `2-3` etc. That would mean we would be say only making 50,000 requests in a day to the Calm API rather than 300,000.

jamieparkinson: I was going to go down this route but decided not to - what's the advantage of it? Would the Calm API "prefer" it? As I saw it, the average time to notice a deletion stays exactly the same but we suffer having to introduce state

jamieparkinson:
@kenoir unfortunately the Calm API is read-only 😞  besides, I feel like introducing a new process to another team is orders of magnitude more complex than any amount of code.

rfcs/032-calm-deletions/README.md
jamieparkinson: Ah, that does change things then! I'll have a look at what that might look like

rfcs/032-calm-deletions/README.md
warrd: Well if we don't want to DOS the Calm API we will artificially limit it in some way (via limiting number of concurrent workers, time in between requests etc), this is just another way of doing so. This suggestion does not introduce state any more than otherwise, it would be hard coded into the worker code to know what range a particular scan is.

But yeah, can see it might not be necessary to do this :shrug:

alexwlchan: Not necessarily, if you use the reindexer you can get something close to this.

DynamoDB automatically divides the table into segments when you do a Parallel Scan (https://alexwlchan.net/2018/08/parallel-scan-scanamo/). On Monday ask for segment 0 of 7, Tuesday 1 of 7, Wednesday 3 of 7, etc. You won't get contiguous ranges, but since the Calm IDs are UUIDs that's probably fine.

I'm still not sure it's worth doing – I'd rather blast through the table once a week and be done with it – but implementation isn't a bar to doing this if we want to.

rfcs/032-calm-deletions/README.md
alexwlchan: This is the case class that gets passed by the adapter/reindexer to the transformers: https://github.com/wellcomecollection/catalogue/blob/master/common/source_model/src/main/scala/weco/catalogue/source_model/SourcePayload.scala

It's not quite the DynamoDB row – it papers over some naming weirdness in the tables, e.g. the use of the word "payload" when it's really just a "location".

rfcs/032-calm-deletions/README.md
jamieparkinson: Doesn't that segment counter state have to live somewhere? Or are you saying the app looks at the clock and decides a segment based on that?

Anyway, I think I would regard this as something useful to come back to if we run into performance/throughput issues.

alexwlchan: > Doesn't that segment counter state have to live somewhere? Or are you saying the app looks at the clock and decides a segment based on that?

Pretty much – or whatever schedules it does.

> Anyway, I think I would regard this as something useful to come back to if we run into performance/throughput issues.

Agreed.

kenoir: I seem to remember Calm being unhappy about high volumes of requests.

warrd: > I seem to remember Calm being unhappy about high volumes of requests.

Yup I can't remember all the details but have a memory that ingesting the data into the adaptors initially took a fair amount of manual intervention and lots of things on DLQs

warrd: :point_up: I think this was made worse by the fact that there are some `ModifiedDate`'s with 10s of 1000s of records. As we are making requests by IDs for this functionality I suspect it will be less of an issue.

rfcs/032-calm-deletions/README.md
jamieparkinson: Hmm, I guess it's because this wouldn't be doing a reindex - but perhaps the reindexer is already far more overloaded than I thought it was.

alexwlchan (commented):
This approach looks much better. 👍 

Next up, probably worth considering how this data looks in VHS. I wonder if we can put the "isDeleted" flag directly in the DynamoDB table? To save the Calm transformer going out to S3 for a record that it'll delete 99% of.

jamieparkinson:
That optimisation would be difficult to implement because the shared `TransformerWorker`/`Transformer` traits used by the transfomers always fetch the full data first - I think adding an exception to that for the comparatively unusual case where this deleted flag is present is overkill. While the Miro data has that flag in Dynamo, the Sierra data doesn't, so we don't have a consistent precedent to follow.

alexwlchan:
> That optimisation would be difficult to implement because the shared TransformerWorker/Transformer traits used by the transfomers always fetch the full data first - I think adding an exception to that for the comparatively unusual case where this deleted flag is present is overkill. While the Miro data has that flag in Dynamo, the Sierra data doesn't, so we don't have a consistent precedent to follow.

I'm not sure I agree. (Again, this code has changed a lot recently so I can imagine it might have been more difficult in the past.)

The TransformerWorker implementations have to supply a way to go `Payload => SourceData`, but each transformer decides how to implement that. You could split the `CalmSourceData` type into a deleted/undeleted variant, which saves you having to synthesise a Calm API response for a non-existent record.

alicefuzier:
I think this is a sensible and relatively simple approach 👍

jamieparkinson:
@alexwlchan I've just realised that the deletion flag will _have_ to live in the Dynamo record (aka `CalmSourcePayload` because the reindexer will still be sending the deleted records, and we'll want to filter those out rather than redundantly checking for their existence every time

PR #45, opened 2021-01-29
RFC about API internal model versioning

alicefuzier:
Closes https://github.com/wellcomecollection/platform/issues/4998

Rendered version: https://github.com/wellcomecollection/docs/blob/api-internal-model-versioning/rfcs/033-api-internal-model-versioning/README.md

### DIFF
--- a/rfcs/033-api-internal-model-versioning/README.md
+++ b/rfcs/033-api-internal-model-versioning/README.md
@@ -13,8 +13,8 @@ can deploy the API and a reindex can take a few hours.
 ## Problem
 
 This means that:
--  If there is a change to the internal model and a susequent change to the API that is unrelated to the internal model change
-   (i.e. a chnage on search queries or on display serialization), the change to the API cannot be deployed until a reindex
+-  If there is a change to the internal model and a subsequent change to the API that is unrelated to the internal model change
+   (i.e. a change on search queries or on display serialization), the change to the API cannot be deployed until a reindex
    has finished successfully because of the internal model dependency.
 
 - If there are problems in the pipeline that delay a reindex, that means a deploy of the API is potentially delayed too.
--- a/rfcs/033-api-internal-model-versioning/README.md
+++ b/rfcs/033-api-internal-model-versioning/README.md
@@ -1,0 +1,78 @@
+# RFC 033: Api internal model versioning
+
+## Current status
+
+Currently our API and catalogue pipeline both live in our catalogue repo and have a dependency to the internal model module.
+
+This means that, if we need to make model changes, we can quickly make the corresponding changes in all services of
+the pipeline that depend on it and in the API at the same time.
+
+However, changes to the internal model that aren't backwards compatible require a reindex before we
+can deploy the API and a reindex can take a few hours.
+
+## Problem
+
+This means that:
+-  If there is a change to the internal model and a susequent change to the API that is unrelated to the internal model change
+   (i.e. a chnage on search queries or on display serialization), the change to the API cannot be deployed until a reindex
+   has finished successfully because of the internal model dependency.
+
+- If there are problems in the pipeline that delay a reindex, that means a deploy of the API is potentially delayed too.
+
+- If there is a bug in the prod API, fixing it means potentially waiting for a reindex or deploying from a patch.
+
+- We need integration tests that tell us if a build of the API is compatible with the currently available index and we
+  can't guarantee that what is in master is always deployable and that integration tests should always pass.
+
+## Proposed solution
+
+The proposed solution is:
+- for every build in master, release the internal model as an artifact with a new version
+  and publish it to our s3 respository in a similar way as to how we release artifacts in the scala-libs repo.
+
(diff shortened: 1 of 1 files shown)


rfcs/033-api-internal-model-versioning/README.md
```diff
@@ -0,0 +1,78 @@
...
+- We can split the API into its own repo which could allow us to think substantially differently about the API than
+  the pipeline and have a different build.
+
+
+## Potential problems with the proposed approach
+
+- It's difficult to know from an index what version of the internal model was used to populate it. Ideally we might want to include
+  the model version in the index name but because the pipeline services, including the ingestor, depend on the local internal_model,
+  they don't know the version, so it's not straightforward to do. This means that we could run into issues if we update the index tha API
+  is pointing to without updating the model dependency and viceversa.
```
alexwlchan, line 70: Couple of thoughts:

* The API starts up and discovers the index isn't compatible. Can we make it hard fail, so an ECS blue-green deployment will never allow a broken API to start?
* I wonder if we can copy the version into the catalogue pipeline code? You could put the version in the Scala, something like:

    ```scala
    object InternalModelVersion {
      val version = "1.2.3"
    }
    ```

    and whatever code is responsible for publishing the library to S3 also has to come in and edit this file.

rfcs/033-api-internal-model-versioning/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+- If we need to make changes to the API that require internal model changes, we need to wait on a new release of the internal model, before
+  we can start working on it.
```
alexwlchan, line 73: I don't think this is necessarily an issue – if we do this, I imagine we'd do pipeline and API patches separately.

The pipeline patch gets merged immediately, gets deployed and we reindex. That patch includes the internal model change, and when it gets merged we publish a new version.

That then allows you to write the API patch separately, but you don't merge and deploy it until the reindex is done and you have an index ready to go. That PR might be open for a while, but it's not merged until you're almost ready to deploy it.

rfcs/033-api-internal-model-versioning/README.md
```diff
@@ -0,0 +1,78 @@
...
+This means that, if we need to make model changes, we can quickly make the corresponding changes in all services of
+the pipeline that depend on it and in the API at the same time.
+
+However, changes to the internal model that aren't backwards compatible require a reindex before we
+can deploy the API and a reindex can take a few hours.
+
+## Problem
+
+This means that:
+-  If there is a change to the internal model and a susequent change to the API that is unrelated to the internal model change
```
alexwlchan, line 16: typo: ```suggestion
-  If there is a change to the internal model and a subsequent change to the API that is unrelated to the internal model change
```

File: rfcs/033-api-internal-model-versioning/README.md [hunk 4]
```diff
@@ -0,0 +1,78 @@
...
+   (i.e. a chnage on search queries or on display serialization), the change to the API cannot be deployed until a reindex
```
alexwlchan, line 17: typo:

```suggestion
   (i.e. a change on search queries or on display serialization), the change to the API cannot be deployed until a reindex
```

alexwlchan (commented):
Really good explanation. 👍

Seeing it written out in detail, it makes more sense now. I wasn't entirely convinced on this approach, but that's because I had a different idea of how this would work than you. Now it seems very sensible.

rfcs/033-api-internal-model-versioning/README.md
alicefuzier: Yeah, I don't think it's an issue but it is a difference compared to what we can do right now. For example, we might do all the pipeline related changes, then move on to the api changes and realise we forgot something on the pipeline side. That means reopening a PR, a new release etc. I think it probably doesn't add that much in terms of time lost and anyway the benefits outweigh the downsides but I thought it was woth mentioning it. 

Perhaps it could be worded better though

rfcs/033-api-internal-model-versioning/README.md
alicefuzier: > The API starts up and discovers the index isn't compatible. Can we make it hard fail, so an ECS blue-green deployment will never allow a broken API to start?

I think that's probably doable and very sensible

Re: model version & index name I was thinking if we could use the git hash for the internal model version instead of semver. We lose the distinction major minor, patch - which I think we get wrong very often anyway -  but we have a version that is shared between ingestor & internal model and possibly it could be easier to automate a process where the index name has the model version

alexwlchan: I like that; could we maybe also incorporate the build number from BuildKite?

e.g. `123-cfee1f61b673f1df660ed5eedebca32204e2980c`

Commit hashes are great for knowing where code came from, but lousy for comparisons. It feels like being able to order versions would be useful.

alicefuzier: ah yes, that sounds like a good idea

kenoir:
This makes sense.

kenoir (approved):

PR #46, opened 2021-02-02
Describe the requirements for our Location model

alexwlchan:
This is part of https://github.com/wellcomecollection/platform/issues/4996

Rendered version: https://github.com/wellcomecollection/docs/blob/location-location-location/rfcs/034-location_location_location/README.md

### DIFF
--- a/rfcs/034-location_location_location/README.md
+++ b/rfcs/034-location_location_location/README.md
@@ -299,6 +299,10 @@ Bibs and items from Sierra:
         <td>OnExhibition</td>
         <td>on exhibition</td>
       </tr>
+      <tr>
+        <td>Online</td>
+        <td>online</td>
+      </tr>
     </table>
 
     This heuristic allows us to create locations in our controlled type for **94.7% of all items**.
@@ -393,6 +397,60 @@ Bibs and items from Sierra:
     If the location code or name are unrecognised, we will skip adding a location to the holding.
 
     This heuristic allows us to create locations in our controlled type for **99.87% of all holdings**.
+
+    The unmapped Sierra holdings locations are as follows:
+
+    <table>
+      <tr>
+        <th>Sierra location</th>
+        <th># of items</th>
+      </tr>
+      <tr>
+        <td>gblip / 215 Information Point</td>
+        <td>15</td>
+      </tr>
+      <tr>
+        <td>multi</td>
+        <td>7</td>
+      </tr>
+      <tr>
+        <td>digi / Digitised Collections</td>
+        <td>3</td>
+      </tr>
+      <tr>
+        <td>wghig / Gallery</td>
+        <td>3</td>
+      </tr>
+      <tr>
+        <td>acql / Wellcome Library</td>
+        <td>2</td>
+      </tr>
+      <tr>
+        <td>wghir / Reading Room</td>
+        <td>2</td>
+      </tr>
+      <tr>
+        <td>wghxg / Folios</td>
+        <td>2</td>
+      </tr>
+      <tr>
+        <td>dept / Wellcome Trust departments</td>
+        <td>1</td>
+      </tr>
+      <tr>
+        <td>gibb / Wellcome Trust departments</td>
+        <td>1</td>
+      </tr>
+      <tr>
+        <td>orio / Asian Collections</td>
+        <td>1</td>
+      </tr>
+      <tr>
+        <td>wqrfe / Enquiry Line</td>
+        <td>1</td>
+      </tr>
+    </table>
+
     </details>
 
 *   We will add an optional `"shelfMark"` field to the PhysicalLocation model.
--- a/rfcs/034-location_location_location/README.md
+++ b/rfcs/034-location_location_location/README.md
@@ -102,13 +102,19 @@ A digitised item (a IIIF Presentation location):
 
 ## What should a location tell you?
 
-When somebody looks at an item, they want to know two things about the location:
+When somebody looks at an item, here's what they want to know about location:
 
 *   **Am I allowed to see it?**
     This is encoded by the *access conditions* field, which includes an *access status* (drawn from a fixed vocabulary: open, closed, requires-permission, etc.) and an *access description* (free-text, written by cataloguers).
 
 *   **If I'm allowed to see it, where can I find it?**
     For example: where it sits on the physical shelves within the library, or a particular URL, or a IIIF resource.
+
+*   **If I'm allowed to see it, can I use it?**
+    This is the *license* on the item.
+
+    Note: Our current model only supports a license on digital locations.
+    Is this correct, or do we need to support licenses on physical locations also?
 
 We need to support filtering and aggregating by location.
 For example, users should be able to find all the material that's openly available online (if they're working remotely), or all the material in the library (if they're in the building).
--- a/rfcs/034-location_location_location/README.md
+++ b/rfcs/034-location_location_location/README.md
@@ -137,7 +137,9 @@ These are the types of location our model needs to support:
 
     Do we need to expose this information for the requesting service?
 
-    -   Can something be requested faster if it's stored on-site?
+    -   We don't need to expose the full detail of our closed stores, but do we want to expose the difference between onsite/offsite?
+        The latter will take longer to retrieve, and may require a manual request to Library Enquiries.
+        Is this something that should be exposed by the Catalogue API, or only by the Requesting service?
 
 *   **A standalone IIIF image.**
 
--- a/rfcs/034-location_location_location/README.md
+++ b/rfcs/034-location_location_location/README.md
@@ -312,14 +312,74 @@ Bibs and items from Sierra:
     The unmapped Sierra locations are as follows:
 
     <table>
-      <tr><th>Sierra location</th><th># of items</th></tr>
-      <tr><td>bwith / bound in above</td><td>218</td></tr>
-      <tr><td>cwith / Contained in above</td><td>28</td></tr>
-      <tr><td>digit / Digitised Collections</td><td>10</td></tr>
-      <tr><td>wqrfe / Enquiry Line</td><td>2</td></tr>
-      <tr><td>wghig / Gallery</td><td>1</td></tr>
-      <tr><td>sgmip / sgmip</td><td>1</td></tr>
-      <tr><td>gblip / 215 Information Point</td><td>1</td></tr>
+      <tr>
+        <th>Sierra location</th>
+        <th># of items</th>
+        <th>examples</th>
+      </tr>
+      <tr>
+        <td>bwith / bound in above</td>
+        <td>218</td>
+        <td>
+          <a href="https://api.wellcomecollection.org/catalogue/v2/works?items.locations.locationType=bwith&include=items%2Cidentifiers">API examples</a> /
+          <a href="https://api.wellcomecollection.org/catalogue/v2/works/a2239muq?include=items,identifiers">a2239muq</a> /
+          <a href="https://search.wellcomelibrary.org/iii/encore/record/C__Rb1519298?lang=eng">b15192982</a>
+        </td>
+      </tr>
+      <tr>
+        <td>cwith / Contained in above</td>
+        <td>28</td>
+        <td>
+          <a href="https://api.wellcomecollection.org/catalogue/v2/works?items.locations.locationType=cwith&include=items%2Cidentifiers">API examples</a> /
+          <a href="https://api.wellcomecollection.org/catalogue/v2/works/a23admwy?include=items,identifiers">a23admwy</a> /
+          <a href="https://search.wellcomelibrary.org/iii/encore/record/C__Rb1986128?lang=eng">b1986128x</a>
+        </td>
+      </tr>
+      <tr>
+        <td>digit / Digitised Collections</td>
+        <td>10</td>
+        <td>
+          <a href="https://api.wellcomecollection.org/catalogue/v2/works?items.locations.locationType=digit&include=items%2Cidentifiers">API examples</a> /
+          <a href="https://api.wellcomecollection.org/catalogue/v2/works/dutxh9yu?include=items,identifiers">dutxh9yu</a> /
+          <a href="https://search.wellcomelibrary.org/iii/encore/record/C__Rb1077140?lang=eng">b10771402</a>
+        </td>
+      </tr>
+      <tr>
+        <td>wqrfe / Enquiry Line</td>
+        <td>2</td>
+        <td>
+          <a href="https://api.wellcomecollection.org/catalogue/v2/works?items.locations.locationType=wqrfe&include=items%2Cidentifiers">API examples</a> /
+          <a href="https://api.wellcomecollection.org/catalogue/v2/works/en3eymfx?include=items,identifiers">en3eymfx</a> /
+          <a href="https://search.wellcomelibrary.org/iii/encore/record/C__Rb1529652?lang=eng">b15296520</a>
+        </td>
+      </tr>
+      <tr>
+        <td>wghig / Gallery</td>
+        <td>1</td>
+        <td>
+          <a href="https://api.wellcomecollection.org/catalogue/v2/works?items.locations.locationType=wghig&include=items%2Cidentifiers">API examples</a> /
+          <a href="https://api.wellcomecollection.org/catalogue/v2/works/k3ee5bwq?include=items,identifiers">k3ee5bwq</a> /
+          <a href="https://search.wellcomelibrary.org/iii/encore/record/C__Rb2472458?lang=eng">b24724580</a>
+        </td>
+      </tr>
+      <tr>
+        <td>sgmip / sgmip</td>
+        <td>1</td>
+        <td>
+          <a href="https://api.wellcomecollection.org/catalogue/v2/works?items.locations.locationType=sgmip&include=items%2Cidentifiers">API examples</a> /
+          <a href="https://api.wellcomecollection.org/catalogue/v2/works/v5rcgfrg?include=items,identifiers">v5rcgfrg</a> /
+          <a href="https://search.wellcomelibrary.org/iii/encore/record/C__Rb1307568?lang=eng">b13075688</a>
+        </td>
+      </tr>
+      <tr>
+        <td>gblip / 215 Information Point</td>
+        <td>1</td>
+        <td>
+          <a href="https://api.wellcomecollection.org/catalogue/v2/works?items.locations.locationType=gblip&include=items%2Cidentifiers">API examples</a> /
+          <a href="https://api.wellcomecollection.org/catalogue/v2/works/z3hgbrbk?include=items,identifiers">v5rcgfrg</a> /
+          <a href="https://search.wellcomelibrary.org/iii/encore/record/C__Rb1497535?lang=eng">b1497535x</a>
+        </td>
+      </tr>
     </table>
 
     </details>
--- a/rfcs/034-location_location_location/README.md
+++ b/rfcs/034-location_location_location/README.md
@@ -115,9 +115,6 @@ When somebody looks at an item, here's what they want to know about location:
 *   **If I'm allowed to see it, can I use it?**
     This is the *license* on the item.
 
-    Note: Our current model only supports a license on digital locations.
-    Is this correct, or do we need to support licenses on physical locations also?
-
 We need to support filtering and aggregating by location.
 For example, users should be able to find all the material that's openly available online (if they're working remotely), or all the material in the library (if they're in the building).
 
@@ -229,3 +226,7 @@ Bibs and items from Sierra:
 
     -   `ByAppointment`, which the current implementation of the Catalogue API treats as synonymous with `Restricted`
     -   `TemporarilyUnavailable`, so we can distinguish between, say, material that's being digitised and will be available later, and material that has been de-accessioned and will never become available.
+
+*   Currently the `"license"` field only appears on digital locations.
+
+    Although we don't have any examples of it on physical locations yet, a license is sufficiently core that we'll make it part of the PhysicalLocation model.
--- a/rfcs/034-location_location_location/README.md
+++ b/rfcs/034-location_location_location/README.md
@@ -138,7 +138,6 @@ These are the types of location our model needs to support:
     Do we need to expose this information for the requesting service?
 
     -   Can something be requested faster if it's stored on-site?
-    -   Does the requesting service need to know the location of something when it tells Sierra "please ask somebody to fetch this item", or can it just provide Sierra the item ID?
 
 *   **A standalone IIIF image.**
 
--- a/rfcs/034-location_location_location/README.md
+++ b/rfcs/034-location_location_location/README.md
@@ -1,0 +1,121 @@
+# RFC 034: Modelling Locations in the Catalogue API
+
+As part of upcoming work to add journals and holdings to the Catalogue API, we need to expand the range of locations which are presented by the API.
+Additionally, we want to model locations in a format suitable for the requesting service.
+
+This RFC describes the types of location we want to model, and proposes how they will be returned in the catalogue API.
+
+**Current status:** this RFC outlines the problem, by describing the types of location we want to model.
+Please check if the information contained is accurate and complete.
+
+
+
+## The existing approach
+
+The Catalogue API contains *Works*, and Works contain *Items*.
+In turn, each Item can have *Locations*, which tell the user how they can get the item.
+We're not planning to change the Item model; we want to consider if the Location model is fit for purpose.
+
+<details>
+<summary>Examples of physical and digital locations in the existing model</summary>
+
+Physical location:
+
+```json
+{
+  "locationType": {
+    "id": "scmac",
+    "label": "Closed stores Arch. & MSS",
+    "type": "LocationType"
+  },
+  "label": "Closed stores Arch. & MSS",
+  "accessConditions": [
+    {
+      "status": {
+        "id": "open",
+        "label": "Open",
+        "type": "AccessStatus"
+      },
+      "terms": "The papers are available subject to the usual conditions of access to Archives and Manuscripts material.",
+      "type": "AccessCondition"
+    }
+  ],
+  "type": "PhysicalLocation"
+}
+```
+
+  Digital location:
+
+```json
+{
+  "locationType": {
+    "id": "iiif-image",
+    "label": "IIIF Image API",
+    "type": "LocationType"
+  },
+  "url": "https://iiif.wellcomecollection.org/image/A0000681.jpg/info.json",
+  "credit": "Royal Veterinary College",
+  "license": {
+    "id": "cc-by-nc",
+    "label": "Attribution-NonCommercial 4.0 International (CC BY-NC 4.0)",
+    "url": "https://creativecommons.org/licenses/by-nc/4.0/",
+    "type": "License"
+  },
+  "accessConditions": [],
+  "type": "DigitalLocation"
+}
+```
+
+</details>
+
+
+
+## What should a location tell you?
+
+When somebody looks at an item, they want to know two things about the location:
+
+*   **Am I allowed to see it?**
+    This is encoded by the *access conditions* field, which includes an *access status* (drawn from a fixed vocabulary: open, closed, requires-permission, etc.) and an *access description* (free-text, written by cataloguers).
+
+*   **If I'm allowed to see it, where can I find it?**
+    For example: where it sits on the physical shelves within the library, or a particular URL, or a IIIF resource.
+
+We need to support filtering and aggregating by location.
+For example, users should be able to find all the material that's openly available online (if they're working remotely), or all the material in the library (if they're in the building).
+
+
+
+## What locations do we need to support?
+
+These are the types of location our model needs to support:
+
+*   **Open shelves.**
+
+    This should include a *shelf location* (e.g. History of Medicine Collection) and a *shelfmark* (e.g. TW.AL) so that a user can find it.
+
+*   **Closed stores.**
+
+    Although the MARC records include a shelf location and a shelfmark, the Collections team would prefer this not to be exposed publicly.
+
+    Do we need to expose this information for the requesting service?
+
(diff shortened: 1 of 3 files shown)


pollecuttn:
@alexwlchan Re External websites or URLs: when in the library some of these can be accessed without the user having to login because access is by IP address. Or at least this definitely used to be the case. It usually still required the proxy URL in the record. Library Systems Support will be more up-to-date on this than I am.

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,133 @@
+# RFC 034: Modelling Locations in the Catalogue API
+
+As part of upcoming work to add journals and holdings to the Catalogue API, we need to expand the range of locations which are presented by the API.
```
jtweed, line 3: Nit: journals are already there, but we don't describe the holdings or how to access online resources.

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,133 @@
...
+      },
+      "terms": "The papers are available subject to the usual conditions of access to Archives and Manuscripts material.",
+      "type": "AccessCondition"
+    }
+  ],
+  "type": "PhysicalLocation"
+}
+```
+
+  Digital location:
```
jtweed, line 47: I think it's worth including a digitised item location here, instead or in addition to a Miro image. Ie, something that is linking to a IIIF Presentation API manifest.

jtweed (changes requested):
There's another major piece that's missing here in addition to the specific comments. We need to know which locations are open shelves, closed stores and online resources.

From my previous looking at this, the best way seemed to be to use the four letter location codes on the bib and ignore the five letter location codes on the items entirely.

The four letter code is available in the `locations` property on bibs in the Sierra API, which is populated from fixed field 026.

My initial, probably incomplete, mapping attempt from a few months back looked like:

| ID            | Label         | Sierra mapping |
| ------------- | ------------- | -------------- |
| open-shelves  | Open shelves  | whml           |
| closed-stores | Closed stores | stax, arch     |
| online        | Online        | elro, digi     |

With any other Sierra bib location code ignored.

rfcs/034-location_location_location/README.md
alexwlchan: Added.

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,133 @@
...
+}
+```
+
+</details>
+
+
+
+## What should a location tell you?
+
+When somebody looks at an item, they want to know two things about the location:
```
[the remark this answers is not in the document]
alexwlchan: Added. This includes a question I don't know the answer to:

> Our current model only supports a license on digital locations. Is this correct, or do we need to support licenses on physical locations also?

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,133 @@
+# RFC 034: Modelling Locations in the Catalogue API
+
+As part of upcoming work to add journals and holdings to the Catalogue API, we need to expand the range of locations which are presented by the API.
+Additionally, we want to model locations in a format suitable for the requesting service.
```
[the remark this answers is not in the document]
alexwlchan: I've tried to clarify this – I don't think we'd use all the same data, but my impression is that we want a blob of API JSON that looks broadly the same. Is that right, or am I still missing something here?

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+    This should include a *shelf location* (e.g. History of Medicine Collection) and a *shelfmark* (e.g. TW.AL) so that a user can find it.
+
+*   **Closed stores.**
+
+    Although the MARC records include a shelf location and a shelfmark, the Collections team would prefer this not to be exposed publicly.
+
+    Do we need to expose this information for the requesting service?
+
+    -   Can something be requested faster if it's stored on-site?
```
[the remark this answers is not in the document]
alexwlchan: Follow-up question (also in the body of the RFC):

> We don't need to expose the full detail of our closed stores, but do we want to expose the difference between onsite/offsite? The latter will take longer to retrieve, and may require a manual request to Library Enquiries. Is this something that should be exposed by the Catalogue API, or only by the Requesting service?

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,133 @@
...
+    For example, the digitised books.
+    As with a standalone IIIF image, this should include a *URL*, and optionally a *license* and *credit line*.
+
+*   **External websites or URLs.**
+
+    For example, a journal with online access.
+    This should include a *URL* (to access the resource), and optional *link text* (e.g. View this e-book).
+
+    If it's a URL with subscription access, users can only view the journal by accessing a URL like `https://{proxyname}.catalogue.wellcomelibrary.org` and logging in to their Library account.
+    This means we need an access status of `login-required`, although this isn't explicitly communicated in the holdings records -- we'd have to infer it from the URL.
```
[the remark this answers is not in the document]
alexwlchan: [nod]

This is sloppy wording on my part – I was wondering whether we'd add `login-required` to the controlled vocabulary that makes up our AccessStatus model (alongside `open`, `restricted`, `permission-required`, etc.), and ended up mentioning it here. It sounds like it doesn't apply though, so I've removed this suggestion.

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+*   **On exhibition.**
+
+    This should include the *name of the exhibition* and the *due date* when the item will be returning to the library.
+
+    For an example of this, see <https://wellcomecollection.org/works/azrhju85>.
+    There might be a free-form text description that lives separately in the "notes" field.
+
+There are a number of other infrequently used location codes (e.g. "offsite", "at digitisation", "Rare Materials Room") that we currently present in the API.
```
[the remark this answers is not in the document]
alexwlchan: Agree it's not worth parsing specific locations and dates out.

What we do currently is look in the "location" field on the item record, which returns `exres` for "on-exhibition", then present the free text description as an ExhibitionNote from MARC 585.

rfcs/034-location_location_location/README.md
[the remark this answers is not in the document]
jtweed: I think it's core enough that there should be an option to include it on any kind of location, even though we're unlikely to have it for physical locations in practice.

rfcs/034-location_location_location/README.md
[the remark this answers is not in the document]
jtweed: No, you're absolutely correct. Just wanted to make sure that we know that eg things like item statuses are not in scope for this as they will come from the Stacks API.

rfcs/034-location_location_location/README.md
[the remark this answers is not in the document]
jtweed: Good question. Requesting should not be possible in the first instance for these items, but we may add it later as a manual request rather than a hold in Sierra. Whether it should be here is more to do with how likely it is to change and where exactly it sits in the model alongside other access methods.

rfcs/034-location_location_location/README.md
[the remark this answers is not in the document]
jtweed: Ah ok, we should do that then and make exres one of the exceptional five letter location codes that we map explicitly.

amme2:
> There's another major piece that's missing here in addition to the specific comments. We need to know which locations are open shelves, closed stores and online resources.
> 
> From my previous looking at this, the best way seemed to be to use the four letter location codes on the bib and ignore the five letter location codes on the items entirely.
> 
> The four letter code is available in the `locations` property on bibs in the Sierra API, which is populated from fixed field 026.
> 
> My initial, probably incomplete, mapping attempt from a few months back looked like:
> 
> Location type	Sierra mapping
> Open shelves	whml
> Closed stores	stax, arch
> Online	elro, digi
> With any other Sierra bib location code ignored.

I don't think you can entirely ignore item locations (unfortunately!), because some bibs will have both an open shelf and closed stacks 4 letter code (and I guess potentially online code too), and it's only by looking at the item that you can distinguish between them e.g. b12047363 has `stax,whml,homc` in `LOCATION` on the bib. 1 copy is on open shelves, the other would need to be requested for retrieval from the stacks: https://search.wellcomelibrary.org/iii/encore/record/C__Rb1204736?lang=eng

jtweed:
> I don't think you can entirely ignore item locations (unfortunately!), because some bibs will have both an open shelf and closed stacks 4 letter code (and I guess potentially online code too), and it's only by looking at the item that you can distinguish between them e.g. b12047363 has `stax,whml,homc` in `LOCATION` on the bib. 1 copy is on open shelves, the other would need to be requested for retrieval from the stacks: https://search.wellcomelibrary.org/iii/encore/record/C__Rb1204736?lang=eng

Thanks that's a good example, looks like we need to dig a bit deeper then.

alexwlchan:
> I don't think you can entirely ignore item locations (unfortunately!), because some bibs will have both an open shelf and closed stacks 4 letter code (and I guess potentially online code too), and it's only by looking at the item that you can distinguish between them e.g. b12047363 has stax,whml,homc in LOCATION on the bib. 1 copy is on open shelves, the other would need to be requested for retrieval from the stacks: https://search.wellcomelibrary.org/iii/encore/record/C__Rb1204736?lang=eng

Ah, that's unfortunate.

What if we were to keep reading the locations from the item records, but stop passing through the exact LocationType from the Sierra location codes?

e.g. the current item on that example is:

```json
{
  "locationType": {
    "id": "sgwel",
    "label": "Closed stores Wellcome Coll.",
    "type": "LocationType"
  },
  "label": "Closed stores Wellcome Coll.",
  "accessConditions": [],
  "type": "PhysicalLocation"
}
```

We could do something like:

```json
{
  "locationType": {
    "id": "closed-stores",
    "label": "Closed stores",
    "type": "LocationType"
  },
  "label": "Closed stores",
  "accessConditions": [],
  "type": "PhysicalLocation"
}
```

based on the presence of "Closed stores" in the label from the Sierra API.

* We're getting the most specific form of location, from the item records
* We can handle bibs that have multiple items
* We're not exposing the layout of our closed stores publicly
* The platform no longer has to keep as close a track of all the Sierra location codes. We get the code and label from the Sierra API, and if we transform that into `closed-stores` then we keep our own, much more limited set to look after.

I still haven't worked out what to do with the long tail, but I want to poke around at the records some more to see if that's possible.

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+These are the types of location our model needs to support:
+
+*   **Open shelves.**
+
+    This should include a *shelf location* (e.g. History of Medicine Collection) and a *shelfmark* (e.g. TW.AL) so that a user can find it.
+
+*   **Closed stores.**
+
+    Although the MARC records include a shelf location and a shelfmark, the Collections team would prefer this not to be exposed publicly.
```
amme2, line 136: I may have overstated this preference, not helped by the indiscriminate use of 'shelfmark' in Wellcome practice to refer to many different things:

- An item identifier e.g. EBP/A/1234, EPH4 - this _does_ need to be exposed publicly
- A non-classified shelfmark, usually a number in sequence with a prefix for the series e.g. M1234, K1234 - this _could_ be exposed publicly, but I'd argue is meaningless to most users
- A broadly classified shelfmark e.g. AFW.AA1 - as above, no harm in exposing but meaningless (also to staff in some instances...)
- A number within a sequence which is only interpretable in combination with shelf location e.g. 3
- The concatenated combination of two separate fields in Calm e.g. PP/SUL/A/3/4/1:Box25 - this _doesn't_ need to be exposed, and personally I think its confusing to users (and potentially a security risk) to do so

There's also:
- no item shelfmark (all visual material) where equivalent data is in a free-text field _so that_ it can't be exposed publicly

This may not be an exhaustive list. Now may not be the best time to tackle all of this either, so perhaps you should look at it the other way round - expose `949` by default with exceptions where user-relevant information is already available elsewhere in the model?

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+    Do we need to expose this information for the requesting service?
+
+    -   We don't need to expose the full detail of our closed stores, but do we want to expose the difference between onsite/offsite?
+        The latter will take longer to retrieve, and may require a manual request to Library Enquiries.
+        Is this something that should be exposed by the Catalogue API, or only by the Requesting service?
```
amme2, line 142: Where material is not held onsite, this needs to be clear on catalogue records (in order to manage user expectations) but I'm not sure whether that means in the catalogue API. 

Related, IMHO it would be good to work towards the eradication of manual requests so that we can provide (as Accreditation requires) a proper audit trail of all requests.

rfcs/034-location_location_location/README.md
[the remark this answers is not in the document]
amme2: That's true for 'at digitisation' but @alexwlchan is correct that `hgser` (offsite) and `wghip` / `wqrfp` (rare materials room - don't ask me why there are two) are locations, not OPACMSG. The use of OPACMSG in combination with these is not consistent...

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,218 @@
...
+Bibs and items from Sierra:
+
+*   We create a physical location whose location type comes from the `"location"` field on the Sierra API response for the item.
+*   The access conditions are based on the varfields with MARC tag 506:
+
+    -   If the first indicator is 0, the access status is "Open"
+    -   We look at subfield $a for a free-text description of the access conditions
+    -   We look at subfield $f for the standardised open/closed/restricted terminology
+
+    Note: if these three values are inconsistent, we do not set an AccessStatus from our controlled vocabulary.
```
amme2, line 218: Collections Information would happily sort those out if you can send a list over sometime.

amme2:
> What if we were to keep reading the locations from the item records, but stop passing through the exact LocationType from the Sierra location codes?
> 
Sadly I fear my comments on closed stacks shelfmarks above may have scuppered this rather neat solution, unless in the cases where the shelfmarks are effectively item identifiers posing under another name you can somehow treat them as `"IdentiferType"` instead of `"LocationType"`?

> I still haven't worked out what to do with the long tail, but I want to poke around at the records some more to see if that's possible.

The long tail is one of the main reasons I wanted to comment on this RFC. The location codes are such a mess, and so out of date. And so muddled, conceptually. Let me know if there's anything Coll.Info. can do to help with this, even if not immediately.

rfcs/034-location_location_location/README.md
alexwlchan: I don't think there are any, at least not right now – if there were, I'd have told you!

The code will drop a warning if this happens, but on the latest reindex I didn't see any.

rfcs/034-location_location_location/README.md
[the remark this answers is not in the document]
amme2: I'd look at AV material for examples of this in relation to physical locations (I don't know for certain that there are any, but seems highly likely)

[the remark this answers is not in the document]
alexwlchan: I've added a note about this in a new "Proposed changes" section.

rfcs/034-location_location_location/README.md
alexwlchan: So I think I understand what you're suggesting:

* We have a place to put identifiers, which are usually pulled from other places – although I can't find EPB/A/1234 or EPH4 on wl.org or wc.org. Are those made-up identifiers, or has something got lost?
* The other four instances of "shelfmark" are things that are either meaningless or dangerous to expose on closed shelves, so either way let's not.

What's in MARC 949? OCLC says it's reserved for local use – what do Wellcome use it for?

amme2: It is indeed a shelfmark. The problem is more around how inconsistently it has been used. 

EPH4 is https://search.wellcomelibrary.org/iii/encore/record/C__Rb1538878?lang=eng
EPB/A/1234 I made up and is typically therefore missing from the sequence, but here's EPB/A/1235 https://search.wellcomelibrary.org/iii/encore/record/C__Rb2139241?lang=eng

rfcs/034-location_location_location/README.md
alexwlchan: Well I sound daft now… just found one. A list will drop out as part of this work.

rfcs/034-location_location_location/README.md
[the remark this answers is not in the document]
alexwlchan: For the catalogue API I'm going to make this "Closed stores", because it's quite a bit easier, and we can return to the "how do you request this thing" question for the requesting service.

Particularly if we want to channel all of these things through online requesting, offsite or not, let's not perpetuate the distinction right now.

rfcs/034-location_location_location/README.md
alexwlchan: As above, I'm not going to expose this distinction in the external catalogue API right now. We can revisit it when we build out requesting.

alexwlchan:
I've had a look through the data, and come up with a rudimentary heuristic for mapping the Sierra location codes to a set of controlled types.

There are data quality issues that Collections Information might want to fix at some point, but none of them are urgent or block this work (e.g. a handful of record have the location "OBSOLETE Closed Stores XYZ", but the proposed heuristic will still recognise that as "Closed Stores").

I've updated the RFC with an explanation of the heuristic, and some examples of JSON from the updated Location models. This approach covers over 99.5% of all items and holdings.

I think it's ~1 day of dev work, and another day to run a reindex.

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+    This is an initial suggestion for the list of types:
+
+    *   ClosedStores
+    *   OpenShelves
+    *   IIIFPresentationAPI
+    *   IIIFImageAPI
+    *   OnExhibition
+    *   OnlineResource
+    *   Conservation
```
jtweed, line 254: I'm not sure about this one. It's not a permanent state or one we have end for, like the others. Does it also apply to things that are at digitisation? Is this one is more of an availability status than a location type? It might make more sense to treat this the same as if it's in closed stores, because to a library user it effectively is.

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,403 @@
...
+      </tr>
+    </table>
+
+    This heuristic allows us to create locations in our controlled type for **94.7% of all items**.
+
+    Some Sierra items have a location like "as above" or "bound in above".
+    For these, we could apply an additional heuristic: if every other item has the same location (e.g. every other item is ClosedStores), then we give this item the same location.
+    Otherwise, we skip adding a Location to that item until we can come back and write more sophisticated logic (e.g. matching on shelfmark).
+
+    This additional heuristic would allow us to create locations in our controlled type for **99.97% of all items**.
```
jtweed, line 308: We do need to handle these like this, same with contained in.

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+    The unmapped Sierra locations are as follows:
+
+    <table>
+      <tr><th>Sierra location</th><th># of items</th></tr>
+      <tr><td>bwith / bound in above</td><td>218</td></tr>
+      <tr><td>cwith / Contained in above</td><td>28</td></tr>
+      <tr><td>digit / Digitised Collections</td><td>10</td></tr>
```
jtweed, line 316: Bizarre - have you got an example?

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,403 @@
...
+      <tr><td>wqrfe / Enquiry Line</td><td>2</td></tr>
+      <tr><td>temp1 / At Digitisation</td><td>1</td></tr>
+      <tr><td>wghig / Gallery</td><td>1</td></tr>
+      <tr><td>sgmip / sgmip</td><td>1</td></tr>
+      <tr><td>gblip / 215 Information Point</td><td>1</td></tr>
+    </table>
```
jtweed, line 322: Indeed examples of all of these would be good, bar _At Digitisation_. That effectively means it's currently unavailable and could map to closed stores along with _At conservation_.

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+    <details>
+    <summary>Proposed heuristic for matching Sierra holdings locations to controlled types</summary>
+
+    For Sierra holdings, the API only returns a location *code* (e.g. stax) via fixed field 40.
+
+    We will look up the location code to find the location name, then apply a similar rule to that which we apply for items.
+    If the location code or name are unrecognised, we will skip adding a location to the holding.
+
+    This heuristic allows us to create locations in our controlled type for **99.87% of all holdings**.
```
jtweed, line 334: What are some examples that it doesn't cover?

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,403 @@
...
+    </details>
+
+*   We will add optional `"shelfLocation"` and `"shelfmark"` fields to the PhysicalLocation model.
```
jtweed, line 337: I have never known what to call the open shelves locations. It's currently labelled as _Location_ and the site talks about locations within the library. But `PhysicalLocation` having a `location` property is a bit weird. It's not necessarily on a shelf, of course. But everything does have a shelfmark...

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+    These will only be populated if the locationType is `OpenShelves`.
+    The shelfLocation will be drawn from the location name in Sierra (e.g. "History of Medicine").
+    The shelfmark will be drawn from the callNumber field in Sierra (e.g. "/HIS").
+
+    For item records, these will be populated from the *location.name* and *callNumber* fields from the Sierra API response, respectively.
```
jtweed, line 343: Actually this might answer my comment above. Could we just call it a location name? Or possibly the thing that would be most consistent within the Catalogue API would be to use a `label` property on `Location` that contains the location name?

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,403 @@
...
+    This gives us a good initial implementation, and we can refine it later if necessary.
+
+    For holdings records, only the `"shelfLocation"` will be populated.
+    The holdings records don't have enough information to populate a shelfMark.
+
+    In these cases, we will concatenate the shelfLocation and shelfmark to create the `"label"`.
```
jtweed, line 349: Ha I see you just described it as _label_ here, which might mean that is the right direction.

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,403 @@
...
+        "type": "Location"
+      }
+      ```
+
+      An item on open shelves:
+
+      ```json
+      {
+        "id": "i1000011",
+        "label": "Medical Collection WP300 1902C96c",
```
jtweed, line 374: So I'm really proposing that this just says "Medical Collection"

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,403 @@
...
+        "shelfLocation": "Medical Collection",
```
jtweed, line 375: This is removed

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,403 @@
...
+        "shelfmark": "WP300 1902C96c",
```
jtweed, line 376: And this remains like this. It's up to the front end how to render and they'll want them separate.

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,403 @@
...
+      }
+      ```
+
+      An online journal resource:
+
+      ```json
+      {
+        "id": "h1234567",
+        "url": "https://journals.example.org/login",
+        "link_text": "View this journal",
```
jtweed, line 392: Arguably a `label` property could be reused here, but I'm happy to add another more specific one.

rfcs/034-location_location_location/README.md
alexwlchan: We already present these "shelfmarks" on wellcomelibrary.org, e.g. https://catalogue.wellcomelibrary.org/record=b1818611

As such, I don't think there's any additional harm in exposing them for now. I've updated the RFC to say:

> For now, this will be populated on all "PhysicalLocation" records, including ClosedStores. We may choose to remove it from some locations later, after appropriate discussions with Collections Information to API still includes all the identifiers and information required by users.

rfcs/034-location_location_location/README.md
alexwlchan: I've collapsed this into "ClosedStores with access status TemporarilyUnavailable", and rolled in digitisation as well.

rfcs/034-location_location_location/README.md
alexwlchan: Okay, I'll make sure we include this in the initial implementation.

rfcs/034-location_location_location/README.md
alexwlchan: You want me to find _examples_? But it'll take so long, if only we had some way of filtering the locations… 😉 

I've added links to specific examples to the table.

rfcs/034-location_location_location/README.md
alexwlchan: Done, and digitisation is now in the mapping alongside conservation.

rfcs/034-location_location_location/README.md
alexwlchan: Added.

rfcs/034-location_location_location/README.md
alexwlchan: I've gone through and renamed `shelfLocation` to `label`.

rfcs/034-location_location_location/README.md
alexwlchan: Yeah, I went back and forth on this. In the end, I made it a separate property because the label feels like something that should stand alone (e.g. for an aggregation), whereas this text doesn't make sense without the accompanying URL.

alexwlchan:
Thanks for the comments @jtweed; I think I've addressed them all and this is ready for another look.

🤞 we're getting close!

rfcs/034-location_location_location/README.md
jtweed: Makes sense

rfcs/034-location_location_location/README.md
```diff
@@ -0,0 +1,520 @@
...
+    <summary>Proposed heuristic for matching Sierra holdings locations to controlled types</summary>
+
+    For Sierra holdings, the API only returns a location *code* (e.g. stax) via fixed field 40.
+
+    We will look up the location code to find the location name, then apply a similar rule to that which we apply for items.
+    If the location code or name are unrecognised, we will skip adding a location to the holding.
+
+    This heuristic allows us to create locations in our controlled type for **99.87% of all holdings**.
+
+    The unmapped Sierra holdings locations are as follows:
```
jtweed, line 401: I think some of these, like gblipm/wghir are kinds of open shelves with the eg "Reading Room" as the location label.. But I'm sure @amme2 will be able to give you a steer for each of them.

jtweed (approved):

rfcs/034-location_location_location/README.md
alexwlchan: It should be fairly easy to add new values to this list as necessary.

alexwlchan:
Thanks for all the comments @jtweed and @amme2. I'm going to merge this as-is, and we can come back and tweak it later, but it's good enough for me to be able to start implementing these changes.

rfcs/034-location_location_location/README.md
amme2: Yes but they're also almost entirely oversights which need clearing up e.g. 'gallery' is the reading room gallery but there's def more than 1 book up there - must be another code in use! If you can send over a list when you get a chance, I'll get them sorted out. 

The bound in / contained in above combination is a whole other kettle of piranhas though, as you no doubt already know.

rfcs/034-location_location_location/README.md
amme2: Sorry, managed to post this in response to the wrong bit of the thread:

> Yes but they're also almost entirely oversights which need clearing up e.g. 'gallery' is the reading room gallery but there's def more than 1 book up there - must be another code in use! If you can send over a list when you get a chance, I'll get them sorted out.
> 
> The bound in / contained in above combination is a whole other kettle of piranhas though, as you no doubt already know.

I don't quite follow your proposed heuristic for dealing with bound in/contained in. I know there's an issue in ordering the response from the Sierra API but stuff isn't typically catalogued in physical arrangement order. Why would it be appropriate to assume every other item has the same shelf location? Maybe I've misunderstood something.

alexwlchan: (I haven't implemented this heuristic yet. Currently items that say "in above" won't have any location attached.)

This is based on a possibly faulty assumption that "in above" is going to refer to the location of another item attached to the same bib – and if we can unambiguously say what that location would be, we can use it.

Is "in above" going to refer to something completely different?

Possibly an example would be helpful?

---

Suppose we have a bib with items like:

```
i102 Closed stores MS&AB
i101 Bound in above
i104 Closed stores MS&AB
i103 Bound in above
```

but because of problems getting the order from the Sierra API, we get something like:

```
i101 Bound in above
i102 Closed stores MS&AB
i103 Bound in above
i104 Closed stores MS&AB
```

Because `Closed stores MS&AB` is the only non "in above" location in this list, we can assume that's what the "in above" refers to. When we create locations for those items in the Catalogue API, we'll use "Closed stores MS&AB".

On the other hand, suppose we got something like:

```
i101 Bound in above
i102 Closed stores MS&AB
i103 Bound in above
i104 Open shelves
```

There's no good way for us to know whether the "in above" is attached to the Closed stores or Open shelves, so we wouldn't attach any location to items i101 or i103.

amme2: Ah yes, that makes sense - sorry I'd forgotten you were looking at bibs first (two days on leave...).

alexwlchan: No worries, always good to double check!

PR #50, opened 2021-03-02
RFC 036: How are we going to model holdings data?

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/4992

This is meant to be enough for us to build an initial prototype, which we'll likely iterate on once the data is in the API/visible on wellcomecollection.org behind a toggle.


rfcs/036-holdings-records/README.md
```diff
@@ -0,0 +1,261 @@
...
+
+-   Two items titled *Volume 1* and *Volume 2*, and a holdings record titled *Missing volume 3*.
+    Here the holdings record is even more important, but there's no way to encode the absence of an item.
+
+-   A run of journals, with each issue itemised, and a holdings record titled *100 issues, 1957–1972*.
+    The holdings record can provide useful summary information that isn't easily observed from the items.
+
+### Creating a separate Holdings type
+
+Rather than reuse the `Item` type, we will create a new `Holdings` type with the same fields (at least, initially).
```
jtweed, line 109: `Holding`?

rfcs/036-holdings-records/README.md
```diff
@@ -0,0 +1,261 @@
...
+This will appear as a list on a Work called `"holdings"`.
+
+We will populate it as follows:
+
+*   The `title` will come from the concatenated contents of MARC field 866 subfields $a and $z.
```
jtweed, line 114: I wonder if we should structure a little bit more than this, but it does depend on the data. Eg, should be have `description`, `note`, `enumeration[]`? Or is that overkill for what is usually there. I'm thinking about leaving just enough structure there so that the front end doesn't need to understand the MARC, but can choose how to render large lists.

rfcs/036-holdings-records/README.md
```diff
@@ -0,0 +1,261 @@
...
+    We will append the label from 853 and 863 (newline-separated), one per 853/863 pair.
+    (We might need to split this out if it becomes unwieldy.)
+
+*   The `locations` list will contain:
+
+    *   A digital location using field 856 if the location in fixed field 40 is `elro` (online)
+    *   A physical location using fixed field 40 for the location type/label, and 949 for the shelfmark.
+
+*   For now, we won't add any identifiers.
+    We could add them as h-prefixed Sierra IDs, but I don't see much use of those identifiers -- usually people use the b number when discussing the associated holdings record.
```
jtweed, line 124: True, though we do for items, so it possibly makes sense to treat these in the same way.

rfcs/036-holdings-records/README.md
```diff
@@ -0,0 +1,261 @@
...
+Rather than reuse the `Item` type, we will create a new `Holdings` type with the same fields (at least, initially).
+This will appear as a list on a Work called `"holdings"`.
+
+We will populate it as follows:
+
+*   The `title` will come from the concatenated contents of MARC field 866 subfields $a and $z.
+    We will append the label from 853 and 863 (newline-separated), one per 853/863 pair.
+    (We might need to split this out if it becomes unwieldy.)
+
+*   The `locations` list will contain:
```
jtweed, line 118: I wonder how this displays in practice. Is every holding location also an item location? If so, can we drop and just have items as the sole way of actually locating something?

rfcs/036-holdings-records/README.md
```diff
@@ -0,0 +1,261 @@
...
+
+```json
+{
+  "type": "Work",
+  "id": "esstapc9",
+  "holdings": [
+    {
+      "identifiers": [
+        {
+          "identifierType": {"id": "sierra-system-number", "label": "Sierra system number", "type": "IdentifierType"},
```
jtweed, line 149: Ah you did put it in here, or do you mean just not including it again with an h prefix?

rfcs/036-holdings-records/README.md
```diff
@@ -0,0 +1,261 @@
...
+          "locationType": {
+            "id": "closed-stores",
+            "label": "Closed stores",
+            "type": "LocationType"
+          },
+          "label": "Closed stores",
+          "type": "Location"
+        }
+      ],
+      "type": "Holdings"
```
jtweed, line 165: As above, if `holdings` is a list, this should probably be singular, like `Item`. Have you seen multiple holdings in the data? It would be good to see an example of that to see how things split across multiple vs an enumeration within.

rfcs/036-holdings-records/README.md
```diff
@@ -0,0 +1,261 @@
...
+    }
+  ],
+  …
+}
+```
+
+</details>
+
+<details>
+<summary>A bib with a single electronic holdings record</summary>
```
jtweed, line 175: This really is the situation where I wonder whether it's better to create an implicit Item like we do for Miro. The holdings tell us nothing useful and just create another situation the front end has to deal with to display online access.

rfcs/036-holdings-records/README.md
```diff
@@ -0,0 +1,261 @@
...
+  ],
+  …
+}
+```
+
+</details>
+
+<details>
+<summary>A bib with multiple physical holdings records</summary>
+
```
jtweed, line 218: I need to read whole pull requests before I add comments.... It's interesting here that there is very little data on each holding. Is that the norm, ie are things like enumeration quite uncommon? Again I do wonder about the utility of including the locations here, as surely they are on the items and it's redundant. Ie can we keep holdings as a summary of, um, holdings and then items as the sole locatable object.

rfcs/036-holdings-records/README.md
alexwlchan: This is a mistake – I thought about putting it here and apparently wrote the JSON, then thought better of it.

rfcs/036-holdings-records/README.md
alexwlchan: This is a bit of a weird one – I don't really understand why, but the term for a single holdings record is “a holdings”. If you listen to the Collections team, they'll talk about "bib records" and "item records" and "holdings records". I don't really see people talk about a singular "holding", hence this choice of name.

e.g. https://techdocs.iii.com/sierraapi/Content/zReference/resources.htm

rfcs/036-holdings-records/README.md
alexwlchan: I went back and had another look at this; in one case there are **86** parts on a single record, so this isn't going to fly.

We aren't particularly consistent about using subfields $a and $z, but for now I've split them out as separate. They can always be combined in the front-end, and making the distinction explicit might improve the cataloguing guidelines.

rfcs/036-holdings-records/README.md
alexwlchan: I've added a bit more about this to the RFC.

With items, the identifier to use is unambiguous – add a check digit, add an `i` prefix. With holdings, it's a bit less clear – I've suggested `h` because these are holdings records, but I've seen at least one person talk about these with a `c` prefix for "checkin records".

We also use the item identifier in certain contexts (e.g. to link to a digitised item page), and it's less obvious to me that we'd use a holdings identifier for anything. AIUI, these records are typically looked up in Sierra by the bib number, not the holdings number.

I'm not opposed to adding these identifiers if there's a good use case, and we can nail down the prefix, but my default position is to omit them in v1.

rfcs/036-holdings-records/README.md
alexwlchan: I didn't mean to put this in! Oops, removing.

rfcs/036-holdings-records/README.md
alexwlchan: See below!

rfcs/036-holdings-records/README.md
alexwlchan: Ah yeah, we discussed this, and then I completely forgot. 🙈 

I've amended the RFC to reflect this.

rfcs/036-holdings-records/README.md
alexwlchan: What do you do on Works that have a holdings but not any items? Presumably we do want location information there?

rfcs/036-holdings-records/README.md
jtweed: I think this might be because "The Holdings object returns a list of holdings records." The docs then go on to say things like: "The links to the bib records associated with this holding.".

I'm not massively concerned ether way, but it might be good to see how a single holding is described within the Sierra app itself.

rfcs/036-holdings-records/README.md
jtweed: Certainly wouldn't want to invent a prefix, so let's leave out if we're not sure.

rfcs/036-holdings-records/README.md
jtweed: Can we create an unidentified item like we do for Miro?

alexwlchan: I don't think so.

For some of them, we will, because they have a digital location so we'll create the implicit item.

But we do have some records which:

1. Have a physical location on the holdings
2. Don't have any items
3. Have an enumeration (of volumes/issues) that we can't neatly fit onto the `Item` type

e.g. https://search.wellcomelibrary.org/iii/encore/record/C__Rb1311372?lang=eng

So unfortunately I think we really do need some location data here, even if it's not ideal.

Since a holdings record only has one location, we could reduce this down to `"location"` or `"locationType"`, and guarantee its presence, rather than proliferating ever more lists.

jtweed: That's annoying. Are the locations simpler on holdings? Looks that way from that example? If so, does it even make sense to plug in the full location model? Or can we have strings not things here? It's not particularly useful saying something is on open shelves and then not saying where...

jtweed (approved):

rfcs/036-holdings-records/README.md
alexwlchan: The locations tend to be simpler on holdings, yes – but we do get a shelfmark (field 949) and a general location code (e.g. `stax`, `whml`).

Let's see what the data looks like when we actually run through it.

(It looks like something is busted in the reporting cluster, because you can't see fixed field 40.)

PR #51, opened 2021-03-18
Add RFC for API principles/expectations

jamieparkinson:
This is intended to be a data-agnostic specification of how we expect our API to work, particularly regarding filtering and aggregations. Formatted file here: https://github.com/wellcomecollection/docs/blob/api-expectations/rfcs/037-api-faceting-principles/README.md

### DIFF
--- a/rfcs/037-api-faceting-principles/README.md
+++ b/rfcs/037-api-faceting-principles/README.md
@@ -1,0 +1,146 @@
+# API faceting principles & expectations
+
+**Status:** Draft
+
+**Last updated:** 16/03/2021
+
+## Background
+
+We've had a few conversations that keep coming up around how we expect the API to behave, particularly in terms of:
+
+- Filter naming
+- Aggregation naming
+- Combined effects of filters/aggregations
+- Aggregation response types
+- Empty aggregation buckets
+
+These are the components required to build a useful faceted search interface which covers numerous dimensions: so that we can do this effectively, we want these expectations to be both explicit and adhered to. This RFC is an attempt to document, question, and codify those expectations.
+
+## Principles
+
+**1. Filters are named by the JSON paths of the attribute they use**
+
+For example, given a display document that looks like:
+
+```json
+{
+  "a": {
+    "b": [
+      {
+        "c": "thing1"
+      },
+      {
+        "c": "thing2"
+      }
+    ]
+  }
+}
+```
+
+Then a document filter that filtered by the values of the `c` property would use a query string like `a.b.c`. In this example, filtering for `c` equal to `thing2` might look like:
+
+```
+http://host.name/path/docs?a.b.c=thing2
+```
+
+**2. Aggregations are always paired with identically named filters**
+
+It is not strictly necessary that all filters have aggregations, but all aggregations must be present alongside an identically named filter for the property that is being aggregated upon. This document will refer to these as "paired" filters and aggregations. For the above example, an aggregation on the values of the  `c` property would be used like this:
+
+```
+http://host.name/path/docs?aggregations=a.b.c
+```
+
+**3. Aggregations are returned in an `aggregations` field, with the same name by which they were requested** 
+
+This means JSON paths are still represented as strings, rather than being expanded. For example, the response to the previous example would include at the top level
+
+```json
+{
+  ...,
+  "aggregations": {
+    "a.b.c": {
+       "buckets": [
+         ...
+       ]
+    }
+  }
+}
+```
+
+
+
+**4. Aggregation buckets contain a `data` field of the same type as the aggregated property's parent entity**
+
+That is to say, while we aggregate on a specific string field (for example, an ID or a label), we want to return the full entity that contains the field. This is easiest to explain by example. Given the following display documents:
+
+```json
+{
+  "a": {
+    "b": [
+      {
+        "id": "id1",
+        "label": "Thing 1"
+      },
+      {
+        "id": "id2",
+        "label": "The 2nd thing"
+      }
+    ]
+  }
+}
+```
+
+Then if we aggregate on the labels like this:
+
+```
+http://host.name/path/docs?aggregations=a.b.label
+```
+
+Then our response buckets will look something like this:
+
+```
+{
+  "data": {
+    "id": "id1",
+    "label": "Thing 1"
+  },
+  "count": 1234
+}
+```
+
+**5. When a filter and its paired aggregation are both applied, that aggregation's buckets are not filtered**
+
+Conversely, filters do apply to the buckets of all aggregations other than the paired aggregation. This initially confusing requirement is necessary because - for mutually exclusive values - application of the filter to the aggregation buckets will remove all but the selected bucket, thus removing the ability of the interface to show other options for the given filter. This is explained further in the [Elasticsearch documentation for `post_filter`](https://www.elastic.co/guide/en/elasticsearch/reference/current/filter-search-results.html#post-filter).
+
+**6. When a filter and its paired aggregation are both applied, the bucket corresponding to the filtered value is always present**
+
+Explicitly: even if other filters or queries are present which cause that bucket to be empty (ie, it has a count of 0), it still appears in the aggregation. This is necessary so that the interface for the filter can still be rendered.
+
+**7. Aggregations on fields contained in sum types return buckets of the type's components**
+
+In other words - there can be a discriminator present on objects meaning that aggregations on identical string properties of those objects return separate objects for each type. For example, given the following display documents:
+
+```json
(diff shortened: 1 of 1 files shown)


rfcs/037-api-faceting-principles/README.md
```diff
@@ -0,0 +1,146 @@
...
+    ]
+  }
+}
+```
+
+Then an aggregation `a.b.label` would return separate buckets for each of the objects in `b`, even though their labels (the property being aggregated) are identical, because of the presence of the discriminator field `type`.
+
+## Open questions
+
+- Are these rules sufficient to tell us what we expect regarding empty buckets? Namely, that they should only be present when necessary to satisfy principle (6).
```
jtweed, line 145: Yes. I think we only want empty buckets when we need to show that something has been applied as a filter, not when applied filters cause buckets that would otherwise be there to be empty.

rfcs/037-api-faceting-principles/README.md
```diff
@@ -0,0 +1,146 @@
...
+- Are there cases when aggregations/filters should be named differently to rule (1)? For example, if we want to aggregate on an `id` property is it sufficient to use the name of the identified entity (eg, `a.b` rather than `a.b.id`)?
```
jtweed, line 146: Yes, you have pointed out the one exception to this. We don't add the `.id` on filters, but treat is as implied, and so I think it's correct to not add on aggregations.

jtweed (approved):
This looks really good to me, super clear and much needed clarification.

rfcs/037-api-faceting-principles/README.md
jamieparkinson: I think this might require a bit of thought to implement but will mean we can get rid of the hated `min_doc_count: 0`, which is an added bonus

rfcs/037-api-faceting-principles/README.md
jamieparkinson: > We don't add the .id on filters

Ah yeah this was unclear, I meant re both - was wondering whether this was intentional that we hadn't added it so far (eg for `workType`).

Can you give some justification for treating `id` properties as implied? Don't necessarily object to it but if we have an edge case that's treated differently we should be able to explain why easily.

rfcs/037-api-faceting-principles/README.md
```diff
@@ -0,0 +1,146 @@
...
+    "id": "id1",
+    "label": "Thing 1"
+  },
+  "count": 1234
+}
+```
+
+**5. When a filter and its paired aggregation are both applied, that aggregation's buckets are not filtered**
+
+Conversely, filters do apply to the buckets of all aggregations other than the paired aggregation. This initially confusing requirement is necessary because - for mutually exclusive values - application of the filter to the aggregation buckets will remove all but the selected bucket, thus removing the ability of the interface to show other options for the given filter. This is explained further in the [Elasticsearch documentation for `post_filter`](https://www.elastic.co/guide/en/elasticsearch/reference/current/filter-search-results.html#post-filter).
```
alexwlchan, line 114: I agree with the conclusion, but I think this would be better explained with an example. You also need to cover the case where values aren't mutually exclusive.

AIUI, our filters use OR rather than AND – e.g. if you apply the filter `?language=english,french,german`, it will find

```
(documents with "english") ∩ (documents with "french") ∩ (documents with "german")
```

When we present a UI for this, we want to be able to show all three options. If I select one, the other two should be available, so I can add that to expand my search.

rfcs/037-api-faceting-principles/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+**6. When a filter and its paired aggregation are both applied, the bucket corresponding to the filtered value is always present**
+
+Explicitly: even if other filters or queries are present which cause that bucket to be empty (ie, it has a count of 0), it still appears in the aggregation. This is necessary so that the interface for the filter can still be rendered.
```
alexwlchan, line 118: Ah, possibly this is what I was just saying? I think an example would help.

rfcs/037-api-faceting-principles/README.md
jtweed: Yeah it was deliberate, whether it was a good idea or not is debatable. It was intended as a shorthand, put in when I think all filters (except query) were id filters. The thinking was it made sense to say "I'm filtering for X, give me this one".

rfcs/037-api-faceting-principles/README.md
jamieparkinson: This is separate to your other comment but I will add an example

rfcs/037-api-faceting-principles/README.md
jamieparkinson: Hmmm, I will explain this better and add an example. The case where the values aren't mutually exclusive is not relevant here - the problem only exists for when the values are mutually exclusive. I will clarify this though.

jamesgorrie (commented):
The comments on label aggregations I think is true, but only half way there.

The fact that we are using labels is a stopgap solution to being able to filter/agg on identified objects. I feel we should be capturing that in this document, explaining that we are using a said stopgap. Filtering and aggregating on strings is something we will hope not to do once we have identified data in the API.

The reasons being is that we should not codify something we don't mean to perpetuate, we should think about the solution we want to make available to people that aren't the frontend, and think about how we demarcate that this functionality will be sunset once we have the new ✨ version.

It feels like we have two different filter/aggs
* *labelled enums* - such as `workType` (<= is an anomaly because of the sierra IDs), `languages`, `license`, `availabilities`
* *identified objects* - `subjects`, `genres`, `contributors.agents` <- these work off the functionality that has been spoken about a lot in this PR, [namely the inferred `.id` functionality](https://github.com/wellcomecollection/docs/pull/51/files#r596961477). These filter/aggs will point to the object e.g. `contributors.agent` and assume an identifier exists as it is needed to aggregate on, and that it will carry some data on how to reconstruct the object, e.g. derived data

Things I would like to think about:
* How do we mark these as "For internal use only" so we can remove it while being "responsible"?
* When do we think that we will look at getting IDs in?
*

jamesgorrie:
Perhaps another section would be to talk about expectations on how combining filters works with aggregation counts?

Currently how it works is:

When selecting filters that are of the same category, the results are additive. e.g. The contributor list:
* Contributor A (20)
* Contributor B (10)
* Contributor C (5)

When selecting A, I have 20 results, B, 30 results, C, 35 results.


If I combine the genre list:
* Genre A (20)
* Genre B (10)
* Genre C (5)

And select Contributor A ☝️ these counts will change, and in some instances some will disappear.

rfcs/037-api-faceting-principles/README.md
```diff
@@ -0,0 +1,146 @@
...
+    ]
+  }
+}
+```
+
+Then a document filter that filtered by the values of the `c` property would use a query string like `a.b.c`. In this example, filtering for `c` equal to `thing2` might look like:
+
+```
+http://host.name/path/docs?a.b.c=thing2
+```
```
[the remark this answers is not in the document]
jamieparkinson: Yeah, I will get to adding this eventually (see discussion in other comments)

rfcs/037-api-faceting-principles/README.md
```diff
@@ -0,0 +1,146 @@
...
+- Aggregation response types
+- Empty aggregation buckets
+
+These are the components required to build a useful faceted search interface which covers numerous dimensions: so that we can do this effectively, we want these expectations to be both explicit and adhered to. This RFC is an attempt to document, question, and codify those expectations.
+
+## Principles
+
+**1. Filters are named by the JSON paths of the attribute they use**
+
+For example, given a display document that looks like:
```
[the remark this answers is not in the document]
jamieparkinson: Yeah I'll add a short clarification to the first use

jamieparkinson:
- Filter counts are already covered by points 5 and 6 - I don't think anything you've described is additional? I might expand a bit on the explanation but the expectations are already covered. 

- Re label aggregations... I'm not sure that I understand/agree. Label aggregations exist, they will exist for a while, and this describes the expected behaviour for them if they do exist. If they're removed, then there is still no confusion as the documentation is explicit. 

  This RFC is _not_ a description of API architecture or intentions, but rather it's a description of how we expect filters and aggregations to behave if they exist. I will add something about preferring id-type aggregations over labels, but I don't want to add anything about data types or our model or anything like that, so I think your questions about deprecation and planning (which are important questions) should be addressed elsewhere.

- I'll copy the contents of this RFC into an ADR when we're happy with it, I think that's sufficient documentation. The only other output is to actually make it true!

jamesgorrie (approved):

PR #53, opened 2021-04-13
[WIP] Series modeling RFC

alicefuzier:
rendered version https://github.com/wellcomecollection/docs/blob/sierra-series/rfcs/044-sierra-series/README.md


alicefuzier:
Some points of discussion: 
- what type should be the thing contained in a serie statement? I've gone for `Work` because there`s some overlap between series and host item entries but there is an argument for a type `Serie`
- should volumes be modelled as Items? I have gone for yes in the current examples, mostly because things in 440/490/773 can have identifiers and there can be many works with the same identifier in 440/490 pr 773 but generally the volume can be different, so the identifier doesn't seem to refer to volume

rfcs/038-sierra-series/README.md
```diff
@@ -0,0 +1,292 @@
...
+            {
+                "title": "2",
+                "type": "Item",
+                "locations": []
+            }
+        ],
+        "type": "Work"
+    }
+]
+```
```
alicefuzier, line 67: Comments from @jtweed:
> - We should make it clear we dedupe if 490 + 430 are the same
> - If the partOf is from 490/830 the type should probably be Series
> - How do we support the search on the series like Encore does? ?partOf.title= ?
> - If we make the volume part of the title, how do we square that with a title keyword filter?
> - Maybe that's fine and we should be considering the title + volume combination as the thing this is partOf?

rfcs/038-sierra-series/README.md
```diff
@@ -0,0 +1,292 @@
...
+    {  // 830 with slighly different information
+        "title": "Fordham perspectives in continental philosophy",
+        "type": "Work"
+    },
+    {   // 830 & 773 contain the same information
+        "title": "ACLS Humanities E-Book.",
+        "type": "Work"
+    }
+]
+```
```
alicefuzier, line 125: Comments from @jtweed:
> - What a great example!
> - [Aside: I love how the first series link doesn't even return any results in Encore...]
> - When we create this identified work I'm assuming we're using the ISSN as the source identifier?
> - Items is probably the _better_ way to handle volumes, but for simplicity might be easier to make part of the work title?
> - That might be even more true if this isn't really a volume but is actually potentially part of a hierarchy?
> - Again series fields should probably be Series, if we don't know what the 773 is, Work is fine

rfcs/038-sierra-series/README.md
```diff
@@ -0,0 +1,292 @@
...
+- [1110225](https://search.wellcomelibrary.org/iii/encore/record/C__Rb1110225?lang=eng&suite=cobalt&marcData=Y)
+Serie statement in 440 with only the title
+```
+"partOf": [
+    {
+        "title": "Monographs on inorganic and physical chemistry",
+        "type": "Work"
+    }
+]
+```
```
alicefuzier, line 179: Comments from @jtweed:
> - Simplest possible example, though should possibly be Series as above.

rfcs/038-sierra-series/README.md
```diff
@@ -0,0 +1,292 @@
...
+            {
+                "title": "v. 13",
+                "type": "Item",
+                "locations": []
+            }
+        ],
+        "type": "Work"
+    }
+]
+```
```
alicefuzier, line 207: Comments from @jtweed:
> - As above and I'm increasingly convinced that volumes are works in their own right for published material
> - That makes them different from eg how we have digitised manuscripts
> - But seems to be borne out in the catalogue e.g. https://search.wellcomelibrary.org/iii/encore/record/C__Rb3043903

rfcs/038-sierra-series/README.md
```diff
@@ -0,0 +1,292 @@
...
+        "items": [
+            {
+                "title": "page 9", 
+                "type": "Item",
+                "locations": []
+            }
+        ]
+    }
+]
+```
```
alicefuzier, line 246: Comments from @jtweed:
> - This is a link to something that exists in the catalogue: https://wellcomecollection.org/works/tw5yfsfv
> - I'll be honest, my inital thought was it was a ID from WorlCat, but it's not, it's one of our iconographic numbers
> - These are unhelpfully currently in shelfmarks, but there is a piece of work for Alex to do soon that moves them to referenceNumber
> - We should probably add them as identifiers too when we do that work
> - That would mean we could search in $w against the identifiers that exist on works, assuming it's a id scheme we recognise
> - I don't know how consistent the use of Wcat is, I would have expected UkLW, so that's a question for #collections-info
> - Given that, in this instance the linked work should be embedded as per (and possibly by) the relation embedder

rfcs/038-sierra-series/README.md
alicefuzier: > When we create this identified work I'm assuming we're using the ISSN as the source identifier?

that would be my plan, yes

alicefuzier: > Items is probably the better way to handle volumes, but for simplicity might be easier to make part of the work title?
That might be even more true if this isn't really a volume but is actually potentially part of a hierarchy?

The main thing that bothers me is if we use ISSN or whatever to identify them, those refer to the whole Serie or host item entry (773) not to the volumes so you have many bibs referring the same ISSN but the volume they refer to is different. It makes me think that somehow volume should be a separate thing.
 https://search.wellcomelibrary.org/iii/encore/record/C__Rb2125597?lang=eng&suite=cobalt&marcData=Y
is an example of that. That issn is used in 490 by 1037 bibs but the volumes are different.

alicefuzier: > Again series fields should probably be Series, if we don't know what the 773 is, Work is fine

What is the difference between a Serie and a Work though? A Serie can have a ISSN and there is definitely overlap in some instances between 440/490 and 773 but it’s a bit weird if we use different types in those instances? Or do we decide to dedupe in those cases?
I know we said no `workType` cause it should be called Format etc, but couldn’t it be used for Works extracted from 440/490? eg
```
{
...
"workType": "Serie",
"type": "Work:,
 }
 ```

File: rfcs/038-sierra-series/README.md (hunk 1 above)
Comment (alicefuzier) [reply rfcs/038-sierra-series/README.md:67]: > We should make it clear we dedupe if 490 + 430 are the same

I'm trying to find examples where they are not the same, but while I can definitely find examples where 830 doesn't have a 490, so far, if it has both they either have the same info or overlapping info so it seems to me they can be collapsed in most examples?

Comment (alicefuzier) [reply rfcs/038-sierra-series/README.md:67]: > Maybe that's fine and we should be considering the title + volume combination as the thing this is partOf?

I think technically that's true, although if we collapse volume and title in one field it might be more difficult to search by title only?

File: rfcs/038-sierra-series/README.md (hunk 2 above)
Comment (alicefuzier) [reply rfcs/038-sierra-series/README.md:188]: I just found out that works from calm can have type `Series`, like [this](https://api.wellcomecollection.org/catalogue/v2/works/dq3spb42). Maybe that is an argument in favour of using type `Series` if they are used in a similar way in calm and sierra?
I guess if we decided to go for type `Series` for things extracted from 440/490 the front end would need to use the `workType` property to decide how to render, instead of `type`.

File: rfcs/038-sierra-series/README.md (hunk 1 above)
Comment (alicefuzier) [reply rfcs/038-sierra-series/README.md:143]: I've changed the examples to do ☝️  instead of using Items, i think it's more clear what it means

File: rfcs/038-sierra-series/README.md (hunk 4 above)
Comment (alicefuzier) [reply rfcs/038-sierra-series/README.md:278]: yeah, I think I agree now that Items don't work for this

File: rfcs/038-sierra-series/README.md (hunk 1 above)
Comment (jtweed) [reply rfcs/038-sierra-series/README.md:176]: > > We should make it clear we dedupe if 490 + 430 are the same
> 
> I'm trying to find examples where they are not the same, but while I can definitely find examples where 830 doesn't have a 490, so far, if it has both they either have the same info or overlapping info so it seems to me they can be collapsed in most examples?

Quite possibly, that would be one for Collections Info to confirm.

**Comment by jtweed** [comment]
> * what type should be the thing contained in a serie statement? I've gone for `Work` because there's some overlap between series and host item entries but there is an argument for a type `Serie`

If it comes from the series statement fields, it should probably be `Series`.

File: rfcs/038-sierra-series/README.md (hunk 1 above)
Comment (jtweed) [reply rfcs/038-sierra-series/README.md:176]: > I think technically that's true, although if we collapse volume and title in one field it might be more difficult to search by title only?

This is definitely the primary consideration here. If we go for a `partOf.label` search that works like the `subjects.concepts.label` etc then the value of the label will be treated as a single token and so only works in the same volume will be returned. That's arguably correct, but it doesn't give a way to navigate up a level.

> It is true that this model makes it not clear that the `Item` is the thing that the work is part of. Ie `Morphogenesis of the vertebrate brain` could and probably does have many more volumes but they don't appear as more items because they are not what this work is part of, which is a bit confusing.
> What if we in the case of things with volumes we had two levels of `partOf`. Ie in this case:
> 
> ```
>  "partOf": [
>   {
>     "title": "Morphogenesis of the vertebrate brain; 2",
>    "partOf": [
>    {
>         "title": "Morphogenesis of the vertebrate brain",
>         "type": "Work"
>     }],
>     "type": "Work"
>   }
>  ]
> ```

I definitely wouldn't model this as items and I think if we are going to split it, something like the above would be the way to do it. However, in that example there is a duplication of the main part of the title that is problematic.

I'd be tempted to not overthink the first iteration of this and make the example above partOf a single work, that includes the volume number.

File: rfcs/038-sierra-series/README.md (hunk 2 above)
Comment (jtweed) [reply rfcs/038-sierra-series/README.md:242]: > The main thing that bothers me is if we use ISSN or whatever to identify them, those refer to the whole Serie or host item entry (773) not to the volumes so you have many bibs referring the same ISSN but the volume they refer to is different. It makes me think that somehow volume should be a separate thing.

That is (possibly unfortunately) correct. An ISSN refers to an entire ongoing serial publication, eg a journal, not a specific volume or issue of that publication. So it's not correct for us to assume it refers to a specific volume or part and we shouldn't try to use it to uniquely identify a part.

Comment (jtweed) [reply rfcs/038-sierra-series/README.md:242]: > What is the difference between a Serie and a Work though? A Serie can have a ISSN and there is definitely overlap in some instances between 440/490 and 773 but it’s a bit weird if we use different types in those instances? Or do we decide to dedupe in those cases?

Nothing really, a Series in our model is just a more specific type of Work. It's unlikely to be even rendered on the page as such. It's just an attempt to hint that it came from a series statement rather than a more general relationship, so is likely to be something considered a Series. A work can (and many of ours do) have an ISSN, eg journals. If there's overlap, possibly it's better to just keep them all the more generic Work if we can't say what it really is.

> I know we said no `workType` cause it should be called Format etc, but couldn’t it be used for Works extracted from 440/490? eg
> 
> ```
> {
> ...
> "workType": "Serie",
> "type": "Work:,
>  }
> ```

I'd disagree with this. We use format/workType to describe what Sierra calls the material type, ie it's ephemera, it's a book, it's a journal, it's a manuscript etc. A Series could in theory be any of those formats.

Comment (jtweed) [reply rfcs/038-sierra-series/README.md:242]: > I just found out that works from calm can have type `Series`, like [this](https://api.wellcomecollection.org/catalogue/v2/works/dq3spb42). Maybe that is an argument in favour of using type `Series` if they are used in a similar way in calm and Sierra?

This is why we added the more specific subtypes of Work, to handle the hierarchical relationships in archives and, we hoped, to help model library series statements in a similar way.

> I guess if we decided to go for type `Series` for things extracted from 440/490 the front end would need to use the `workType` property to decide how to render, instead of `type`.

Yes, the front end will have to render based on the workType. Eg show archive tree for archives, but not for books as the UI for archives won't handle a single level with tens of thousands of entries. That's why we don't want to denormalise here and instead intend to link to a search results page.

File: rfcs/038-sierra-series/README.md [hunk 6]
```diff
@@ -0,0 +1,432 @@
...
+to mint a canonical identifier in the pipeline and to expose that identifier in the API. 
+Finally, the proposal is to model a series statement or a Host Item Entry with a type `Work`.
+
+### Examples
+ 
+- [1074785](https://search.wellcomelibrary.org/iii/encore/record/C__Rb1074785?lang=eng&suite=cobalt&marcData=Y) 
+This is an example of a series statement in 490 + 830. In this example the information contained in 830 and 490 is 
+  exactly the same, so the result is just one statement. 
+```yaml
+"partOf": [
```
jtweed, line 159: The other option if we want to include the series both with and without the volume is to flatten them into a list, rather than trying to make it a hierarchy. It's kind of a truer representation of what we're doing with the data and might also be easier to represent on the front end, given the duplication of the whole title.

jtweed: I still think it might be better to just treat as a single partOf though initially, ie a single work that includes the volume number. Then iterate based on feedback.

rfcs/038-sierra-series/README.md
```diff
@@ -0,0 +1,432 @@
...
+      }
+    ],
+    "type": "Series"
+  }
+]
+```
+- [3109805](https://search.wellcomelibrary.org/iii/encore/record/C__Rb3109805?lang=eng&suite=cobalt&marcData=Y)
+  490 + 773 with slightly different information
+```yaml
+"partOf": [
```
jtweed, line 180: This example neatly contradicts what I just said above, as it does look more clearly hierarchical.

Indeed in this instance it would be most correct to have "Early European Books : Printed sources to 170" as partOf "Early European Books" (ie the 773).

If that's too difficult, it could also be list that is de-duped across the values generated from 490/830 and 773. It's likely to rendered as a list on the front end anyway.

rfcs/038-sierra-series/README.md
```diff
@@ -0,0 +1,432 @@
...
+  }
+]
+```
+- [2125597](https://search.wellcomelibrary.org/iii/encore/record/C__Rb2125597?lang=eng&suite=cobalt&marcData=Y)
+  490 with ISSN + 830 and 773. There are 1037 bibs that have the same ISSN in a series statement but
+  with different volume subfield.
+```yaml
+"partOf": [
+  {   // 490 +830 contain duplicated info
+    "partOf": [
```
jtweed, line 249: Is there an unintended extra level here?

rfcs/038-sierra-series/README.md
```diff
@@ -0,0 +1,432 @@
...
+- [3001878](https://search.wellcomelibrary.org/iii/encore/record/C__Rb3001878?lang=eng&suite=cobalt&marcData=Y)
+  Only 830 with no 490/440 or 773
+```yaml
+"partOf": [
+    {
+        "title": "Adamson Collection Wellcome Library", // subfield $a concatenated with $p
+        "type": "Series"
+    }
+]
+```
```
jtweed, line 284: Love the simple examples! So much easier than the other ones :)

rfcs/038-sierra-series/README.md
```diff
@@ -0,0 +1,432 @@
...
+- [1204561](https://search.wellcomelibrary.org/iii/encore/record/C__Rb1204561?lang=eng&suite=cobalt&marcData=Y) 440 with id
+```yaml
+"partOf": [
+  {
+    "title": "Brill's studies in intellectual history; v. 13",
+    "workType": {
```
jtweed, line 300: Wouldn't use workType here. In any case it would be Journals, not Volumes. If we wanted to model journals more granularly it would either be with Items or another subtype of Work called Volume. Neither of which I think we want to do here, so it's probably better to collapse it and consider the ISSN non-unique.

rfcs/038-sierra-series/README.md
```diff
@@ -0,0 +1,432 @@
...
+```yaml
+"partOf": [
+   {
+    "title": "Basil Hood. Photograph album; page 9",   
+    "partOf": [
+      {
+        "id": "abcdefgh",
+        "identifiers": [
+          {
+            // identifier from $w
```
jtweed, line 348: Correct thing to do here would be to use that id to look up the existing work (which I think we will be added when we sort the shelfmarks that should really be ids), then denormalise the existing work into here with the relation embedder.

rfcs/038-sierra-series/README.md
alicefuzier: yup, sorry my bad, will fix

rfcs/038-sierra-series/README.md
alicefuzier: By "consider ISSN non unique" what do you mean? I think I see your point of not overthinking it and probably it's easier if we collapse the two, but I struggle to understand how we want to include the identifiers block then. So far (I think) we've always put the identifier on the thing it refers to but if we collapse the volume, we break that and the model becomes a bit misleading I think? We could decide to include the nested `partOf` only when we have an identifier and not if we don't have it?

alicefuzier: > Wouldn't use workType here. In any case it would be Journals, not Volumes. If we wanted to model journals more granularly it would either be with Items or another subtype of Work called Volume

I see. I think it would be a bit difficult in the transformer to figure out when `workType` `Volume` makes more sense than `Journal`. May e it would be best to leave the `workType` out always (even in the previous examples)? I'm not sure it's adding anything

rfcs/038-sierra-series/README.md
alicefuzier: yup, I think that makes a lot of sense

alicefuzier: For now, as shelfmarks are not ids, I think it would make sense to not include the identifers field and possibly the whole nested `partOf`

rfcs/038-sierra-series/README.md
alicefuzier: I think the rules to implement in the transformer to figure out when a host item entry field completely matches a series statement, when it's a more general higher level and when it's not related at all could become quite complicated. In these examples I tried to go for: if it matches completely a series statement -> collapse; if it doesn't -> keep it separate. The main thinking is that this is easy to implement and limits the chance that we present information that is wrong. 
I think it might be difficult to come up with a rule that could figure out in this instance what case this should fall into, so I would just go for not trying to be too clever at least as a first iteration and just accept that on the frontend there will be a list and that some elements on that list will appear as duplicates to a human

rfcs/038-sierra-series/README.md
alicefuzier: > I'd disagree with this. We use format/workType to describe what Sierra calls the material type, ie it's ephemera, it's a book, it's a journal, it's a manuscript etc. A Series could in theory be any of those formats.

Yes, after talking with @jamesgorrie , I see this now

rfcs/038-sierra-series/README.md
alicefuzier: AS I said below, I think it would make sense to collapse. The only case where I don't agree with this is when we have identifiers to present. I think in that case we would present something that is potentially misleading. As I said below, I think maybe we could include the nested `partOf` only if we have identifiers?

> However, in that example there is a duplication of the main part of the title that is problematic.
yes, it doesn't look very good, but it is technically correct I think? I couldn't think of a better way to present this

rfcs/038-sierra-series/README.md
```diff
@@ -0,0 +1,432 @@
...
+    ],
+    "type": "Series"
+  },
+  {  // Second 830 with slighly different information
+    "title": "Fordham perspectives in continental philosophy",
+    "type": "Series"
+  },
+ {   // 830 & 773 contain the same information
+    "title": "ACLS Humanities E-Book.",
+    "type": "Work" //or Series?
```
alicefuzier, line 239: Just realised that 773 for `ACLS Humanities E-Book.` has URL: http://www.humanitiesebook.org/ in subfield $n and that is shown in Encore, so it should probably made to fit in this model. Not sure where it would fit in the `Work` model though?

rfcs/038-sierra-series/README.md
jtweed: I think let's collapse these then down into a single partOf. We do have other examples of the same identifier on multiple works. If we do split, then, yes this is the best way to treat the titles, not trying to be too clever about it.

rfcs/038-sierra-series/README.md
jtweed: Agreed, let's de-duplicate an exact match. We can then keep an eye out for things that need data cleanup and also any examples that we need to look at further.

rfcs/038-sierra-series/README.md
jtweed: Yeah if we don't know for certain (ie it's a link to a work we actually have in the catalogue) then let's not try and add a workType.

rfcs/038-sierra-series/README.md
jtweed: @alexwlchan has started to pick this up and so it's likely that, given we're going to the TEI work first, that they will be by the time we implement library linking.

rfcs/038-sierra-series/README.md
jtweed: It would be good to know how many have a $n, but I'm going to go with thinking it's ok to drop this. You have to access and sign in via the library site anyway, so a direct link to the publisher site isn't that useful.

jtweed: If we did have to do this, in theory we could handle it like other resource links, but I have no idea how we would then render it on the page. Given the limited utility anyway, I don't think we should try to shoehorn it in.

rfcs/038-sierra-series/README.md
alexwlchan: \*swears loudly*

I haven't read this whole RFC, but this might get a bit complicated with shelfmarks. In particular:

* IDs in that field don't have the (Wcat) prefix
* Not every (Wcat)-prefixed ID in a 773-field appears in the shelfmarks (e.g. 570997i), but some of them do (e.g. 5894i)
* How do we tell these apart from an Iconographic Collections identifier?

aaaarrggghhh

rfcs/044-sierra-series/README.md
```diff
@@ -0,0 +1,503 @@
...
+    490 1  Morphogenesis of the vertebrate brain ;|v2
+    830  0 Morphogenesis of the vertebrate brain ;|v2
+    ```
+
+    We'll deduplicate because the information in 490 and 830 is the same:
+
+    ```json
+    "partOf": [
+      {
+        "title": "Morphogenesis of the vertebrate brain; 2",
```
paul-butcher, line 216: In the current implementation, the title of a partOf appears to be the name of the thing that this is part of.

In this proposal, it appears to be the name of this thing considered in the context of the other.

By reusing the field in this way, a client would have to examine other properties in order to work out what the field means.

e.g.
This means "I am a part of the work 'My Greatest Hits', which has the identifier 'baadf00d'".
```
"partOf": [{
  "title": "My Greatest Hits",
  "id": "baadf00d"
}]
```


Whereas this seems to mean "I am the part called 'Portraits of Cool People, Page 23' of 'Portraits of Cool People'"

```
"partOf":[{
  "title": "Portraits of Cool People, Page 23",
  "partOf": [{
   "title": "Portraits of Cool People",
    "id": "deadbeef"
  }]
}]
```

So an existing application that asks of a record "what are you a part of?" will currently just take partOf.title as the answer. Given this change, it would have to check whether there is a partOf.partOf, then drill into that to find it.

Perhaps this property should have a different name e.g. "partTitle"?

rfcs/044-sierra-series/README.md
```diff
@@ -0,0 +1,503 @@
...
+    }
+]
+```
+- [1186777](https://search.wellcomelibrary.org/iii/encore/record/C__Rb1186777?lang=eng&suite=cobalt&marcData=Y)
+  773 with an id in subfield $w which links to another Work in our library (see example below)
+```yaml
+"partOf": [
+   {
+    "title": "Basil Hood. Photograph album; page 9",
+    "partOf": [
```
paul-butcher, line 414: I'm not sure the nested partOf should be a list.  Particularly considering that the title above seems to refer to the part in context of what it is part of.

e.g. 
A portrait of Basil Hood in a white coat outside Charing Cross Hospital might belong to any number of containers (portraits of BH, photos of CXH ...), so a partOf list is needed at the top level.

I would expect that Page 9 of the Basil Hood Photo Album is only ever going to be part of the Basil Hood Photo Album. (not Top 10 ninth pages).

This actually calls into question the need to have nested partOf objects. If we were to rename the part title, then all of the properties in the nested partOf could be promoted to the top-level partOf.

rfcs/044-sierra-series/README.md
```diff
@@ -0,0 +1,503 @@
...
+These are [rendered on the website](https://wellcomecollection.org/works/dq3spb42) as a collapsible hierarchy:
+
+![](archives.png)
+
+Some series can have thousands of entries, so this UI isn't practical – we'll need to render the information differently.
+We'll determine how to render them based on the `type` property of the entry in `partOf`.
+
+*   For archive works, we'll use `partOf: Collection` and render an archive tree
+*   For works which are part of a series (440, 490, 830), we'll use `partOf: Series` and display a link as part of the work metadata (similar to Encore)
+*   For works which have constituent parts or are part of something else, we'll use `partOf: Work`
```
paul-butcher, line 196: Should this render an archive tree in the same manner as partOf: Collection?

rfcs/044-sierra-series/README.md
```diff
@@ -0,0 +1,503 @@
...
+        "type": "Work"
+    }
+]
+```
+- [1186777](https://search.wellcomelibrary.org/iii/encore/record/C__Rb1186777?lang=eng&suite=cobalt&marcData=Y)
+  773 with an id in subfield $w which links to another Work in our library (see example below)
+```yaml
+"partOf": [
+   {
+    "title": "Basil Hood. Photograph album; page 9",
```
paul-butcher, line 413: In the current implementation, the title of a partOf appears to be the name of the thing that this is part of.

In this proposal, it appears to be the name of this thing considered in the context of the other.

By reusing the field in this way, a client would have to examine other properties in order to work out what the field means.

e.g.
This means "I am a part of the work 'My Greatest Hits', which has the identifier 'baadf00d'".
```
"partOf": [{
  "title": "My Greatest Hits",
  "id": "baadf00d"
}]
```


Whereas this seems to mean "I am the part called 'Portraits of Cool People, Page 23' of 'Portraits of Cool People'"

```
"partOf":[{
  "title": "Portraits of Cool People, Page 23",
  "partOf": [{
   "title": "Portraits of Cool People",
    "id": "deadbeef"
  }]
}]
```

So an existing application that asks of a record "what are you a part of?" will currently just take partOf.title as the answer. Given this change, it would have to check whether there is a partOf.partOf, then drill into that to find it.

Perhaps this property should have a different name e.g. "partTitle"?

paul-butcher: Actually, I think I'm a bit wrong here.

What this is saying is I am a part of page 9 of the Basil Hood Photo Album which is part of the Basil Hood Photo Album.

In that case, nested partOf makes sense, semantically.  However, I'm still slightly concerned about the change of interface here.

rfcs/044-sierra-series/README.md
paul-butcher: Looking wider, at the idea that this construct also serves to put a journal article into a particular volume of a journal, without having to maintain a record for that specific volume:

What would happen if that volume becomes significant in its own right and deserves a record?  Is that a thing that might happen (collection: journal volumes with discredited articles)?

rfcs/044-sierra-series/README.md
paul-butcher: It's notable that "page 9" is not something with an identifier in either direction. The 773s refer to the id of the album, and the 774s refer to the ids of each photo.

rfcs/044-sierra-series/README.md
```diff
@@ -0,0 +1,503 @@
...
+    ],
+    "type": "Work"
+  }
+]
+```
+- [1172977](https://search.wellcomelibrary.org/iii/encore/record/C__Rb1172977?lang=eng&suite=cobalt&marcData=Y)
+  774 with ids (related to above 773)
+```yaml
+"parts": [
+  {
```
paul-butcher, line 441: In the existing API, I think parts objects must have an id. However, unless we do treat individual pages or volumes as objects in their own right, I don't think this interface change is avoidable.

rfcs/044-sierra-series/README.md
```diff
@@ -0,0 +1,503 @@
...
+
+These are [rendered on the website](https://wellcomecollection.org/works/dq3spb42) as a collapsible hierarchy:
+
+![](archives.png)
+
+Some series can have thousands of entries, so this UI isn't practical – we'll need to render the information differently.
+We'll determine how to render them based on the `type` property of the entry in `partOf`.
+
+*   For archive works, we'll use `partOf: Collection` and render an archive tree
+*   For works which are part of a series (440, 490, 830), we'll use `partOf: Series` and display a link as part of the work metadata (similar to Encore)
```
paul-butcher, line 195: Do we need to change the front end before changing the API to include this?  If we add this to the API, will it cause an explosion of a tree to appear?

rfcs/044-sierra-series/README.md
alexwlchan: I don't think so – archive collections are ordered, whereas we don't have any way to order these parts.

rfcs/044-sierra-series/README.md
```diff
@@ -0,0 +1,503 @@
...
+        ],
+        "type": "Series"
+      },
+      {  // Second 830 with slighly different information
+        "title": "Fordham perspectives in continental philosophy",
+        "type": "Series"
+      },
+     {   // 830 & 773 contain the same information
+        "title": "ACLS Humanities E-Book.",
+        "type": "Work" //or Series?
```
paul-butcher, line 314: This "//or Series?" raises a question that runs through a lot of this document.

If all we have to go on is a title, how do we determine what kind of thing it refers to?

For unidentified targets, should they have a type of "Unknown" (or no type at all?)

paul-butcher: 4xx and 83x properties are about Series, so they can be identified as such even in the absence of an ISSN in the x field.

773 is not necessarily a series, but may refer to things "serial in nature", (a rose by any other name). So I would guess that if we get a matching 830 and 773, then the 830 "wins", and it's a series.

That leaves the scenario where we have 773 on its own, and no id.  I think in that situation, we can't call it a Work, because we can't find the work.  We shouldn't call it a series, because it may not be, it might just be a single book published in three volumes (or even just a single volume, 773 isn't picky).

rfcs/044-sierra-series/README.md
paul-butcher: Another thing that worries me about nesting partOf objects (from a client PoV) is "where will it end?"

Were I an external consumer of this API, I would want to know how far I need to recurse down the nest until I get to the thing I want.

From an internal perspective, it looks like we would only make this one level deep and with only one entry in the nested one, so _we_ could write things like partOf[n].partOf[0] and know we are at the terminus.

Someone else could not (or they could, but would not necessarily be confidently correct).  They would have to loop and recurse until some condition is met (no more partOfs, a partOf with an id?)

Similarly, if we do nest them, would a consumer have an expectation that the nesting should be comprehensive?  e.g. "article A is part of vol2 of journal B which is in collections D, E and F, collection D is in ..."

rfcs/044-sierra-series/README.md
paul-butcher, line 216: If the relationship could be presented in a flat object, then I think that maintaining both the $a and $v or $g subfields in this property is probably redundant.
e.g. 
`{"subpart": "2", "title": "Morphogenesis of the vertebrate brain"}`, rather than repeating the title of the superobject in both properties

PR #55, opened 2021-04-16
Add an RFC proposing a change to matcher versioning

alexwlchan:
Rendered version: https://github.com/wellcomecollection/docs/blob/change-matcher-versioning/rfcs/038-matcher-versioning/README.md

Thinking about this makes my head hurt; the new system has the virtue of being quite a bit simpler, which hopefully means it’s easier to reason about and will thus be less likely to be broken in subtle ways.

I'd particularly like a review from @jamieparkinson, who's spent the most time thinking about how versioning works recently.

For https://github.com/wellcomecollection/platform/issues/5132


alicefuzier:
Thanks! really clear explanation of the current problem. Could you run through the same example at the beginning with the proposed change? I think I don't understand 100% how using the timestamp from the matcher solves the problem in this instance while also allow us to unlink

alexwlchan:
> Could you run through the same example at the beginning with the proposed change? I think I don't understand 100% how using the timestamp from the matcher solves the problem in this instance while also allow us to unlink when necessary

Good suggestion, added.

rfcs/038-matcher-versioning/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+    But it can't be fixed once it's occurred, unless you manually edit the post-merger Elasticsearch indexes (scary).
+
+*   Until the matcher/merger, we have a 1:1 correspondence between "source record" and "catalogue Work".
+    After those services, we potentially break that link, and relying on it for ordering gets more difficult.
+
+
+## Proposed fix
+
+**We track the latest modified time of a *matcher graph*, and use that to version works post-matcher.**
```
jamieparkinson, line 83: How (or where) do we do this? In the dynamo rows for graph nodes?

jamieparkinson (approved):
I think this will work, nicely done and reasoned

rfcs/038-matcher-versioning/README.md
alexwlchan: Let me add a sentence on this.

jamieparkinson: Ah, I see now - I think you could actually get this off the SQS message already https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html

alexwlchan: Ooh, neat. I think I prefer it in our message – the SQS sending time could get messed up if we have to redrive messages.

alicefuzier (approved):
Thanks, really clear. Let's hope it works this time!

PR #56, opened 2021-04-19
Add an initial description of some requesting APIs

alexwlchan:
Rendered version: https://github.com/wellcomecollection/docs/blob/describe-apis/rfcs/039-requesting-api-design/README.md

An initial attempt at describing some requesting APIs.


rfcs/039-requesting-api-design/README.md
```diff
@@ -0,0 +1,215 @@
...
+    ]
+    ```
+
+    The `id` is a value which should not be shown to the user, but can instead be used to relinquish requests.
+
+-   To relinquish a pending request, a user makes a DELETE request to `/item-requests/{requestId}`
+
+Notes:
+
+-   Because these operations modify a user's account, they must be authenticated first.
```
jtweed, line 215: It would be good to include a bit here about how that authentication process works, the tokens that the app needs to send etc.

jtweed (approved):

rfcs/039-requesting-api-design/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Background
+
+As part of the work to add requesting, we need to add new some new APIs.
+In particular, these APIs should allow a user to:
+
+-   Get the status of a given item (e.g. available, on hold)
+-   Request an item
+-   See a list of their current requests
+-   Relinquish a pending request
```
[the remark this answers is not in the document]
alexwlchan: Yeah, I’m going to de-scope this.

rfcs/039-requesting-api-design/README.md
```diff
@@ -0,0 +1,215 @@
...
+      "type": "ItemStatus"
+    }
+    ```
+
+-   This will return an item for every item in the corresponding `/works` request.
+
+    If an item cannot be requested (e.g. it is tracked in Calm, not Sierra), we will return a `not-requestable` status.
+    This includes items that only have digital locations.
+
+-   Using this endpoint will require an API key.
```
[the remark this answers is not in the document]
alexwlchan: That’s the current implementation, but of course that’s not used by the front-end yet.

It’s fairly easy to toggle in API Gateway, and I figured it was better to leave it “on” until I have a good reason otherwise.

rfcs/039-requesting-api-design/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+    Requests to this endpoint will trigger requests to Sierra (and possibly other systems in future).
+    We don't want it to be used to DDOS those systems, so we need a way to turn off misbehaving clients.
```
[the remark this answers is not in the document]
alexwlchan: \*nod*

Whether we make API calls in the browser relies on a better understanding of the front-end than I currently possess.

rfcs/039-requesting-api-design/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+-   To retrieve a list of their pending requests, a user makes a GET request to this endpoint.
+
+    This will return a list of ItemRequest objects:
+
+    ```json
+    [
+      "requests": [
+        {
+          "id": "sierra/1234567",
```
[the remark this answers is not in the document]
alexwlchan: I'm going to drop this until we bring relinquishing requests back into scope.

I'm not sure if you can rely on GUIDs – it adds another data store and potential failure point. e.g. if we successfully record the hold in Sierra but don't manage to store our GUID, what do we show the user? The request didn't totally "succeed" (because they can't relinquish it), but nor did it "fail" (staff will be retrieving the item for them).

rfcs/039-requesting-api-design/README.md
```diff
@@ -0,0 +1,215 @@
...
+    If an item cannot be requested (e.g. it is tracked in Calm, not Sierra), we will return a `not-requestable` status.
+    This includes items that only have digital locations.
+
+-   Using this endpoint will require an API key.
+
+    Requests to this endpoint will trigger requests to Sierra (and possibly other systems in future).
+    We don't want it to be used to DDOS those systems, so we need a way to turn off misbehaving clients.
+
+<details>
+<summary>Example request/response</summary>
```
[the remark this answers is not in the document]
alexwlchan: I think the front-end will continue to behave as it currently does – it has to get enough of the items to know it's worth making this request (no point requesting items on a sans-item work). The items are a useful thing it can render immediately, without hitting this API – in particular, digitised items.

If the initial request to /works succeeds but this API fails, we should still show the items on the page.

rfcs/039-requesting-api-design/README.md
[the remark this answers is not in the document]
jamieparkinson: We would almost certainly want to be proxying requests to this API if they were initiated from the browser - I would typically expect browser requests to be made to a proxy API at the same host, which can then securely hold an API key or whatever.

rfcs/039-requesting-api-design/README.md
```diff
@@ -0,0 +1,150 @@
...
+      "workId": "a23upgqy",
+      "itemId": "fwsbxqp4",
+      "type": "ItemRequest"
+    }
+    ```
+
+    This endpoint will return the following status codes:
+
+    -   201 Created if the request is successful (or already exists)
+    -   403 Forbidden if the user is unable to place requests (e.g. at their account limit, banned account)
```
jamieparkinson, line 122: We will want a `401` as well for if they are not authenticated

rfcs/039-requesting-api-design/README.md
```diff
@@ -0,0 +1,150 @@
...
+      "id": "fwsbxqp4",
+      "status": {
+        "id" : "not-requestable",
+        "label" : "Not requestable",
+        "type": "ItemStatus"
+      },
+      "type": "Item"
+    }
+  ],
+  "type": "Work",
```
jamieparkinson, line 91: I'm not sure about this (and hence also https://github.com/wellcomecollection/docs/pull/56/files#diff-23afd2605893f14d49f2d9d49cc5aa1966d3e72a02bffe5958ef560e17591321R41) - this doesn't have very much in common with a work and from a REST pedantry perspective the URL implies a collection / resource list rather than an atomic resource...

Could it be an `ItemsList` or something?

jamieparkinson (approved):

rfcs/039-requesting-api-design/README.md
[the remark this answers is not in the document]
kenoir: > if we successfully record the hold in Sierra but don't manage to store our GUID
We would record the attempt to place a request before making one, then update with a confirmation that the request succeeded I think. But we shouldn't think about this yet (as you say).

rfcs/039-requesting-api-design/README.md
```diff
@@ -0,0 +1,150 @@
...
+          "workId": "a23upgqy",
+          "itemId": "fwsbxqp4",
+          "type": "ItemRequest"
+        },
+        {
+          "workId": "a23upgqy",
+          "itemId": "fwsbxqp4",
+          "type": "ItemRequest"
+        }
+      ]
```
kenoir, line 143:
```suggestion
      ],
```

rfcs/039-requesting-api-design/README.md
kenoir: I agree a nice sequence diagram would be good.

kenoir (commented):
Looks good.

rfcs/039-requesting-api-design/README.md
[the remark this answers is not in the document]
kenoir: I think it's worth adding the point @jamieparkinson has made in this doc (that requests would be proxied to the node server from the browser client, and the node server would hold the api key securely).

rfcs/039-requesting-api-design/README.md
kenoir: That or we modify our description of a `Work` to imply that all the missing fields are optional. I think @jtweed will have an opinion on this.

jamieparkinson: You mean the missing fields in `items`? hmmmm...

rfcs/039-requesting-api-design/README.md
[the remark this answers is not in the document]
alexwlchan: Added.

rfcs/039-requesting-api-design/README.md
alexwlchan: Both, I think.

alexwlchan: Having checked – the only compulsory fields on DisplayItem are id and locations (which can be an empty list), so omitting that doesn't seem too much of a sin. The Work is a bigger issue, I think.

jamieparkinson: `DisplayResultList[DisplayItem]`?

jamieparkinson: although that doesn't give you the id back...

rfcs/039-requesting-api-design/README.md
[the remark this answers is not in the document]
alexwlchan: > I'm assuming the generally serialisation here is designed to be consistent with the rest of the identity API?

Yup.

PR #57, opened 2021-04-22
TEI adapter RFC

alicefuzier:
Rendered version https://github.com/wellcomecollection/docs/tree/tes-adapter/rfcs/040-tei_adapter


rfcs/0xx-tei_adapter/README.md
```diff
@@ -0,0 +1,93 @@
...
+GitHub is a relatively reliable and low latency source, so we could in theory just send the changes to a TEI transformer 
+and do a reindex by cloning the repository and/or crawling the TEI files. 
+There are three disadvantages to this approach:
+- We need a "special" reindexer for TEI.
+- We can't keep track of deleted files. If something is deleted, the change will go through the pipeline once, 
+  but there's no way of "remembering" it.
+- We need to assign versions to works resulting from TEI files in order to ensure subsequent updates are 
+  processed correctly in the pipeline. This versioning logic would have to be added explicitly to the 
+  TEI transformer whereas having an adapter and an intermediate store would allow us to use the already existing versioning
+  logic in the versioned store library.
```
alexwlchan, line 21: one more reason: currently we use the # of works in the source table to track the success of a reindex.

50 records in => 50 records out

You can't do that if you don't have a canonical store.

rfcs/0xx-tei_adapter/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+However, since GitHub is a relatively reliable and low latency source, the proposal is to only store the HTTP URLs
+to the TEI files in a TEI store, and not the full XML. This has the advantage that we don't have to manage an 
+ever growing S3 bucket (we don't have a way of cleaning up old versions of a record) but it has some disadvantages:
+- GitHub outages (which are rare but happen) might affect our reindexes.
+- We might be throttled during a reindex (?)
+- We go to the public internet for every TEI file for every reindex as opposed to getting each file exactly once 
+  per every change that affects it.
+  
+For these reasons, the proposal here is to store thw whole TEI file in a full VHS.
```
alexwlchan, line 33: I'm confused – L25 says the proposal is to only store the HTTP URLs, L33 says we're storing the whole TEI file.

I think you're trying to say "there are two approaches we could pick from, here is why approach #1 is bad, so we're going to use approach #2".

rfcs/0xx-tei_adapter/README.md
```diff
@@ -0,0 +1,93 @@
...
+### TEI Adapter
+![](tei_adapter_stack.png)
+The proposal is to make the TEI adapter 2 services
+#### GitHub Api service
+
+It receives time windows from the lambda and for each one:
+  
+- It calls the GitHub API to get the commits within that window
+- Call the GitHub API to get the files changed in each commit
+- Send a message for each file
```
alexwlchan, line 80: It'd be useful to explain the exact GitHub API calls you're planning to make. It wasn't immediately obvious to me that this would be possible, so an example might be good.

How often do you imagine we might poll GitHub?

rfcs/0xx-tei_adapter/README.md
```diff
@@ -0,0 +1,93 @@
...
+####  TEI Adapter service
+
+It receives file URLs from the GitHub Api service and for each one:
+
+- Retrieves it from GitHub & store it in the TEI store
+- Send each record stored to the transformer
+
+### Full reharvest
+
+For the full reharvest we crawl the repo and send the URL of each file to the TEI Adaper service.
```
alexwlchan, line 92: What does "crawl the repo" mean in this context?

I wonder if looking at Git trees might be useful? Quick Git recap: you have "blobs" (files) and "trees". Trees form a recursive data structure:

image

You can see the root tree here: https://api.github.com/repos/wellcomecollection/wellcome-collection-tei/git/trees/master

If you used the trees to build a file map (path -> sha), that's another way to do it. Have some code that compares the current tree to a stored tree in VHS or whatever, and use that to identify files that need updating.

(Noticing deleted files gets a bit trickier in this case.)

So your process would go something like:

```
[get current tree] -> [compare tree to stored tree] -> [get each different file]
```

with the advantage that doing a full reharvest is just "get the very latest tree", and this runs in constant-time, rather than slowing down if there are more commits.

OTOH, you now have to hold an entire file map somewhere, hmm.

alexwlchan: This is where a bit more detail on the GitHub API calls would be useful, I wonder a bit about something like https://stackoverflow.com/questions/26925312/github-api-how-to-compare-2-commits

rfcs/0xx-tei_adapter/README.md
alicefuzier: sorry, my bad. Will fix thanks

rfcs/0xx-tei_adapter/README.md
alicefuzier: Makes, sense, thanks. I will add the examples. I haven't actually thought about that, but looking at the activity on the repo, I think something like every 30 mins with 40 min windows should probably not overwhelm the adapter & allow changes to go through reactively enough

rfcs/0xx-tei_adapter/README.md
alicefuzier: Thanks, this is super helpful!

rfcs/0xx-tei_adapter/README.md
alicefuzier: Added thanks

rfcs/0xx-tei_adapter/README.md
```diff
@@ -0,0 +1,95 @@
...
+#### GitHub Api service
+
+It receives time windows from the lambda and for each one:
+  
+- It calls the GitHub API to get the commits within that window
+- Call the GitHub API to get the files changed in each commit
+- Send a message for each file
+
+
+####  TEI Adapter service
```
kenoir, line 85: Is the TEI service filtering interesting files or the GitHub service?

rfcs/0xx-tei_adapter/README.md
```diff
@@ -0,0 +1,95 @@
...
+This is an approach we already use in the Sierra adapter and in the Calm adapter.
+The big advantage of this approach is that we control what happens if one request fail, how we should be notified
+and what action we should take.
+
+The proposal in this RFC is to use a pull approach so that we can be confident that all updates have either gone through
+successfully, or we are notified of any failure, and we can retry.
+
+This is a diagram of the proposed architecture: ![](architecture.png)
+
+The Lambda in the picture should generate windows and send them to the TEI adapter
```
kenoir, line 71: Are we reusing existing lambdas to generate the time-window?

kenoir: Are we concerned (as with the sierra adapter) about missing windows?

alicefuzier: > Are we concerned (as with the sierra adapter) about missing windows?

I think we should cause the processing of a window can always fail. We don't have something like the bucket with sierra reader windows to create an alert though, we can only rely on them ending up in the dlq, which means they disappear after a while

alicefuzier: > Are we reusing existing lambdas to generate the time-window?

At the moment the code for the sierra adapter and calm adapter is different and the calm one has a bunch of calm specific things, so there is some work to extract the common bits methinks. Not sure where those common bits should go cause I'm not sure we can extract the whole lambda as the calm one has calm specific stuff

rfcs/0xx-tei_adapter/README.md
alicefuzier: ah right good point, forgot about that. I would leave that to the TEI one since it implies some understanding of the TEI repo structure, rather than just the GitHub api

rfcs/0xx-tei_adapter/README.md
alicefuzier: Clearly, I have spent a lot less time elaborating the reharvest cause it's not that intersting usually. It's a thing that runs once or a handful of times and we don't usually make it fully production quality (they're run locally, they have no tests, no dedicated infra etc). But it is a bit laconic so I will expand on that. I was thinking to use the tree (I didn't know you could get that from the github api, thanks for that), but basically go through each file and send each one to the TEI adapter service, by passing the GIthub service entirely. This bit is a bit crucial to understand what was going on in my head, I realise, and I didn't think to add it 🤦‍♀️

rfcs/040-tei_adapter/README.md
```diff
@@ -0,0 +1,109 @@
+# TEI Adapter
+
+## What is TEI
+
+TEI is an XML standard to represent any source in any script or language. 
+It's used at Wellcome to describe manuscripts in non latin languages.
```
alexwlchan, line 6:
```suggestion
It's used at Wellcome to describe manuscripts in non-Latin languages.
```

rfcs/040-tei_adapter/README.md
```diff
@@ -0,0 +1,109 @@
...
+The TEI files are in a [GitHUb Repository](https://github.com/wellcomecollection/wellcome-collection-tei). At the 
+time of writing in April 2021 there are 622 TEI files, but the collection of manuscripts is much larger, around 10k.
+
+Currently, all information stored in TEI files is not surfaced anywhere except the GitHub repository. 
+Many of the TEI files have a dummy Sierra record that they link to, but not all
```
alexwlchan, line 11:
```suggestion
Many of the TEI files have a Sierra record that they link to, but not all of them – and the record contains minimal information. 
```

rfcs/040-tei_adapter/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+We need to be able to propagate updates to TEI files as they happen through to the API, and we also need to be able to
+reindex when we make changes in the pipeline. 
+
+GitHub is a relatively reliable and low latency source, so we could in theory just send the changes to a TEI transformer 
+and do a reindex by cloning the repository and/or parsing the tree of the TEI files. 
+There are some disadvantages to this approach:
+- We need a "special" reindexer for TEI.
+- We can't keep track of deleted files. If something is deleted, the change will go through the pipeline once, 
+  but there's no way of "remembering" it.
```
alexwlchan, line 23: I think I get what you're saying, but it's almost the opposite problem – we'll remember it forever, we can't forget it.

rfcs/040-tei_adapter/README.md
```diff
@@ -0,0 +1,109 @@
...
+For these reasons, the proposal here is to store the whole TEI file in a full VHS.
+### Store structure
+
+Therefore, taking as an example
+[WMS_Arabic_529](https://github.com/wellcomecollection/wellcome-collection-tei/blob/master/Arabic/WMS_Arabic_529.xml), 
+the proposed store record would be:
+
+| Id | Payload | deleted  | version |
+| --- | ------- | -------| --------|
+| WMS_Arabic_529 | ... | false | 1 |
```
alexwlchan, line 50: Presumably the payload would be an object in S3?

You'll need to change the stores somewhat (or create your own) to put that deleted metadata in DynamoDB.

alicefuzier: hmm good point, let me see how it's done in calm

rfcs/040-tei_adapter/README.md
alicefuzier: You're right. What I meant to say is that there's no way of remembering that it used to exist. I'll rephrase, thanks

rfcs/040-tei_adapter/README.md
alicefuzier: So the calm vhs has a isDeleted flag in the dynamo table, so I think we can steal whatever code the Calm adapter uses for that.

Yes the payload is the location in s3. I'll write it better

rfcs/040-tei_adapter/README.md
```diff
@@ -0,0 +1,108 @@
...
+For these reasons, the proposal here is to store the whole TEI file in a full VHS.
+### Store structure
+
+Therefore, taking as an example
+[WMS_Arabic_529](https://github.com/wellcomecollection/wellcome-collection-tei/blob/master/Arabic/WMS_Arabic_529.xml), 
+the proposed store record would be:
+
+| Id | Payload | deleted  | version |
+| --- | ------- | -------| --------|
+| WMS_Arabic_529 | [pointer to location in s3] | false | 1 |
```
jamesgorrie, line 50: Just a question on this ID, and around moving of items.

It seems like the XML does have an ID: `manuscript_16172` - would it not be better to use that so that if they change the filenames, we don't delete and create new works risking losing access to URLs?

alicefuzier: I picked what is in `<idno type="msID">WMS_Arabic_529</idno>` because @adrianplau said that's the id used for TEI (if I understood correctly)>. I'm not actually sure what that is, it might make reasoning about TEI a bit difficult if we pick something that is not the actual id. Would be good to have @adrianplau input on which one is the best id

jamesgorrie: Great! I missed that piece. My concern was exactly that, what is the idempotent identifier.

jamesgorrie (approved):

rfcs/040-tei_adapter/README.md
adrianplau: Sorry, this one actually doesn't have a Sierra number attached to it yet! The xml id (manuscript_16172) comes from the Fihrist database, to which we also contribute our Arabic manuscript files. We're working on updating the Arabic files with Sierra numbers now and there might already be a batch of these coming later this week.

adrianplau: Quick update: None of the Arabic files have the Sierra number entered yet, but we'll try to get this up relatively soon.

alicefuzier: Actually, I don't think it would be a good idea to use Sierra ids to store TEI because it would prevent us from storing files that don't have a sierra id. we will definitely need a sierra id for linking but that happens in another part of the pipeline. To store our replica of the TEI files, I think we should use whatever is the id in TEI world, which is why I went for `<idno type="msID">WMS_Arabic_529</idno>`. is that correct? Or would @jamesgorrie suggestion be more accurate?

adrianplau: Sorry, then I understand! I think either is fine then. Following on James' suggestion, every TEI file does have a unique xml:id attribute to be valid, while the <idno type="msID"> tends to refer to the now canonical reference ID in the catalogue. In most cases, these overlap (though the Arabic ones have different xml:id attributes), but the xml:id attributes tend to be more consistent in format and are less likely to be edited. On the other hand, if we contribute to other catalogues, such as with the Arabic ones, there is the chance that there will be even more variation between <idno type="msID"> and <TEI xml:id="">. So perhaps stick with <idno type="msID">, as you suggest? Sorry again, this turned out to be a ramble!

alicefuzier: I guess necessary conditions for a field to be useful as an id are:
- It's unique across all TEI files. Ie no 2 files have the same value
- All files have exactly one value for that field. 
- It's not likely to change once created.
If the format varies across files it's unfortunate but less catastrophic than if we pick something that could be edited.
From what you're saying it looks like `xml:id` might be better as `<idno type="msID">` could be edited @adrianplau? or I'm not following 100%.

adrianplau: Thank you, this is helpful. Yes, then I agree that <TEI xml:id=""> is better, as it ticks all three boxes. <idno type="msID"> could be edited if we were to update the cataloguing metadata with new references (this is what we see traces of in the <altIdentifier> in <msIdentifier>.

alicefuzier: Thanks @adrianplau, that is really useful. I've amended the document to reflect this

PR #58, opened 2021-05-04
Add a proposal for the list of statuses we'll present through the items API

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5160

Rendered version: https://github.com/wellcomecollection/docs/blob/rfc-for-item-statuses/rfcs/041-item-statuses/README.md

Currently: the items API returns the item status pretty much as we get it from Sierra. This includes terms that don't make sense to users, and which I don't think we'd want to display on a page. e.g. _"Billed not paid"_, _"CL Returned"_, _"In transit"_.

The proposal: we create our own list of statuses, and map the Sierra statuses onto those. That list would use the same terms we plan to use on the works pages, so we're consistent throughout. This RFC describes how we could do that.

This affects the whole thing, so I've tagged a bunch of people for review.

### DIFF
--- a/rfcs/041-item-statuses/README.md
+++ b/rfcs/041-item-statuses/README.md
@@ -1,0 +1,34 @@
+# RFC 041: Statuses in the items API
+
+As part of the [requesting API](../039-requesting-api-design), we need to present an item status.
+We already have a model that we'll use:
+
+```json
+"status": {
+  "id" : "available",
+  "label" : "Available",
+  "type": "ItemStatus"
+}
+```
+
+but we need to decide what value it can take.
+
+This RFC proposes a set of values that we will present through this API.
+
+
+
+## Guiding principles
+
+1.  **We should not assume requests are sent to a particular service.**
+
+    Although the initial implementation will be making requests for items in Sierra, we may add support for requesting from other systems in a future update.
+    We should not assume the use of Sierra, or of Sierra item statuses.
+
+2.  **A user's list of holds is private.**
+
+    Alice should not be able to find out what items Bob has placed on hold, and vice versa.
+
+3.  **The status information will be publicly visible, and displayed on the works page.**
+
+    For consistency, we should use the same terminology as will be used on the works page.
+    We should assume users will see these statuses, and not expose information about the library workings.
--- a/rfcs/041-item-statuses/README.md
+++ b/rfcs/041-item-statuses/README.md
@@ -207,19 +207,36 @@ Instead, we will create a new set of statuses, which will be mapped as follows:
     <td></td>
   </tr>
   <tr>
+    <td>closed</td>
+    <td>Closed</td>
+    <td>
+      <code>h</code> / Closed <br/>
+    </td>
+    <td>
+      We distinguish between closed/restricted/unavailable in the access statuses.
+      I'm assuming it's a useful distinction here, but we could collapse this into "Unavailable" if not.
+    </td>
+  </tr>
+  <tr>
+    <td>restricted</td>
+    <td>Restricted</td>
+    <td>
+      <code>6</code> / Restricted <br/>
+    </td>
+    <td>
+      Ditto above.
+    </td>
+  </tr>
+  <tr>
     <td>unavailable</td>
     <td>Unavailable</td>
     <td>
-      <code>h</code> / Closed <br/>
       <code>r</code> / Unavailable <br/>
-      <code>6</code> / Restricted <br/>
       <code>m</code> / Missing <br/>
       <code>x</code> / Withdrawn
     </td>
     <td>
-      We distinguish between closed/restricted/unavailable in the access statuses.
-      Is that a useful distinction here?<br/><br/>
-      We map the "Missing" status to "Unavailable" in the access statuses.
+      I assume we wouldn't want to advertise if an item was missing.
     </td>
   </tr>
   <tr>
--- a/rfcs/041-item-statuses/README.md
+++ b/rfcs/041-item-statuses/README.md
@@ -43,7 +43,7 @@ Sierra presents the item status in the [Item STATUS field](https://documentation
 There are standard codes supplied as part of Sierra, but libraries can define additional item status values.
 
 This is a list of item statuses in use at Wellcome, based on the reporting cluster data.
-(Deleted and suppressed items have been )
+(Deleted and suppressed items have been omitted from this analysis.)
 
 <table>
   <tr>
--- a/rfcs/041-item-statuses/README.md
+++ b/rfcs/041-item-statuses/README.md
@@ -288,6 +288,7 @@ Instead, we will create a new set of statuses, which will be mapped as follows:
     </td>
     <td>
       This is arguably only useful as long as COVID lasts, but it wouldn't be too tricky to add if we wanted it.
+      Normally this status is "in transit", but it seems to have been renamed temporarily.
     </td>
   </tr>
   <tr>
@@ -305,3 +306,6 @@ Instead, we will create a new set of statuses, which will be mapped as follows:
     </td>
   </tr>
 </table>
+
+As with the Sierra locations, we will match on the text description of the status, rather than the exact codes.
+This will make us a bit less coupled to the Sierra implementation.
--- a/rfcs/041-item-statuses/README.md
+++ b/rfcs/041-item-statuses/README.md
@@ -32,3 +32,259 @@ This RFC proposes a set of values that we will present through this API.
 
     For consistency, we should use the same terminology as will be used on the works page.
     We should assume users will see these statuses, and not expose information about the library workings.
+
+
+
+## Existing Sierra statuses
+
+This section describes the existing status data in Sierra items.
+
+Sierra presents the item status in the [Item STATUS field](https://documentation.iii.com/sierrahelp/Default.htm#sril/sril_records_fixed_field_types_item.html#STATUS).
+There are standard codes supplied as part of Sierra, but libraries can define additional item status values.
+
+This is a list of item statuses in use at Wellcome, based on the reporting cluster data.
+(Deleted and suppressed items have been )
+
+<table>
+  <tr>
+    <th>code</th>
+    <th>name</th>
+    <th>count</th>
+    <th>comment</th>
+  </tr>
+  <tr>
+    <td><code>-</code></td>
+    <td>Available</td>
+    <td>616,145</td>
+    <td>
+      The Sierra docs say this status is "items that can circulate", and that "the item retains this status even when checked out".
+      We should test with this use case, and make sure the items API responds appropriately.
+    </td>
+  </tr>
+  <tr>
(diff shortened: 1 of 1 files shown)


rfcs/041-item-statuses/README.md
```diff
@@ -0,0 +1,311 @@
...
+
+## Existing Sierra statuses
+
+This section describes the existing status data in Sierra items.
+
+Sierra presents the item status in the [Item STATUS field](https://documentation.iii.com/sierrahelp/Default.htm#sril/sril_records_fixed_field_types_item.html#STATUS).
+There are standard codes supplied as part of Sierra, but libraries can define additional item status values.
+
+This is a list of item statuses in use at Wellcome, based on the reporting cluster data.
+(Deleted and suppressed items have been )
```
jamieparkinson, line 46: Have been...?!

rfcs/041-item-statuses/README.md
```diff
@@ -0,0 +1,311 @@
...
+    <td>Available</td>
+    <td>616,145</td>
+    <td>
+      The Sierra docs say this status is "items that can circulate", and that "the item retains this status even when checked out".
+      We should test with this use case, and make sure the items API responds appropriately.
+    </td>
+  </tr>
+  <tr>
+    <td><code>b</code> / <code>c</code></td>
+    <td>As above</td>
```
jamieparkinson, line 66: Optimism: can we infer what "above" was/is?

jamieparkinson (commented):
Wider question: did we ever get anywhere discussing how item statuses might - or might not - relate to work `availabilities`?

rfcs/041-item-statuses/README.md
alexwlchan:
```suggestion
(Deleted and suppressed items have been omitted from this analysis.)
```

Oops.

rfcs/041-item-statuses/README.md
alexwlchan: Ish.

We can't get the item ordering from the Sierra API, so we can't tell what the "above" item is.

We do have a workaround in the catalogue pipeline, because you get similar "bound in above" or "contained in above" in the location field. We look at all the other items for locations that aren't "X in above", and if there's only one distinct location among them, we assume the "X in above" refers to that, and use that instead.

See https://github.com/wellcomecollection/catalogue-pipeline/blob/master/pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraItems.scala#L55-L90

That's not perfect because:

* Some bibs only have a single item that's "as above", so there are no other items we can look at
* I would be a bit more wary of this for status, where items can presumably move out of sync with each other

I've heard vague mutterings from C&I that they might consider ditching this, and this might be another reason to do so.

rfcs/041-item-statuses/README.md
```diff
@@ -0,0 +1,311 @@
...
+    </td>
+  </tr>
+  <tr>
+    <td>on-holdshelf</td>
+    <td>On holdshelf</td>
+    <td>
+      <code>!</code> / On holdshelf
+    </td>
+    <td>
+      Is the "shelf" part important?
```
DominiqueMarshall, line 277: I think the 'shelf' part is there to indicate that it's physically on the enquiries desk shelf in the library, but I'm not entirely sure if researchers understand that to be the case (regular users of the collections might though). Probably worth checking with the library team.

rfcs/041-item-statuses/README.md
```diff
@@ -0,0 +1,311 @@
...
+  </tr>
+  <tr>
+    <td>closed</td>
+    <td>Closed</td>
+    <td>
+      <code>h</code> / Closed <br/>
+    </td>
+    <td>
+      We distinguish between closed/restricted/unavailable in the access statuses.
+      I'm assuming it's a useful distinction here, but we could collapse this into "Unavailable" if not.
```
DominiqueMarshall, line 217: Having the distinction feels like the right thing to do, especially when considering the different access conditions that be applied to each one.

kenoir (approved):
Nice clear summary & solution.

rfcs/041-item-statuses/README.md
```diff
@@ -0,0 +1,311 @@
...
+    <th>count</th>
+    <th>comment</th>
+  </tr>
+  <tr>
+    <td><code>-</code></td>
+    <td>Available</td>
+    <td>616,145</td>
+    <td>
+      The Sierra docs say this status is "items that can circulate", and that "the item retains this status even when checked out".
+      We should test with this use case, and make sure the items API responds appropriately.
```
jtweed, line 61: Finding out what actually happens to the item status in Sierra when a hold is placed and when the item is available for pickup is crucial here. I know we can't test right now, but we need to find out the answer to that question in order to work out what the utility of item status even is.

rfcs/041-item-statuses/README.md
jtweed: In the meantime do we remove this status and copy the status from the [as best can be determined] "above" item?

rfcs/041-item-statuses/README.md
jtweed: This is an interesting thing to look into further. I know not everything has access conditions, whereas presumably all items have a status. But what is the real difference between them, when from this list there looks to be a considerable overlap.

It could be confusing to display what looks like the same info twice, if it's not clear that one is about usage and the other is about whether you can request the thing.

Ie, do we really use item status as a way to determine whether the thing can be requested, rather than displaying the values directly?

rfcs/041-item-statuses/README.md
```diff
@@ -0,0 +1,311 @@
...
+    </td>
+  </tr>
+  <tr>
+    <td>in-cataloguing</td>
+    <td>In cataloguing</td>
+    <td>
+      <code>p</code> / In cataloguing
+    </td>
+    <td>
+      This feels analogous to the "on order" or "on exhibition" statuses.
```
jtweed, line 267: Indeed and also all of them have an alarmingly large overlap with their equivalent locationTypes. Another example of something we should try to rationalise, rather than display the same info twice.

rfcs/041-item-statuses/README.md
jtweed: Agreed. We should establish when this status is used and then label it in a way that makes sense to users of the library.

rfcs/041-item-statuses/README.md
```diff
@@ -0,0 +1,311 @@
...
+    <td>on-holdshelf</td>
+    <td>On holdshelf</td>
+    <td>
+      <code>!</code> / On holdshelf
+    </td>
+    <td>
+      Is the "shelf" part important?
+      Could this be "on-hold"?<br/><br/>
+      Do we need a way to signal "on hold for the logged-in user"?
+      We don't want to expose the user ID here, but knowing "it's on hold by someone" vs "it's on hold by you" might be useful.
```
jtweed, line 280: I'm going to say no, at least for now. Anything else requires the catalogue API to interact with the identity service and I don't want to go there right now.

rfcs/041-item-statuses/README.md
```diff
@@ -0,0 +1,311 @@
...
+  </tr>
+  <tr>
+    <td>in-quarantine</td>
+    <td>In quarantine</td>
+    <td>
+      <code>t</code> / In quarantine
+    </td>
+    <td>
+      This is arguably only useful as long as COVID lasts, but it wouldn't be too tricky to add if we wanted it.
+      Normally this status is "in transit", but it seems to have been renamed temporarily.
```
jtweed, line 291: If that's the case, I'd probably id it as in transit, but temporarily change the label. As you say, it's not going to last forever.

rfcs/041-item-statuses/README.md
```diff
@@ -0,0 +1,311 @@
...
+    </td>
+  </tr>
+  <tr>
+    <td>permission-required</td>
+    <td>Permission required</td>
+    <td><code>y</code> / Permission required</td>
```
jtweed, line 297: Another good example of something that I'd like to consider as a potential point of intersection with access conditions.

rfcs/041-item-statuses/README.md
```diff
@@ -0,0 +1,311 @@
...
+    <td>Unknown</td>
+    <td>(all others)</td>
+    <td>
+      This would be a catchall for all Sierra statuses we can't map to another status.
+    </td>
+  </tr>
+</table>
+
+As with the Sierra locations, we will match on the text description of the status, rather than the exact codes.
+This will make us a bit less coupled to the Sierra implementation.
```
jtweed, line 311: Honestly, I'm not sure this is a good idea. The text label is more likely to change. We don't do this for eg material types, why would we do it here?

I'd be careful of overgeneralising the fact that we had to do that for locations. I see that as papering over a data quality issue that was too big to take on, I'm not sure that's the case here.

jtweed (changes requested):
Looking at the analysis here, I don't think we can consider this in isolation. I'd like to see an RFC that looked at both item statuses and opac messages, potentially also considering the intersection with access conditions.

rfcs/041-item-statuses/README.md
jamieparkinson: https://www.nature.com/articles/d41586-021-00396-2 please, less optimism

rfcs/041-item-statuses/README.md
kenoir: It is also possible to synthesise this information in the browser client.

rfcs/041-item-statuses/README.md
alexwlchan: Hmm. This is what that looks like in the OPAC:

Screenshot 2021-05-07 at 14 56 40

Notice you aren't allowed to request an item that's "bound in above" – this is set by the "Rules for requesting" file.

Because we can't order these sensible, I'm not sure what's best to do here. Possibly these get folded into "not-requestable"?

rfcs/041-item-statuses/README.md
alexwlchan: BLARGH.

I have a hold on item 1277629. We can see the hold count, but the status stays the same:

```
 'copyNo': 2,
 'holdCount': 1,
 'status': {'code': '-', 'display': 'Available'},
```

Screenshot 2021-05-07 at 15 05 21


Screenshot 2021-05-07 at 15 04 42

alexwlchan:
These are the OPAC messages we get:

```
- -
a By appointment
b @ digitisation
d New books
e Exhib. Gallery
f Online request
i Ask at desk
n Manual request
o Open shelves
p By approval
q Donor permission
s Staff use only
u Unavailable
x --
```

rfcs/041-item-statuses/README.md
jtweed: The problem is they are requestable, they're just requestable by requesting another item.

alexwlchan:
Closing as superseded.

PR #60, opened 2021-05-14
Requesting, access statuses, and whether you can find stuff in the library

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5171

Supersedes #58

Rendered version: https://github.com/wellcomecollection/docs/blob/big-requesting-rfc/rfcs/requesting-model/README.md


pollecuttn:
Fixed field 88 (status) - On search: A member (or staff) tries to find an item, but can't find it. Things that can't be found where they should be are then put on search: a staff member looks for the item a number of times. If it still can't be found, the status is changed to missing. So on search gives an indication that might save someone else time looking for something that we already know might not be where it should be, but it's not gone as far as being deemed missing yet.

pollecuttn:
Strictly, people place holds or requests, not orders, for items.

Orders (and order records) in Sierra relate to acquisition of items (e.g. order a book from a bookseller). So 'orders' is already used for something specific in Sierra, and it's got nothing to do with requesting.

alexwlchan:
You'll notice this RFC only summarises the current state of the world; I’m doing some more work to explain how we'll corral all this data.

kenoir:
What does ǂ mean in this context?

alexwlchan:
> What does ǂ mean in this context?

It's one of the characters used for a subfield, I'll clarify.

kenoir:
Really useful and clear summary of the problem and direction to go in.

jtweed:
This is top notch 👍

jtweed (approved):

rfcs/requesting-model/README.md
```diff
@@ -390,145 +435,72 @@ This is my best understanding of what happens:
...
-
--   Items on open shelves aren't requestable.
-
--   If the status is "As above", there's not much we can do.
-
-The included script will apply this transformation, and flag cases that don't work.
-This rule doesn't work for 98 cases, which covers 9203 items -- these are in the spreadsheet [unhandled.csv](unhandled.csv).
-
-We should work with Collections to either define rules or fix the item data.
+    > Question: Do we need the distinction between Unavailable and NotRequestable?
```
kenoir, line 506: There is a difference as you explain - I guess it's a UX issue (ping @DominiqueMarshall) if we want to distinguish "might be available later (can't tell you when though)" and "not available ever (though I suppose we might change our minds later"). Does seem to me like they are practically be the same thing from a user perspective.

DominiqueMarshall: Yeah I would agree. It feels like using both status types could be redundant if the `message` can do the heavy lifting in terms of explaining why researchers can't view/request an item.

PR #61, opened 2021-05-18
Add a brief RFC for tracking changes to Miro data

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5176

Rendered version: https://github.com/wellcomecollection/docs/blob/tracking-changes-to-miro-data/rfcs/041-miro-data-changes/README.md

### DIFF
--- a/rfcs/041-miro-data-changes/README.md
+++ b/rfcs/041-miro-data-changes/README.md
@@ -108,6 +108,8 @@ for image_id in ["A0000001", "A0000002", "A0000003"]:
     suppress_image(image_id, message="We were asked to take these images down; see email from John Smith on 18 May 2021")
 ```
 
+These functions will send a message to the Miro updates topic, so the record gets re-transformed by the Miro transformer.
+
 ### Worked example
 
 Suppose we have the following Miro record:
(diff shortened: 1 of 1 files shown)


kenoir:
This makes sense to me. I don't think this allows us to track changes where multiple changes have happened to the same type if thing e.g. a license changes multiple times, we don't keep track of all changes just the last. Though I don't think we care about having an audit trail, only that the source data is preserved?

kenoir (approved):

alexwlchan:
> I don't think this allows us to track changes where multiple changes have happened to the same type if thing e.g. a license changes multiple times, we don't keep track of all changes just the last. Though I don't think we care about having an audit trail, only that the source data is preserved?

I hadn't thought about tracking multiple changes or old values, but that's a good point.

But you could track this as part of the Python helpers, e.g.

```
set_license_override("A1234", license="cc-by", message="We realised we could make this more permissive")

-> sets MiroSourceOverride
-> adds an update event with the message
"Change license from 'in-copyright' to 'cc-by'. User message: We realised we could make this more permissive"
```

If you updated the license multiple times, you'll get multiple update events.

I'll add an example to clarify.

PR #62, opened 2021-05-20
Change TEI Adapter design

alicefuzier:
So, I've was in the process of implementing the first design using commits from GitHub and suddenly realised that using commits opened a huge can of worms.
So I thought about a comment made by @alexwlchan about using trees instead of commits in the [original pr](https://github.com/wellcomecollection/docs/pull/57). At the time, I thought it was complicated because it needed to store the tree somewhere and chasing deletions could be more difficult, but upon further reflection, I think it's actually much more sensible.

This PR is changing the original RFC to use trees instead and I think it generally simplifies everything. I also implemented part of the Lambda as a proof of concept Jupyter notebook

Rendered version https://github.com/wellcomecollection/docs/tree/tei_adapter_take2/rfcs/040-tei_adapter


rfcs/040-tei_adapter/README.md
```diff
@@ -68,43 +68,117 @@ The biggest disadvantage is that we don't have control or visibility if the Webh
...
+- Retrieves the old tree from S3.
+- Compares the two and for every change it sees in the tree sends a message.
+- Replaces the old version of the tree with the new one in S3
+This is a sample of the message sent by the Lambda:
+```yaml
+{
+  "path": "Arabic/WMS_Arabic_1.xml",
+  "url": "https://api.github.com/repos/wellcomecollection/wellcome-collection-tei/git/blobs/17f54c8c5ebccec643e23b54b3f4277f86204282",
+  "deleted": false 
+}
```
alexwlchan, line 170: Presumably you'd get a difference response if the file has been deleted?

alexwlchan (approved):

kenoir:
What would happen in cases where files are moved but not changed? I'm guessing the sha hash would uniquely identify that file?

rfcs/040-tei_adapter/README.md
```diff
@@ -68,43 +68,117 @@ The biggest disadvantage is that we don't have control or visibility if the Webh
...
+history of the repo, which is very fiddly and error prone.
+
+Another approach is to ask the GitHub API for trees. Trees are a representation of the file hierarchy in the repo at a 
+certain commit. The following request gives us all the files in the repository at the latest commit in master:
+```
+https://api.github.com/repos/wellcomecollection/wellcome-collection-tei/git/trees/master?recursive=true
+```
 
+This is a small sample of the response:
+```yaml
```
jamesgorrie, line 99:
```suggestion
```json
```

rfcs/040-tei_adapter/README.md
```diff
@@ -68,43 +68,117 @@ The biggest disadvantage is that we don't have control or visibility if the Webh
...
+    "url": "https://api.github.com/repos/wellcomecollection/wellcome-collection-tei/git/blobs/17f54c8c5ebccec643e23b54b3f4277f86204282"
+  },
+...
+  ]
+}
+```
+Every entry in the `tree` can be either a `tree` or a `blob`. A `blob` is a file in the repo and has a hash associated to it. 
+If the content of a file changes, the hash changes. 
+This means that we can query for the tree in master at regular intervals and store the last version of the tree in S3. 
+At every request, we can compare the old tree with the new one to know what changes have happened.
```
jamesgorrie, line 150: I'm not sure what the limit is (although [this post suggests quite a lot](https://github.community/t/github-get-tree-api-limits-and-recursivity/1300/2), but I am not sure how big this repo is expected to grow) - but is it worth thinking about?

jamesgorrie (approved):
Makes loads of sense, and seems simple and elegant.

As a point, I noticed you were using python, they do have an [ocotokit.js](https://github.com/octokit/octokit.js) version (no python unfortunately) which is typed which we can use for lambdas.

rfcs/040-tei_adapter/README.md
alicefuzier: ah yes, I know. I did this on purpose because I prefer how the json is rendered when markdown thinks is yaml 😬

rfcs/040-tei_adapter/README.md
alicefuzier: I should have added in the rfc (and I will, thanks). I've checked and now it looks like the limit is 100k files and 7mb in the response. It's not documented in the github api though, I suppose because they want to reserve themselves the opportunity to change that. Tentatively trying, you can use the same endpoint to retrieve all the files in the catalogue-pipeline repo with no issues in one request, so I think we're covered for tei. The entire manuscript collection is about 10k so even if everything is described in tei tomorrow we should still fit within those limits. Anyhow, if the response is truncated, there is a `truncated` flag that we should defnitely check, and implement a more recursive tree fetching strategy. Even in that case, we know the size of the repo so we can always calculate how many requests we will need and it's probably below 10

alicefuzier:
> What would happen in cases where files are moved but not changed? I'm guessing the sha hash would uniquely identify that file?

I was thinking about it too. You can't reliably know in the lambda if something is moved but not changed or only slightly changed, I don't think. But because we use the `xml:id` attribute to identify the in the TEI adapter service, I think we can spot then that something is the same file moved (changed or not changed). The problem there is at that point, for everything moved, the lambda would have sent two messages: a deleted message and a file changed message (new file is represented the same as file changed). Then when the TEI adapter service processes those messages, the result is different depending on the order those two messages are processed:
 - file deleted is processed first and file changed after:  this is fine and results in the changes to the contents of the file being sent to the transformer.
 - file created is processed first and file deleted after: this is problematic, we would make things disappear from the api. 
 
 I think we can prevent the second case by sending a timestamp or the sha of the commit in the message, so that we can prevent a delete message undoing the result of a file changed message with the same sha/timestamp.

alicefuzier:
> I think we can prevent the second case by sending a timestamp or the sha of the commit in the message, so that we can prevent a delete message undoing the result of a file changed message with the same sha/timestamp.

I will add a bit about this. I didn't want to go too much in detail on this because I expect that the format of the message and vhs record will change as I implement this. I imagine we have to carry some more info to deal with the file moved case, but at this stage I'm more concerned with establishing that it is solvable, rather than solving it in detail.

alicefuzier:
> As a point, I noticed you were using python, they do have an ocotokit.js version (no python unfortunately) which is typed which we can use for lambdas.

No idea what ocotokit is...Is it some github library? I don't think we need a specific github library to be honest, the requests that we do are prretty simple that just the plain http library will do I think

rfcs/040-tei_adapter/README.md
alicefuzier: yes, will change, thanks

rfcs/040-tei_adapter/README.md
jamesgorrie: Great, saw this implemented too.

jamesgorrie:
@alicefuzier Octokit is the clients for the GitHub API:
https://github.com/octokit

alicefuzier:
Thanks @jamesgorrie, will check it out. Yes, agreed I like types too. 
But in this instance, we have a whole load of infrastructure for building python lambdas that I don't really want to recreate for js. Also, all of our Lambdas are Python and I'm not sure it would be a good idea to have some in python and some in js. 
In this instance, I don't think a library and types would bring a massive advantage because the lambda is very very simple, as all of our lamdas generally are. As long as we keep them simple and very very focused on one problem, I think python is perfectly suited (can't believe I'm saying that).

jamesgorrie:
@alicefuzier 

> Also, all of our Lambdas are Python and I'm not sure it would be a good idea to have some in python and some in js.

[We do use JS / TS edge lambdas on the CloudFront edge](https://github.com/wellcomecollection/platform-infrastructure/tree/main/cloudfront/wellcomelibrary.org/edge-lambda) 

Super keen to use whatever - just spreading the news... 📰

jamesgorrie (approved):

PR #64, opened 2021-07-21
Add an RFC about removing deleted records from (re)indexes

alexwlchan:
Rendered version: https://github.com/wellcomecollection/docs/blob/soft-deletions/rfcs/043-recording-deletions/README.md

Follows the Slack thread here: https://wellcome.slack.com/archives/C3TQSF63C/p1626854060144700

I think we could implement this fairly quickly – a couple of hours, maybe – and start getting faster, cheaper reindexes.


kenoir (approved):
Needs some eyes on by @jtweed to confirm 404 vs 410 if we plan to ignore deletions within the lifetime of the Catalogue API.

rfcs/043-recording-deletions/README.md
```diff
@@ -0,0 +1,90 @@
...
+## Risks and mitigations
+
+-   **410 Gone would become 404 Not Found for any Works that we moved.**
+    If we don't put deleted Works in the final Elasticsearch index, the API will serve them as 404 Not Found instead of 410 Gone.
+    This is potentially annoying for anybody who has a URL to a work that was previously available.
+
+    As a first pass, we can avoid this by only removing Works that were deleted before the new catalogue API/website -- and so never had public URLs with the current URL scheme.
+    If you filter to anything deleted before 1 Jan 2017, there are 876k Works (27% of the 3.2M total Works), and so still a significant saving.
+
+    After that, we could move records to the "deleted" table on a rolling window -- say, anything deleted more than a year previously.
```
jamesgorrie, line 67: This is the piece that still concerns me in the sense of we still land up with 404 that should be 410s if they're a year old. Maybe I am being a pedant about this, but it feels we could have an publically accessible version of 410s somehow and this statement seems to suggest we just shouldn't.

jamesgorrie: It could even mean that we have a separate pipeline or similar setup, which could even put deleted works into a separate index, which we can search.

jamesgorrie (commented):
Comment on 410/404s I wonder if we should be more clear about - even if more clear is saying we don't know but won't make any assumptions yet.

A thought as well is that we _should_ see a search improvent, although not significantly as this is used at filter context.

I wonder if we could think about generlising this to the other non `Visible` works?

rfcs/043-recording-deletions/README.md
```diff
@@ -89,3 +89,8 @@ We're not permanently deleting any data, and we could revert it later if there a
 
 -   **Some as-yet-unforeseen issue.**
     If this doesn't work in practice, we reverse the process: merge the "deleted" table back into the "live" table, remove the "deleted" table, and we're back at our current setup.
+
+## Possible future extensions
+
+-   We could move more records to the "deleted" table if we're happy to live with the 404/410 distinction.
```
jamesgorrie, line 95:
```suggestion
We could move more records to the "deleted" table if we're happy to live without the 404/410 distinction.
```

?

jamesgorrie (approved):

PR #67, opened 2021-08-20
RFC about internal model discussion

alicefuzier:


jamieparkinson:
Superseded by the inexorable march of time

PR #72, opened 2022-02-08
RFC 044: Patron Deletions

jamieparkinson:
Rendered version: https://github.com/wellcomecollection/docs/blob/f143168ccbc0a4285d5090d777c89d1b51c67eb7/rfcs/044-patron-deletions/README.md

### DIFF
--- a/rfcs/044-patron-deletions/README.md
+++ b/rfcs/044-patron-deletions/README.md
@@ -14,8 +14,8 @@ Sierra does not provide a change feed or any other kind of notification for even
 
 ## Proposal
 
-1. A **CloudWatch event** is run at some interval *T* to trigger a Lambda
-2. The **Lambda** queries the Sierra API for patron records<sup>†</sup> that were deleted in the previous time window *W > T*. That is to say, time windows overlap to avoid records getting lost at the boundary.
+1. A **CloudWatch event** is run daily to trigger a Lambda
+2. The **Lambda** queries the Sierra API for patron records<sup>†</sup> that were deleted at some point on the _previous_ day - so as to ensure the entire day is covered.
 3. The Lambda calls the [**Auth0 Management API**](https://auth0.com/docs/api/management/v2#!/Users/delete_users_by_id) to delete each of the records found in the query. If these are already deleted, or don't have any data in Auth0, this query 404s (which is fine).
 
 <sup>†</sup> The Sierra API allows us to query by a `deletedDate` like `yyyy-MM-dd`, or an `updatedDate` like `yyyy-MM-dd'T'HH:mm:ssZZ` along with a `deleted: true` filter. While `deletedDate` has poor granularity, we should use it rather than `updatedDate`, as a deletion does not count as an update: the second option would not return records that were recently deleted but not otherwise updated.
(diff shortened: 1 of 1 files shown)


rfcs/044-patron-deletions/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+While Sierra is the source of truth, Auth0 does store a minimal amount of patron data (email, barcode, patron number). We need to be able to get rid of the records in Auth0 when a record is removed from Sierra, for GDPR reasons as much as anything else.
+
+Sierra does not provide a change feed or any other kind of notification for events, so we need a polling solution. We already do this for items and bibs [in the Sierra adapter](https://github.com/wellcomecollection/catalogue-pipeline/tree/main/sierra_adapter#readme), by polling for changes in a given time window. We do not need the complex architecture of the Sierra adapter for tracking patron deletions: we don't need any information from within the records, and we don't need to merge anything. 
+
+## Proposal
+
+1. A **CloudWatch event** is run at some interval *T* to trigger a Lambda
+2. The **Lambda** queries the Sierra API for patron records<sup>†</sup> that were deleted in the previous time window *W > T*. That is to say, time windows overlap to avoid records getting lost at the boundary.
```
paul-butcher, line 18: Given that the relevant query only has date granularity, I'm not sure this is the right way to think of the Window/Interval relationship.

As long as we regularly run it against completed days, there is no boundary in which things may get lost.

Because the granularity of the data we are looking at is per-day, then the logical frequency would also be daily, e.g. run it in nightly in the wee hours to bring yesterday up to date.

If we do run it against today, e.g. because we want to be more responsive to changes than daily, then we will also need to run the whole day again once it has completed, in order to be sure that we catch everything.  Running against today is always going to be a window anchored on midnight.

jamieparkinson: Ah, good point - hadn't been thinking about it in terms of acting on the previous day, and wanted the overlapping windows (as in the Sierra adapter) because I'd imagined a case where you'd miss deletions that happened on a given day after the service had run. A non-issue if just looking at the previous day - will update.

rfcs/044-patron-deletions/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+1. A **CloudWatch event** is run daily to trigger a Lambda
+2. The **Lambda** queries the Sierra API for patron records<sup>†</sup> that were deleted at some point on the _previous_ day - so as to ensure the entire day is covered.
+3. The Lambda calls the [**Auth0 Management API**](https://auth0.com/docs/api/management/v2#!/Users/delete_users_by_id) to delete each of the records found in the query. If these are already deleted, or don't have any data in Auth0, this query 404s (which is fine).
+
+<sup>†</sup> The Sierra API allows us to query by a `deletedDate` like `yyyy-MM-dd`, or an `updatedDate` like `yyyy-MM-dd'T'HH:mm:ssZZ` along with a `deleted: true` filter. While `deletedDate` has poor granularity, we should use it rather than `updatedDate`, as a deletion does not count as an update: the second option would not return records that were recently deleted but not otherwise updated.
+
+## Questions and potential issues
+
+- What happens if there are more deleted records than can be processed at once by the Lambda? *Mitigation: increase Lambda timeout to maximum*.
```
alexwlchan, line 25: We can also just delete the records from Auth0 ourselves.

alexwlchan (approved):

PR #73, opened 2022-02-18
Sierra Work relationships

paul-butcher:
This PR includes the original RFC from alicefuzier in a subdirectory, as it provides some useful context.

My current proposal (the part under review here) is in [rfcs/045-sierra-work-relationships/README.md](https://github.com/wellcomecollection/docs/blob/sierra-series-updated/rfcs/045-sierra-work-relationships/README.md)

### DIFF
--- a/rfcs/045-sierra-work-relationships/README.md
+++ b/rfcs/045-sierra-work-relationships/README.md
@@ -414,7 +414,8 @@ punctuation to permit this, e.g. the colon and semicolon at the ends of subfield
 `830 Published papers (Wellcome Chemical Research Laboratories) ;|vno. 149.`
 `774 0  |gPage 6 :|tCharing Cross Hospital: a portrait of house surgeons. Photograph, 1906.|w(Wcat)28916i`
 
-Ideally, these separators would not be presented unless the fields are output together.
+Ideally, these separators would not be presented unless the fields are output together.  We will need to iterate
+towards the correct way to trim them.
 
 Note that the order of subfields is defined in the data, so the full string cannot be simply reconstructed from 
 individual fields.


rfcs/045-sierra-work-relationships/README.md
```diff
@@ -0,0 +1,377 @@
...
+as a Series link.
+
+If the title of a 773 field matches one of the Series fields, but it also
+contains a $w subfield, then 
+__Investigate where this happens in order to decide, before developing a solution__
+ * Find out if it ever happens
+ * Find out how many objects there are in each "series" where it does happen
+ * Ask C&R what it means
+ 
+### subPart property name
```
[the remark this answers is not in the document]
paul-butcher: I also don't love the exact name, so bike shed away if you have any better ideas

rfcs/045-sierra-work-relationships/README.md
```diff
@@ -0,0 +1,377 @@
...
+### Other potential subPart subfields
+
+There are also subfields $p (Name of part/section of a work) and $n (Number of part/section of a work)
+Both of these are present on at least one example:
+
+[Madame Delait, the bearded woman of Plombières](https://search.wellcomelibrary.org/iii/encore/record/C__Rb1532371__Sfallaize__P0,2__Orightresult__U__X3?lang=eng&suite=cobalt&marcData=Y)
+
+However, in this case, the values do not look useful:
+
+`830  0 Fallaize Collection.|pname of grouping ;|nnumber`
```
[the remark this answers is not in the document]
paul-butcher: Yes. With these, I fancy that the right thing to do (at our end) is not worry about them yet, but bear in mind that they do exist and may hold useful information.

We can combine that with asking whether they do ever hold useful information, so that we can include it if we need to in the future.

In terms of linking, because this is (at best) another subPart, it falls under step 4 - display (or use, as your other comment suggests) the subPart information

rfcs/045-sierra-work-relationships/README.md
```diff
@@ -0,0 +1,377 @@
...
+      "id": "[canonical id derived from](Wcat)28924i"
+      "subPart": "Page 7"
+      "title": "Charing Cross Hospital: a portrait of house surgeons. Photograph, 1906."
+    }
+    ...
+  ]
+```
+## Unanswered Questions
+
+### 774 without id
```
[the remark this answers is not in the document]
paul-butcher: Yes. good point. If there is a symmetrical 773/774 by titles, then there is no need to worry about matching on ids.

rfcs/045-sierra-work-relationships/README.md
```diff
@@ -0,0 +1,377 @@
...
+
+
+## Common features
+
+### Most Precise Wins
+
+The 773 field is not restrictive about what it can point to, whereas the 4x0 and 830 
+fields explicitly refer to a Series.  Records with a 773 field that matches
+one of the series fields should ignore the 773 field and just present it
+as a Series link.
```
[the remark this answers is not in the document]
paul-butcher: Yes. Normally they are in a 773/774 pairing.  In that case it should become a Work->Work relationship.

This section is about the "winning" concept where they do match.  It may be a bit redundant, because the likelihood is that a 773 matching a Series field will not have a $w, so it would be treated as a Series anyway.

It is possible that there is no 4x0 or 830 _or even_ corresponding 774.  That is also expected to behave like a series. This is described below see the "James Gardiner Collection" example.

Despite the possible redundancy, I wanted to include this section to ensure that the situation below has been considered.

rfcs/045-sierra-work-relationships/README.md
```diff
@@ -0,0 +1,377 @@
...
+##### Why not store it all in title?
+
+Storing the series title and subPart in the same field makes this difficult.
+
+One could expect there to be only one "Published papers (Wellcome Chemical Research Laboratories) ; no. 149",
+and at least 149 "Published papers (Wellcome Chemical Research Laboratories)". One would
+also expect there to be many Series with titles containing most of those tokens.
+
+* A search for all those tokens would return too many results (anything containing Published, or papers, or Wellcome etc.)
+* A search for the exact string would return only one result (i.e. this one)
```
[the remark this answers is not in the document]
paul-butcher: It may be that we end up presenting two links. I'm not sure what the filter would look like. Perhaps we would index the whole field as one, but return the two separately in the API.

rfcs/045-sierra-work-relationships/README.md
jtweed, line 81: Whilst I can see the potential need for this property, I think it should be called `partNumber` not `subPart`. There is even a potential argument for a separate `partName` but I think given the examples we have `partNumber` is more appropriate. Note that in MARC a part number does not have to be a literal number, it's usually a string that describes a numbered part eg "Volume 1".

The rationale for the naming is that it aligns better with the possible parts of a title in MARC `245`, which include `$p part number` and `$n part name`. We have previously had these separate properties, before we decided to just roll them into the `title`. These are just qualifiers for the title, meaning we should follow the conventions in `245`.

These are also the names used in `830` and `430 $v` is broadly equivalent to `830 $n`.

rfcs/045-sierra-work-relationships/README.md
```diff
@@ -0,0 +1,377 @@
...
+  }
+]
+```
+
+A MARC value may have a subfield denoting a "part" of the series, this should
+be separated from the series title, and presented in a new field, `subPart`
+
+In the case of 830, 430 nd 490 fields, this is in the $v subField.
+
+`830 Published papers (Wellcome Chemical Research Laboratories)
```
jtweed, line 149: This example demonstrates a key aspect of MARC, that subfields are generally written to be concatenated in order and contain the punctuation required to do so.

That means in this instance we'd really need to strip the ` ;`. However, past experience shows we need to be a bit careful doing that, as it's not always so straightforward. We'd also then have to consider how we display the separate parts when we join them back together. And if we are just going to join them back together, it may make sense not to split them in the first place...

Most of the times we've split in the past, including `title` and `physicalDescription`, we've ended up backing it out and just combining the MARC as intended. I think given that, there's a strong argument to start with the combined title, then look again at this as a potential improvement if the shape of the data shows that it's necessary.

jtweed (changes requested):

rfcs/045-sierra-work-relationships/README.md
```diff
@@ -0,0 +1,377 @@
...
+774 1  |tLists of varieties of rice and rattan.|h2 p.; 34 x 21 cm
+774 1  |tNames of plants.|h2 p.; 34.5 x 20 cm 
+774 1  |tLists of plant names and uses.|h3 p.; 34.5 x 21.5 cm 
+774 1  |tLists of grasses collected in Malacca.|h13 p.; 34 x 21 
+       cm 
+774 1  |tNames of plants collected at Mĕrliman.|h9 p.; 34 x 21 cm
+
+```
+
+### Other potential subPart subfields
```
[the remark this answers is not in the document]
paul-butcher: Yes.  Much better naming

rfcs/045-sierra-work-relationships/README.md
paul-butcher: Although MARC allows it, I'm a bit uncomfortable with exposing a field called somethingNumber that is not numeric.

In these examples, it does fit - volume X, page X etc.  But will that always be the case?

Although the source for the data we are doing this for is Sierra (which allows Numbers like this that are not purely numeric, but ) might we also end up using this field for data from other sources, where the partName could be anything?

rfcs/045-sierra-work-relationships/README.md
paul-butcher: OK.  So we could say 

Step 1: include the whole title including subfields
Step 4: crack the title into its constituent subfields
Step 5: work out what to do with the subfields as distinct from the main title.

Doing this would not give us the advantage described here: https://github.com/wellcomecollection/docs/blob/4cb54a95884a8b0821ceb0f73b7284e8976d1fbb/rfcs/045-sierra-work-relationships/README.md?plain=1#L239-L249 until we do crack the subfields, but we would reach the MVP quicker as we wouldn't have to work out those punctuation kinks early on.

As it links to a search, the likely User Journey would be:
0. From the record - 
1. Click on the Series Link e.g. "Published papers (Wellcome Chemical Research Laboratories) ; no. 149"
2. See the search page with one item on it (the one we have just come from)
3. manually remove the suffix from the search query and search again

However, this raises UI issue.  This should ideally result in a search filtered on a specific field, rather than full text, and all our existing field-based filters are checkboxes.  To facilitate this, we would have to allow free-text fielded search.

Perhaps the first step then is to link to a full-text search for the series title, then later refine this to search in the appropriate field.

paul-butcher: Drafting this response, I discovered this bug. https://github.com/wellcomecollection/platform/issues/5436

paul-butcher: Following discussions here: https://wellcome.slack.com/archives/C3TQSF63C/p1645440176157099

It looks like we have to split the title from the beginning.

rfcs/045-sierra-work-relationships/README.md
```diff
@@ -0,0 +1,377 @@
...
+fields explicitly refer to a Series.  Records with a 773 field that matches
+one of the series fields should ignore the 773 field and just present it
+as a Series link.
+
+If the title of a 773 field matches one of the Series fields, but it also
+contains a $w subfield, then 
+__Investigate where this happens in order to decide, before developing a solution__
+ * Find out if it ever happens
+ * Find out how many objects there are in each "series" where it does happen
+ * Ask C&R what it means
```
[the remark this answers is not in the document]
paul-butcher: Yes, I included this because it is logically possible, I am yet to investigate whether such a thing actually exists.  If it does, then getting it cleaned sounds like the most likely outcome.

rfcs/045-sierra-work-relationships/README.md
paul-butcher: I have added a section in Unanswered Questions about working out how to deal with subfield separator punctuation.

rfcs/045-sierra-work-relationships/README.md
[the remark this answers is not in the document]
jtweed: You'll note that Encore only searches for `Early European Books : Printed sources to 1700` and doesn't include the `Collection 15`. I'm not saying it might not be _better_ to allow searching for both, but I do think that it's not necessary at this stage.

Volumes are often, but not always, a meaningless construct when it comes to research as they are primarily an organisational construct to help the publisher. That's the "how do we know problem", so I'd agree we make this a problem for the future, should it prove to be one.

rfcs/045-sierra-work-relationships/README.md
jtweed: Don't think it matters which one we go for here, so I'm happy with using partName, but just wanted to note that:

> Although MARC allows it, I'm a bit uncomfortable with exposing a field called somethingNumber that is not numeric.

is obviously also using a specific definition of number. Library catalogues built up over decades/centuries often use much looser concepts than we might starting today. However, there are also case where the above is also just flat out wrong in the context of library collections.

The canonical example of that would be page numbers, which are not guaranteed to be numeric, consecutive, sequential or present. And in every possible combination of the above, sometimes within the same item!

rfcs/045-sierra-work-relationships/README.md
jtweed: I see this as a "Series" filter, that does a keyword match on the title of the series (whether that includes the volume tbd). That will then appear in the "applied filters" section of the search bar, not as text in the search field.

Does look like we'll want that to just be on the series title, not the volume, making the partNumber/partName something we should probably do.

alexwlchan (approved):
I haven't read every bit of your discussion with Jonathan, but it looks like you've reached a consensus – this seems like a good basis for starting to implement something! 🎉

PR #74, opened 2022-03-21
RFC: Secrets in Lambdas

jamieparkinson:
Rendered version: https://github.com/wellcomecollection/docs/blob/lambda-secrets-rfc/rfcs/046-lambda-secrets/README.md

### DIFF
--- a/rfcs/046-lambda-secrets/README.md
+++ b/rfcs/046-lambda-secrets/README.md
@@ -1,0 +1,49 @@
+# RFC 046: Secrets in Lambda functions
+
+**Status:** Draft
+**Last updated:** 21/03/2022
+
+## Context
+We have historically used AWS's per-invocation "serverless" environment, Lambda, as a kind of glue between more complex applications or as part of our monitoring stacks. Increasingly, we have been using it for complete applications: most notably, for the Identity API.
+
+With this increased usage, our Lambdas have been running more privileged workloads, and the existing approach to secrets - that is, unencrypted environment variables - has become inadequate. While ECS [provides](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/specifying-sensitive-data-secrets.html) a method for injecting secrets as environment variables at container startup, no such built-in solution exists for Lambda. This RFC proposes one such solution.
+
+#### Desiderata
+As a starting point, we'd like any solution to satisfy the following:
+- *AWS Secrets Manager* as the source for secrets, to match our current usage and to allow explicit permissions on individual secrets via IAM.
+- Secrets are fetched when the Lambda starts: no need to keep them "live".
+- Secrets are injected transparently as environment variables: no requirement for additional/custom APIs or clients within runtimes.
+- Language-agnostic: our solution will work for Lambdas written in any language.
+
+#### Prior art
+1. Lambda includes the feature to [encrypt environment variables at rest](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-encryption). 
+
+   This doesn't account for getting the secrets into the function configuration (ie Terraform still needs to deal with them), or access control, and if we wanted to encrypt in transit we'd need to add decryption code into the Lambda. :x:
+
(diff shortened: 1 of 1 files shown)


rfcs/046-lambda-secrets/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+With this increased usage, our Lambdas have been running more privileged workloads, and the existing approach to secrets - that is, unencrypted environment variables - has become inadequate. While ECS [provides](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/specifying-sensitive-data-secrets.html) a method for injecting secrets as environment variables at container startup, no such built-in solution exists for Lambda. This RFC proposes one such solution.
+
+#### Desiderata
+As a starting point, we'd like any solution to satisfy the following:
+- *AWS Secrets Manager* as the source for secrets, to match our current usage and to allow explicit permissions on individual secrets via IAM.
+- Secrets are fetched when the Lambda starts: no need to keep them "live".
+- Secrets are injected transparently as environment variables: no requirement for additional/custom APIs or clients within runtimes.
+- Language-agnostic: our solution will work for Lambdas written in any language.
```
alexwlchan, line 17: I think it's worth breaking out "must have" vs "nice to have", and discussing the rationale for them – while (3) and (4) are nice, it's not obvious to me that we should require them.

e.g. doing compiled TypeScript and Lambda layers is quite different to what we've done before, which adds a degree of engineering complexity. Another requirement might be "keep it simple".

I'm not saying they're wrong, but let's get the rationale down so we can discuss them together.

jamieparkinson: What do you mean by "the rationale" other than the reasons given here? I feel quite strongly that these are both must-haves...

alexwlchan: Let's talk about this in person.

jamieparkinson: To clarify further (and I shall add this to the RFC):

- Lambda layers/extensions are new to us, but they are [designed](https://aws.amazon.com/blogs/compute/introducing-aws-lambda-extensions-in-preview/) for exactly this kind of use case. I realise that elsewhere (usually in Python scripts) we fetch secrets "manually", but as outlined in the context I think we need a solution for Lambda secrets that treats Lambdas as cattle, not pets.
- The above reasoning applies to the need to be language-agnostic too, but I do see that compiling Typescript is a reasonably large unknown (although I have used `pkg` before successfully). We can always fall back to a Node-only solution if we can't get that working?

jamieparkinson: @alexwlchan I would prefer to discuss here otherwise the benefits of the RFC are lost!

alexwlchan: Right, I get where you're coming from now.

You're assuming an approach where we supply env vars like `MY_SECRET = /aws/secretsmanager/secret-id` and they get magically transformed a la ECS, whereas I was starting from the more general point of "we need to get secrets into a Lambda without using vanilla env vars". The particular reason I don't want them in env vars is because:

* env vars end up in Terraform state
* env vars are readable with the read-only role and visible in the console; secrets aren't

I can see the consistency arguments, but I'm not wholly convinced. I think simplicity is also important, and I want to be careful about adding yet another layer to the fragmented layer cake of our Lambda deployment approach.

I was imagining something more like a snippet we'd copy/paste or *maybe* put in a library (but again, Lambda libraries have been somewhat inconsistently used and maintained) – for Python, it's a seven line snippet:

```python
import boto3
import functools

@functools.cache
def get_secret(*, secret_id):
    client = boto3.client("secretsmanager")
    return client.get_secret_value(SecretId=secret_id)["SecretValue"]
```

and anybody reading the Lambda can immediately see how it works, vs having to hunt around to find where a layer is defined that magically makes the secrets appear.

I'm not trying to pick sides or claim the superiority of my approach, but hopefully it illustrates my confusion – I think we're coming at this from very different angles, hence my initial comment. The one hard requirement is "don't put secrets in environment variables", and stuff like language agnosticism are more fungible.

jamieparkinson: Hmm, yeah, to be honest I think I had the appetite for building a more complete solution (ie this RFC) last week but now other more important stuff has come up I will add some helpers to https://www.npmjs.com/package/@weco/ts-aws as you say.

This RFC was explicitly intended to supersede that approach - I would like to return to it if we end up leaning more heavily on Lambdas in future.

PR #75, opened 2022-04-13
046 transitive Sierra hierarchies

paul-butcher:
This relates to https://github.com/wellcomecollection/platform/issues/5479
rendered: https://github.com/wellcomecollection/docs/blob/1e3fd3183e536fe27a6aee99317dba5170c93dec/rfcs/046-transitive-sierra-hierarchies/README.md

### DIFF
--- a/rfcs/046-transitive-sierra-hierarchies/README.md
+++ b/rfcs/046-transitive-sierra-hierarchies/README.md
@@ -52,7 +52,7 @@ ephemera, where there is currently no clickable journey from an individual objec
 
 ## Proposed Solution
 
-A new "path merger" stage, operating on works-merged (both read and write), triggered by the Router on encountering a 
+A new "path concatenator" stage, operating on works-merged (both read and write), triggered by the Router on encountering a 
 document with both a collectionPath and a sourceIdentifier with an identifierType of sierra-system-number.
 
 The new stage will:
--- a/rfcs/046-transitive-sierra-hierarchies/README.md
+++ b/rfcs/046-transitive-sierra-hierarchies/README.md
@@ -52,8 +52,8 @@ ephemera, where there is currently no clickable journey from an individual objec
 
 ## Proposed Solution
 
-A new stage, operating on works-merged (both read and write), triggered by the Router on encountering a document with 
-a collectionPath and a sourceIdentifier with an identifierType of sierra-system-number.
+A new "path merger" stage, operating on works-merged (both read and write), triggered by the Router on encountering a 
+document with both a collectionPath and a sourceIdentifier with an identifierType of sierra-system-number.
 
 The new stage will:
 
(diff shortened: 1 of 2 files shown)


rfcs/046-transitive-sierra-hierarchies/README.md
```diff
@@ -0,0 +1,79 @@
...
+The knowledge that any of these individual portraits is part of the Fallaize Collection can only be known once all three
+levels have been ingested by the pipeline.  Because of this, the completion of the hierarchy for these records
+needs to be deferred until then.
+
+This is currently only to serve the Fallaize Collection, but may be required by other items in the future, e.g. boxes of
+ephemera, where there is currently no clickable journey from an individual object to the series the box is part of.
+
+## Proposed Solution
+
+A new stage, operating on works-merged, triggered by the Router on encountering a document with a collectionPath and
```
jamieparkinson, line 53: Can you add some detail on how data in the above example becomes a `collectionPath`? I know this is existing logic but it's not immediately obvious how the path relates to the 773/774 links

rfcs/046-transitive-sierra-hierarchies/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+I have considered trying to squeeze this into the behaviour of other stages.  The two most obvious places are 
+the transformer, and the relation embedder.
+
+It is inappropriate to place this in the transformer, because transformers operate on a single input document to 
+produce a single output document.  This behaviour requires multiple inputs, and can have multiple outputs.
+
+It is inappropriate to place this in the relation embedder, because it would add inappropriate complexity to an otherwise 
+stable application that is pretty good at turning a full path into a hierarchy. We would need to add behaviour to 
+match the right partial paths and to sum up depth values.
```
jamieparkinson, line 79: Hijacking your RFC to further the graph agenda: if anything I reckon we'll want to move in the direction of removing the relation embedder (it's too dependent on a certain kind of identifier/path) and making a "dumb" service that just builds graphs, transitive relations and all.

rfcs/046-transitive-sierra-hierarchies/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+This is currently only to serve the Fallaize Collection, but may be required by other items in the future, e.g. boxes of
+ephemera, where there is currently no clickable journey from an individual object to the series the box is part of.
+
+## Proposed Solution
+
+A new stage, operating on works-merged, triggered by the Router on encountering a document with a collectionPath and
+a sourceIdentifier with an identifierType of sierra-system-number.
+
+The new stage will:
```
jamieparkinson, line 56: What's the new stage called?

rfcs/046-transitive-sierra-hierarchies/README.md
paul-butcher: I agree, and that matches my thoughts on allowing the pipeline stages to run in a more DAG-like and parallel fashion.

We can have them all prepare the information required to create the final document, but store it elsewhere, then a final stage can put it all together.  This would be (IMO) a bit easier to maintain and reason about, and might also give us better control over partial runs.

However, for now, I want to do it this way to make better use of what is already in place.

jtweed (approved):

PR #76, opened 2022-04-14
RFC 047: Changing the structure of the Catalogue API index

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/5487; fleshing out https://github.com/wellcomecollection/platform/issues/5449

### DIFF
--- a/rfcs/047-catalogue-api-index-structure.md
+++ b/rfcs/047-catalogue-api-index-structure.md
@@ -20,7 +20,7 @@ We restructure documents in the API index to have three top-level fields:
 -   The `display` field contains the complete display document as a block of JSON.
     The API will return the contents of this field in public responses.
 
-    This field is mapped in the Elasticsearch index as an object field with [enabled=true](https://www.elastic.co/guide/en/elasticsearch/reference/current/enabled.html), meaning Elasticsearch will ignore it for indexing.
+    This field is mapped in the Elasticsearch index as an object field with [enabled=false](https://www.elastic.co/guide/en/elasticsearch/reference/current/enabled.html), meaning Elasticsearch will ignore it for indexing.
 
 -   The `query` field contains the values that we're indexing, e.g. work title.
     This will contain a subset of the work/image data that is indexed and analysed by Elasticsearch.
--- a/rfcs/047-catalogue-api-index-structure.md
+++ b/rfcs/047-catalogue-api-index-structure.md
@@ -15,23 +15,33 @@ This RFC proposes a new structure for the Catalogue API index which should remov
 
 ## Proposal
 
-We add a new field `display` to documents in the works and images index.
+We restructure documents in the API index to have three top-level fields:
 
-*   If the work/image is visible, it contains the complete display document as a block of JSON.
-*   If the work/image is not visible (e.g. redirected, deleted), it contains a `null` value.
+-   The `display` field contains the complete display document as a block of JSON.
+    The API will return the contents of this field in public responses.
 
-This field is mapped in the Elasticsearch index as an object field with [enabled=true](https://www.elastic.co/guide/en/elasticsearch/reference/current/enabled.html), meaning Elasticsearch will ignore it for indexing.
+    This field is mapped in the Elasticsearch index as an object field with [enabled=true](https://www.elastic.co/guide/en/elasticsearch/reference/current/enabled.html), meaning Elasticsearch will ignore it for indexing.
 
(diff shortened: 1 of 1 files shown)


rfcs/047-catalogue-api-index-structure.md
```diff
@@ -0,0 +1,57 @@
...
+This RFC proposes a new structure for the Catalogue API index which should remove this coupling.
+
+## Proposal
+
+We add a new field `display` to documents in the works and images index.
+
+*   If the work/image is visible, it contains the complete display document as a block of JSON.
+*   If the work/image is not visible (e.g. redirected, deleted), it contains a `null` value.
+
+This field is mapped in the Elasticsearch index as an object field with [enabled=true](https://www.elastic.co/guide/en/elasticsearch/reference/current/enabled.html), meaning Elasticsearch will ignore it for indexing.
```
jamieparkinson, line 23: I think you mean `enabled=false`?

jamieparkinson (commented):
I'm tempted to say we should go further, even as a first pass, and completely separate fields that are display-only and query-only

jamieparkinson:
I think in terms of deployable steps (ie PRs) it's right to do it in bits, but I think it's within the scope of the RFC to do all of those bits as one piece of work!

alexwlchan:
Ah, gotcha. I think we agree, but I’ve worded it badly – I’ll rephrase next week.

rfcs/047-catalogue-api-index-structure.md
alexwlchan: Oops, fixed.

alexwlchan:
I've updated the RFC to suggest a three-part index structure: `display`, `query` and `debug`.

Tagging in @harrisonpim for info/thoughts – I think we could flatten out the `query` field into a much simpler structure, e.g. rather we could smush all the bits of identifiers we actually query on into one field, and throw away the rest,

```json
"query": {
  "identifiers": [
    "xyz123ab",
    "b12345678",
    "V0001234"
  ]
  ...
}

jamieparkinson (approved):

paul-butcher:
> I've updated the RFC to suggest a three-part index structure: `display`, `query` and `debug`.
> 
> Tagging in @harrisonpim for info/thoughts – I think we could flatten out the `query` field into a much simpler structure, e.g. rather we could smush all the bits of identifiers we actually query on into one field, and throw away the rest,
> 
> ```json
> "query": {
>   "identifiers": [
>     "xyz123ab",
>     "b12345678",
>     "V0001234"
>   ]
>   ...
> }
> ```

Yes.  Make the data match the expected queries.  We may want to do a bit of requirements analysis before we go too far down that route - e.g. will an API client want to fetch something like "all records about bananas that have a Sierra identifier but not a CALM identifier"?  Even if we do, that could be better served by a different structure (e.g. `"in_source":["Sierra", "CALM"]), so shouldn't stop us doing this flattening, but it may inform exactly how we do it.

paul-butcher:
The API currently filters out certain fields, or leaves them in depending on query string parameters.

This implies that we still need to load the document in a modifiable manner (rather than just opaquely passing the display field on to the client), and remove the unwanted fields (which may not be at the top level).

Is this going to be tricky in the API in its current form (with Circe)?

alexwlchan:
> We may want to do a bit of requirements analysis before we go too far down that route

We're already doing stuff like this in the current API; I think we have an Elasticsearch `copyTo` in the index mapping that does something equivalent to this identifiers list.

> Is [filtering out fields] going to be tricky in the API in its current form (with Circe)?

I think circe-optics should be good for this; it's similar to what we're doing in the ID minter.

harrisonpim:
>Make the data match the expected queries

💯 

>requirements analysis

Yeah, we already do some of this with `copyTo`s, as @alexwlchan mentioned. The `search.titlesAndContributors` and `search.identifiers` fields are good examples.

Broadly, I think having a clearly differentiated set of `query` fields is a great move - we had (tentatively) started moving in that direction with the `search.something` fields, but were still using the regular fields alongside them. 

Counterintuitively, I have a feeling that if we continue down this path where we're focused on the requirements of the query without worrying about duplicating what's there for display, we might actually end up indexing _less_ than we do now 🤞

harrisonpim (approved):

paul-butcher (approved):

alexwlchan:
LET’S DO THIS. :D

PR #80, opened 2022-05-13
RFC 049: Changing how aggregations are retrieved by the Catalogue API

alexwlchan:
Rendered version: https://github.com/wellcomecollection/docs/blob/aggregations-modelling/rfcs/049-catalogue-api-aggregations-modelling/README.md

This is part of https://github.com/wellcomecollection/platform/issues/5530. I posted it as a Slack thread late last night, but I realised it's big and complicated enough to be worth explaining and discussing in a proper RFC, and not a half-baked Slack post.

### DIFF
--- a/rfcs/049-catalogue-api-aggregations-modelling.md
+++ b/rfcs/049-catalogue-api-aggregations-modelling.md
@@ -171,3 +171,5 @@ This is easiest to understand with an example:
 The ingestors would populate these `aggregatableValues` fields when it indexed a work.
 
 The API would aggregate over these fields specifically, and copy the values into the `data` field of our aggregation buckets.
+
+This would allow us to reduce the amount of model logic in the API, and would ensure a consistent rendering of values in aggregations and works.
--- a/rfcs/049-catalogue-api-aggregations-modelling.md
+++ b/rfcs/049-catalogue-api-aggregations-modelling.md
@@ -1,0 +1,173 @@
+# RFC 049: Changing how aggregations are retrieved by the Catalogue API
+
+For [RFC 047], we changed the catalogue API to serialise public API responses from an opaque `display` field in the Elasticsearch documents.
+
+Previously we were storing the pipeline's internal model in Elasticsearch.
+The API would retrieve the internal model, parse it, and convert it into the display model -- all as part of serving the request.
+Now, the pipeline creates the display model and stores it in a new `display` field in Elasticsearch.
+The API retrieves the `display` field, and treats it as opaque JSON -- without any knowledge of its structure.
+
+We'd hoped this would allow us to remove the internal/display models from the API repos, but the current implementation of aggregations make this tricky.
+
+This RFC proposes a change to how aggregations are handled that should remove this obstacle.
+
+[RFC 047]: ./047-catalogue-api-index-structure.md
+
+
+
+## Context: the desired behaviour
+
+Clients of the catalogue API can request aggregations of certain values (e.g. languages, licenses, work types).
+
+These are presented as a list of `AggregationBucket` in the API responses, i.e.:
+
+```
+AggregationBucket[T] {
+  data: T
+  count: Int
+  type: String
+}
+```
+
+where `T` is the same as the value appearing in the Work model.
+
+For example, languages on a work appear as follows:
+
+```
+{
+  "languages": [
+    {
+      "id": "eng",
+      "label": "English",
+      "type": "Language"
+    },
+    {
+      "id": "fre",
+      "label": "French",
+      "type": "Language"
+    }
+  ],
+  ...
+}
+```
+
+and we see that `Language` type in the aggregation buckets:
+
+```
+"aggregations": {
+  "languages": {
+    "buckets": [
+      {
+        "data": {
+          "id": "eng",
+          "label": "English",
+          "type": "Language"
+        },
+        "count": 691840,
+        "type": "AggregationBucket"
+      },
+      {
+        "data": {
+          "id": "fre",
+          "label": "French",
+          "type": "Language"
+        },
+        "count": 67187,
+        "type": "AggregationBucket"
+      },
+...
+```
+
+This presents a clean, consistent interface to clients – a value looks the same whether it's in a work or in an aggregation.
+
+We don't to change this behaviour.
+
+
+
+## The current implementation, and the problems it poses
+
+The API uses [Elasticsearch aggregations][es_aggs] to aggregate over a field in internal model.
+Elasticsearch will return single string values, which the API then serialises into the display model.
+
+For example, our language aggregation starts as [an ES terms aggregation][terms_agg] over the `data.languages.id` field:
+
+```http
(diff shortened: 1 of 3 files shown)


rfcs/049-catalogue-api-aggregations-modelling.md
```diff
@@ -0,0 +1,176 @@
...
+        },
+        "count": 67187,
+        "type": "AggregationBucket"
+      },
+...
+```
+
+This presents a clean, consistent interface to clients – a value looks the same whether it's in a work or in an aggregation.
+
+We don't to change this behaviour.
```
jamieparkinson, line 83:
```suggestion
We don't want to change this behaviour.
```

rfcs/049-catalogue-api-aggregations-modelling.md
```diff
@@ -0,0 +1,176 @@
...
+
+```
+type Query {
+  aggregatableValues: Map[String, List[String]]
+  ...
+}
+```
+
+The labels are the aggregation types (languages, work types, licenses) and the values are blobs of display JSON stored as strings.
+That is, we create the display JSON, then serialise the whole object as a string, and store that.
```
jamieparkinson, line 148: Github can't handle all the backticks if I make this a suggestion, but this was confusing to read as-is, how about...

```
We add a new field `query.aggregatableValues` to the Elasticsearch model:

type Query {
  aggregatableValues: Map[String, List[String]]
  ...
}

where the keys are the aggregation types (languages, work types, licenses) and the values are lists of display JSON serialised as strings.
```

jamieparkinson (approved):
I think this will probably work well - I do have some vague memories of it being discussed before, but I'm not sure that there was a good reason we never did it. I imagine it will improve aggregation performance (a little) as well.

Note: have we intentionally stopped putting RFCs in directories? Should I close https://github.com/wellcomecollection/docs/pull/78?

rfcs/049-catalogue-api-aggregations-modelling.md
```diff
@@ -0,0 +1,176 @@
...
+It would be nice if we could remove this coupling and simplify how aggregations work in the API.
+
+[es_aggs]: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html
+[terms_agg]: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html
+
+
+
+## Proposed solution
+
+We add a new field `query.aggregatableValues` to the Elasticsearch model.
```
jamieparkinson, line 137: "Elasticsearch model" is a bit unclear - how about "to the documents that we store in Elasticsearch"?

rfcs/049-catalogue-api-aggregations-modelling.md
```diff
@@ -0,0 +1,176 @@
...
+        },
+        {
+          "key" : "fre",
+          "doc_count" : 67187
+        },
+        ...
+```
+
+The API has to contain enough internal/display model logic to interpret these values -- to know that, say, `eng` means English and it's serialised as id/label/type.
+This is precisely the sort of model coupling we're trying to get away from.
```
paul-butcher, line 124: For a public standard (like languages), I don't see this problem as being one of model coupling in the way formats are.  Having code to translate from `ger` to "German" on both sides makes some sense to me.  However, as a general purpose rule (e.g. complex types, below), and also for efficiency/simplicity at the front end, it is a good idea to return ids and labels together.

alexwlchan:
> have we intentionally stopped putting RFCs in directories?

Err, no, I wasn't really paying attention when I did that. I'll fix that.

rfcs/049-catalogue-api-aggregations-modelling.md
```diff
@@ -0,0 +1,176 @@
...
+    }
+  },
+  ...
+}
+```
+
+The ingestors would populate these `aggregatableValues` fields when it indexed a work.
+This would be mapped as a `keyword` field in Elasticsearch.
+
+The API would aggregate over these fields specifically, and copy the values into the `data` field of our aggregation buckets.
```
paul-butcher: So in the API, we end up with code that transforms:
```{
           "key" : "blob of JSON as a string",
           "doc_count" : 691840
         },```
into 
```
{
         "data": that same JSON as an object
         "count": 691840,
         "type": "AggregationBucket"
       },
```
essentially ending up with something like this in the API code (excuse the Pythoning)
```
return [ 
  {
     "data: "json.load(bucket["key"]),
     "count": bucket["doc_count"],
     "type": "AggregationBucket"
   }
  for bucket in es_buckets
]
```
which is pretty simple.

Comment (alexwlchan) [reply rfcs/049-catalogue-api-aggregations-modelling.md]: Yes, exactly. (I had something pretty close to that while testing locally.)

Would it be helpful to add something like your comment to the RFC?

File: rfcs/049-catalogue-api-aggregations-modelling.md [hunk 6]
```diff
@@ -0,0 +1,176 @@
...
+
+This is easiest to understand with an example:
+
+```json
+{
+  "id": "example-work",
+  "query": {
+    "aggregatableValues": {
+      "languages": [
+        " { \"id\" : \"eng\", \"label\": \"English\", \"type\": \"Language\" } ",
```
paul-butcher: Are there scenarios where the type value here might not be the same for all aggregations of these values?

alexwlchan: Yes, contributors can be Agent/Person/Organisation.

paul-butcher: e.g. people - might we create aggregations where we look for people in both subject and contributor fields, to create an aggregation called "people", but also one where we look in just the contributor field to find "contributors".

alexwlchan: I can't think of a scenario where we'd aggregate across fields like you're describing.

paul-butcher: The reason for the question is about whether there's any situation where having type in here would be problematic.  I doubt if it would be.

It looks like the type will always be just the type of each object within the aggregation, and not have any relation to the type of the aggregation being performed.

paul-butcher:
I think this would work well.  My only concern is that I don't like strings of JSON inside JSON, but it's not an uncommon thing to do.

I did wonder if nested aggregations would fit our use case, and it's not as neat:
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-nested-aggregation.html

It would require an update to mappings (to make languages "nested"), so I couldn't try it out with our data, but here is a StackOverflow post that shows the kind of query we'd have to write, and response we'd get:
https://stackoverflow.com/questions/65847850/elasticsearch-aggregation-return-entire-sub-object-not-just-the-key

As you see, the query has to essentially reconstruct the object, and the response would then contain a load of ES stuff that we'd need to traverse/discard, so it really doesn't fulfil the "simplify the API code" brief.

PR #81, opened 2022-05-26
RFC 050: initial ideas for the design of the concepts API

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5536

Rendered version: https://github.com/wellcomecollection/docs/blob/050-concepts-api/rfcs/050-concepts-api/README.md

This follows the discussion at planning this morning. I've tried to capture everything we said so we can continue that discussion here. The sense I got from this morning is that we have consensus about a set of "core" fields (id, label, identifiers), and then we need to think more about everything else.

### DIFF
--- a/rfcs/050-concepts-api/README.md
+++ b/rfcs/050-concepts-api/README.md
@@ -1,0 +1,176 @@
+# RFC 050: Design considerations for the concepts API
+
+This RFC collects some initial thinking on how we might represent concepts in the catalogue API.
+It's a starting point for discussions; not a final design.
+
+
+
+## Goals
+
+*   We want to be able to create per-person and per-subject pages on the website.
+    The data on these pages should come from the new concepts API.
+
+*   These pages should contain enough context to distinguish two people/subjects (e.g. whether "Charles Darwin" is the famous naturalist, his uncle, his son, or his grandson).
+
+*   We're not trying to replicate Wikidata -- the APIs shouldn't return large amounts of information we aren't going to use on wellcomecollection.org.
+
+*   Any new APIs should be similar to the existing works and images APIs.
+    We'll use the same conventions for filters, aggregations, including optional fields, and so on.
+
+
+
+## Endpoint #1: retrieve a single concept
+
+This is the single concept endpoint, which is used to render a page for an individual subject or person:
+
+```
+GET /concepts/azxzhnuh
+{
+  "id": "azxzhnuh",
+  "identifiers": [
+    {
+      "identifierType": "lc-names",
+      "value": "n12345678",
+      "type": "Identifier"
+    },
+    ...
+  ],
+
+  "label": "Florence Nightingale",
+  "alternativeLabels": [
+    "The Lady with the Lamp",
+  ]
+
+  "type": "Person|Subject|Organisation|Place"
+
+  // Everything above here is stuff I'm pretty sure we'll need;
+  // everything below it is more nebulous and more likely to change
+  // in the final API.
+
+  "description": "Florence Nightingale was an English social reformer, statistician and the founder of modern nursing. Nightingale came to prominence while serving as a manager and trainer of nurses during the Crimean War, in which she organised care for wounded soldiers at Constantinople.",
+
+  // not locations
+  "urls": [
+    "https://en.wikipedia.org/wiki/Florence_Nightingale"
+  ],
+
+  // cf productionEvent?
+  "dates": [ { date, meaning } ]
+  "places": [ { place, meaning } ]
+  "birthDate/place"
+  "deathDate/place"
+
+  "thumbnail": { … },
+
+  "relatedConcepts": [
+    {
+      "id": "asoiham1",
+      "label": "Crimea",
+      "type": "Place"
+    }
+  ]
+}
+```
+
+Notes:
+
+-   The `id` and `identifiers` fields are the same as on existing works/images.
+
+-   The `label` and `alternativeLabels` fields are named analogously to `title` and `alternativeTitles` on works.
+
+-   The `type` will be an enumeration that includes Person, Organisation, Meeting, Subject.
+    For the initial implementation we'll likely have a few subtypes of Person but not of Subject.
+
+    The alternative is to have `type: Concept` and do subtypes through something similar to workType/format, but I'm not sure that adds anything.
+
+    This is a value users would be able to filter on (if we provide an endpoint for listing/searching concepts).
+
+-   The `description` should provide enough context to help users distinguish subjects/people with the same label but different meaning.
+    This will likely be drawn from LCSH or Wikidata initially, but at some point we might write our own content here.
+
+    Q: Is `description` the best name for this field, or should it be called something different?
+
+-   The `urls` list points to this subject/person elsewhere.
+    We aren't trying to replicate LCSH/MESH/Wikidata; if somebody wants to read more in-depth about a given concept, we should link them to an external entry.
+
+    In works we use digital locations in `work.items.locations` to link to external resources.
+    We don't want to reuse that structure, which is more complicated than we need here.
+
+    Q: Aside from the URL, what other values do we want for these links?
+    e.g. link label
+
+    Q: Do we need these values at all?
+    If we put the Wikidata ID in the list of identifiers, do we leave clients to construct a URL, or do we provide it in the API?
+
+-   The `date`/`place` information is somewhat speculative.
+
+    We think we might include birth/death information on per-person pages; we could also link to significant dates and places and use that as a way to link between concepts.
+    This is similar to production events on works.
+
+    We can start with a `birthDate/birthPlace/deathDate/deathPlace` field on the initial version of the API, and decide if we want to put this information on per-person pages.
+    If yes, we can think about how to model this properly.
+    If no, we can scrap all of these fields.
+
+-   The `relatedConcepts` field lets us link between concept pages.
+    It contains a simplified subset of a Concept's fields; enough to render a link but nothing more.
+
+    We don't want to describe how two concepts are related; this is significantly more complicated.
+
+    Q: Is relatedness symmetric?
+    If `A.relatedConcepts = [B]`, is it true that `B.relatedConcepts = [A]`?
+    This might affect if/how we build an API for listing concepts.
+
+
+
+## Endpoint #2: Concepts in the works API
+
+We'll simplify how concepts appear in the works API.
+We'll provide enough information to render a link to the per-person or per-subject page, but nothing more -- for detailed information, use the concepts API.
+
+For example:
+
+```
+GET /works?contributors.agent=azxzhnuh
+{
+  …,
+  "contributors": [
+    {
+      "agent": {
+        "id": "azxzhnuh",
+        "label": "Florence Nightingale",
+        "type": "Person"
+      },
+      "roles": [
+        {
+          "label": "Author",
+          "type": "ContributionRole"
+        }
+        ...
+      ]
+    }
+  ]
+}
+```
+
+
+
+## Endpoint #3: Listing and searching concepts
+
+If we decide we need this, it would look similar to the works and images endpoints:
+
+```
+GET /concepts
+{
+  "type": "ResultList",
+  "results": [...]
+}
+```
+
+The `results` list would contain the same objects as returned from endpoint #1.
+
+Filtering, querying, aggregations, and so on, would follow the same patterns as the works and images APIs.
+We don't need to decide the details of this yet.
+
+Q: Do we need this endpoint?
+Is listing/searching concepts something we think we'll do?
+Even if we won't use it, is it worth providing for parity with the works/images APIs?
--- a/rfcs/050-concepts-api/README.md
+++ b/rfcs/050-concepts-api/README.md
@@ -85,6 +85,9 @@ Notes:
 
     This is a value users would be able to filter on (if we provide an endpoint for listing/searching concepts).
 
+    Although conceptually we know that Person/Organisation/Meeting are subtypes of Agent, the API doesn't.
+    For example, if you wanted to present a list of agent-like things, you'd use a filter `type=Person,Organisation,Meeting` rather than `type=Agent`.
+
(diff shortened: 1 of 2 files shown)


rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,194 @@
...
+  "type": "Person|Subject|Organisation|Place"
+
+  // Everything above here is stuff I'm pretty sure we'll need;
+  // everything below it is more nebulous and more likely to change
+  // in the final API.
+
+  "description": "Florence Nightingale was an English social reformer, statistician and the founder of modern nursing. Nightingale came to prominence while serving as a manager and trainer of nurses during the Crimean War, in which she organised care for wounded soldiers at Constantinople.",
+
+  // not locations
+  "urls": [
```
jamieparkinson, line 53: Are Uniform Resource Locators really not locations?

Can we find a way of describing what these are _in relation to the concept_? Florence Nightingale didn't have any URLs (she had a photographic memory for IP addresses).

I think these might want to be enriched with some description of what they are - if I want to render a works page with a link to Wikipedia, I'll want to know which of the `urls` is the wikipedia link. 

Starting point (not a final suggestion) to address my points above; inspiration taken from Wikidata:
```json
{
  "describedAt": [
    {
      "source": "wikipedia"
      "url": "https://en.wikipedia.org/wiki/Florence_Nightingale"
    }
  ]
}
```

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,194 @@
...
+    "https://en.wikipedia.org/wiki/Florence_Nightingale"
+  ],
+
+  // cf productionEvent?
+  "dates": [ { date, meaning } ]
+  "places": [ { place, meaning } ]
+  "birthDate/place"
+  "deathDate/place"
+
+  "thumbnail": { … },
```
jamieparkinson, line 63: I reckon we'll want to do away with this and consider a more holistic integration with (our) `images` entities

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,194 @@
...
+
+  "relatedConcepts": [
+    {
+      "id": "asoiham1",
+      "label": "Crimea",
+      "type": "Place"
+    }
+  ]
```
jamieparkinson, line 71: Can we take this out of scope for now? Talking about representing this in the API (let alone how to architect it internally) could occupy many hectares of bikesheds

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,194 @@
...
+    If we put the Wikidata ID in the list of identifiers, do we leave clients to construct a URL, or do we provide it in the API?
+
+-   The `date`/`place` information is somewhat speculative.
+
+    We think we might include birth/death information on per-person pages; we could also link to significant dates and places and use that as a way to link between concepts.
+    This is similar to production events on works.
+
+    We can start with a `birthDate/birthPlace/deathDate/deathPlace` field on the initial version of the API, and decide if we want to put this information on per-person pages.
+    If yes, we can think about how to model this properly.
+    If no, we can scrap all of these fields.
```
jamieparkinson, line 112: Agreed

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,194 @@
...
+
+-   The `relatedConcepts` field lets us link between concept pages.
+    It contains a simplified subset of a Concept's fields; enough to render a link but nothing more.
+
+    We don't want to describe how two concepts are related; this is significantly more complicated.
+
+    Q: Is relatedness symmetric?
```
jamieparkinson, line 119: Good question, I would question the naming `related`

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,194 @@
...
+GET /works?contributors.agent=azxzhnuh
+{
+  …,
+  "contributors": [
+    {
+      "agent": {
+        "id": "azxzhnuh",
+        "label": "Florence Nightingale",
+        "type": "Person"
+      },
```
jamieparkinson, line 142: I think this implies that the works API knows that `Person`s and `Organisation`s (for example) are `agent`s, even when the concepts API doesn't. I think that's a good thing.

jamieparkinson (commented):
You're going to need [one of these](https://wellcomecollection.org/works/gsb72qjz/images?id=kxfxvhjy) for my review

rfcs/050-concepts-api/README.md
alexwlchan: Agreed, and this has a subtle effect on how we might filter for agents – I've added that to the notes on `type`.

rfcs/050-concepts-api/README.md
alexwlchan: I've changed it to `connected` for now; I don't think that's the final term, but it's enough for the discussion I want to have here.

rfcs/050-concepts-api/README.md
alexwlchan: I've clarified it slightly – I've used a deliberately worse name (`connectedConcepts`), because I agree we don't want to be deciding the exact representation yet – but I do think we should decide whether you can get connections through this endpoint, or if they're available from a separate endpoint.

I've added something to the explanation of this field:

> Q: Do we want the connections between concepts to be discoverable from a single concept endpoint, or do we want that elsewhere?
> I don't think we need to decide exactly how we'll model it (yet), but we should decide if it lives here or in at a distinct endpoint.

This feels like something we should decide as part of the list of the new endpoints.

rfcs/050-concepts-api/README.md
alexwlchan: Do you mean on works as well as concepts?

rfcs/050-concepts-api/README.md
alexwlchan: But I do think there's something to the idea of `describedAt` as a way to say "learn more about this concept over here".

rfcs/050-concepts-api/README.md
jamieparkinson: Interesting one - quite possibly for v3! I remember this being talked about when `images` was first implemented/designed. @jtweed might have opinions.

rfcs/050-concepts-api/README.md
jamieparkinson: Yeah, leaning towards `DigitalLocation`s being OK - if we take one thing from this comment then it's that I don't really like `urls` as a name and I do rather like `describedAt` - cf https://www.wikidata.org/wiki/Property:P973

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,200 @@
...
+    },
+    ...
+  ],
+
+  "label": "Florence Nightingale",
+  "alternativeLabels": [
+    "The Lady with the Lamp",
+  ]
+
+  "type": "Person|Subject|Organisation|Place"
```
jtweed, line 44: 100% think that `Subject` is the right type here, but flagging that we already have a `Subject` type that is used to hold a list of `Concept` and related them to a `Work`. We need to work out what we're going to do about that...

rfcs/050-concepts-api/README.md
jtweed: The current designs are shying away from a thumbnail for concepts, though I can potentially see it coming back for creating a richer snippet to link to a concept page.

In that case, thumbnail is correct, like on works. Concepts will also separately have images, which are the images with that concept. If that makes sense.

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+  // cf productionEvent?
+  "dates": [ { date, meaning } ]
+  "places": [ { place, meaning } ]
+  "birthDate/place"
+  "deathDate/place"
+
+  "thumbnail": { … },
+
+  "connectedConcepts": [
```
jtweed, line 65: I can see us breaking this out a bit into multiple kinds of relationship. However, I don't see concept relationships happening for any early version so I wouldn't worry too much for now.

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+-   The `type` will be an enumeration that includes Person, Organisation, Meeting, Subject.
+    For the initial implementation we'll likely have a few subtypes of Person but not of Subject.
+
+    The alternative is to have `type: Concept` and do subtypes through something similar to workType/format, but I'm not sure that adds anything.
+
+    This is a value users would be able to filter on (if we provide an endpoint for listing/searching concepts).
+
+    Although conceptually we know that Person/Organisation/Meeting are subtypes of Agent, the API doesn't.
+    For example, if you wanted to present a list of agent-like things, you'd use a filter `type=Person,Organisation,Meeting` rather than `type=Agent`.
```
jtweed, line 89: I think that fine and arguably preferable. We're really saying agent is the relationship from `Contributor` to `Person` etc. I don't think the complexity added by making the type hierarchy explicit is really necessary.

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+-   The `description` should provide enough context to help users distinguish subjects/people with the same label but different meaning.
+    This will likely be drawn from LCSH or Wikidata initially, but at some point we might write our own content here.
+
+    Q: Is `description` the best name for this field, or should it be called something different?
```
jtweed, line 94: I think it's fine, it matches Wikidata and we don't intend to write lots of long descriptions. That would be the only potential sticking point I see though, what would we call that if we did so?

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,200 @@
...
+    This is similar to production events on works.
+
+    We can start with a `birthDate/birthPlace/deathDate/deathPlace` field on the initial version of the API, and decide if we want to put this information on per-person pages.
+    If yes, we can think about how to model this properly.
+    If no, we can scrap all of these fields.
+
+-   The `connectedConcepts` field lets us link between concept pages.
+    It contains a simplified subset of a Concept's fields; enough to render a link but nothing more.
+
+    Q: Do we want the connections between concepts to be discoverable from a single concept endpoint, or do we want that elsewhere?
```
jtweed, line 120: I think we want as part of the API, but could be an include.

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,200 @@
...
+    I don't think we need to decide exactly how we'll model it (yet), but we should decide if it lives here or in at a distinct endpoint.
+
+    We don't want to describe how two concepts are connected; this is significantly more complicated.
+
+    Q: Is connectedness/relatedness symmetric?
```
jtweed, line 125: No, it's not. No guarantees on what one concept links to for its most relevant related concepts. If we were rendering out the entire graph maybe, but not in practice. Eg a big generic subject like "London" wouldn't link to everything related to it, but many of those related things might link to "London".

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,200 @@
...
+    If `A.connectedConcepts = [B]`, is it true that `B.connectedConcepts = [A]`?
+    This might affect if/how we build an API for listing concepts.
+
+
+
+## Endpoint #2: Concepts in the works API
```
jtweed, line 131: As above, the really important example here is `subjects` and probably corresponding changes to `genres`. `contributors` I think doesn't really need to change.

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,200 @@
...
+}
+```
+
+The `results` list would contain the same objects as returned from endpoint #1.
+
+Filtering, querying, aggregations, and so on, would follow the same patterns as the works and images APIs.
+We don't need to decide the details of this yet.
+
+Q: Do we need this endpoint?
+Is listing/searching concepts something we think we'll do?
```
jtweed, line 181: Yes, though maybe not initially. We will likely ultimately end up using it for both search and building out a browse interface to subjects and people.

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,200 @@
...
+Even if we won't use it, is it worth providing for parity with the works/images APIs?
```
jtweed, line 182: In any case, from an "API as a product" point of view I strongly think the answer to this is yes we should add it to allow other people to do things we aren't doing.

rfcs/050-concepts-api/README.md
alexwlchan: I used a Wikipedia URL as an example; I went for the slightly more generic `urls` because I wasn't sure if we'd ever want other links here (e.g. will every concept have a Wikipedia entry, or might we link elsewhere for some things?)

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+This is the single concept endpoint, which is used to render a page for an individual subject or person:
+
+```
+GET /concepts/azxzhnuh
+{
+  "id": "azxzhnuh",
+  "identifiers": [
+    {
+      "identifierType": "lc-names",
```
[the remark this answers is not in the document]
alexwlchan: I don't want to touch that change as part of this RFC, but I've added it to the list of Catalogue v3 ideas: https://github.com/wellcomecollection/platform/issues/5246#issuecomment-1141874168

rfcs/050-concepts-api/README.md
```diff
@@ -0,0 +1,206 @@
...
+      "value": "n12345678",
+      "type": "Identifier"
+    },
```
jamieparkinson, line 35: Just noticed this isn't actually the `identifier` we use elsewhere: that looks like
```json
{
  "identifierType": {
    "id": "sierra-system-number",
    "label": "Sierra system number",
    "type": "IdentifierType"
  },
  "value": "b24367588",
  "type": "Identifier"
}
```

alexwlchan:
I'm going to merge this as-is; it represents our thinking at a point of time, and is a jumping off point for more discussion.

PR #82, opened 2022-06-16
RFC 053: Filtering by contributor, genre, and subject

alexwlchan:
(Leaving 051 because I saw Jamie take it yesterday, 052 for the concepts pipeline)

Rendered version: https://github.com/wellcomecollection/docs/blob/052-filtering-by-subject/rfcs/052-filtering-by-subject/README.md

### DIFF
--- a/rfcs/053-filtering-by-subject/README.md
+++ b/rfcs/053-filtering-by-subject/README.md
@@ -15,11 +15,14 @@ Strictly speaking we only need subjects and contributors for this work, but genr
 
 1.  On per-concept pages, there's a sample of matching images/works.
 
-2.  On per-concept pages, there's a link to a filtered search for the given concept, directly below the sample results.
+2.  On per-concept pages, there's a link to a filtered search for the identified concept, directly below the sample results.
 
-3.  On work pages, the list of subjects/contributors/genres to link to filtered searches for each subject.
+3.  On work pages, the list of subjects/contributors/genres link to:
 
-4.  In the works API, there are filters and aggregations for subject/contributor/genre.
+    -   a concept page if the subject/contributor is identified (new behaviour)
+    -   a filtered search by label if the subject/contributor/genre is unidentified (existing behaviour)
+
+4.  In the works API, there are filters and aggregations for subject/contributor/genre that support (3).
 
 5.  Given a single work in the works API, there should be an obvious way to construct a filter URL for works with the same subjects/contributors/genres as this work.
 
@@ -39,16 +42,33 @@ If we add filtering/aggregations for subjects by ID, we already know how they'll
 This is consistent with our existing API design.
 
 But do we add filtering/aggregations for subjects by ID?
-We find ourselves in a dilemma:
+How do we handle this in the front-end?
 
-*   We can't rely on filtering by label, because there may be concepts with similar/identical labels but which refer to different things.
+Consider the following flows in the front-end:
+
+1.  A user lands on the concept page for "mental health".
+
+    This includes a list of works with that identified concept.
+    When they click to see the full list of works, they should see filtered search results.
+
+    This filtered search must use ID filtering, because there may be concepts with similar/identical labels but which refer to different things.
     e.g. two members of the same family.
 
-*   We can't rely on filtering by ID, because not all subjects/genres/contributors are identified.
+    Q: How do we distinguish this in the UI from a label search for "mental health"?
 
-From an API perspective, we could easily support both, but it's more complicated in the front-end.
-How do we choose which filter mechanism to use?
-How does an API client choose which filter mechanism to use?
+    Q: Should a user be able to discover this filter through the search UI?
+    If they remove the filter, can they re-add it without going via the concepts page?
+
+2.  A user is on a search page.
+    They want to filter by subject.
+    They click the dropdown to see a list of available subject filters, and pick one.
+
+    Q: Is the list of available subjects based on ID or label?
+    Is it a mixture of both?
+
+    Q: How do we distinguish between a filter for the label "mental health" and the identified concept?
+
+    Q: How do we distinguish between two identified concepts with the same label?
 
 Questions:
 
@@ -58,42 +78,6 @@ Questions:
 *   How many identified/unidentified concepts are there in the catalogue?
 
 *   How do we want to approach this filtering?
-
-## An idea
-
-We create synthetic source identifiers for unidentified subjects, which uses the label as the value, e.g.
-
-```json
-{
-  "identifierType": {
-    "id": "wellcome-catalogue-label",
-    "label": "Wellcome catalogue label",
-    "type": "IdentifierType"
-  },
-  "value": "Mental health",
-  "type": "Identifier"
-}
-```
-
-This would allow these subjects to get canonical IDs within the platform, but we don't want to put synthetic identifiers in public API responses.
-
-Instead, we suppress them in the API response but keep the canonical ID, so you'd get subjects like:
-
-```json
-"subjects": [
-  {
-    "id": "axbum98n",
-    "identifiers": [],
-    "label": "Mental health",
-    "concepts": [],
-    "type": "Subject"
-  }
-]
-```
-
-This means every subject can be filtered by identifier, and API clients (including the front-end) can find the filter URL.
-
-It adds some complexity to the pipeline, because we have to ensure these synthetic identifiers get scrubbed.
 
 ## See also
 
--- a/rfcs/053-filtering-by-subject/README.md
+++ b/rfcs/053-filtering-by-subject/README.md
@@ -33,7 +33,7 @@ Strictly speaking we only need subjects and contributors for this work, but genr
 
 We have filters and aggregations for *label*, not ID.
 
-## Future behaviour
+## Considerations for future behaviour
 
 If we add filtering/aggregations for subjects by ID, we already know how they'll be named: `subjects` and `source.subjects` for works and images, respectively.
 This is consistent with our existing API design.
@@ -59,24 +59,41 @@ Questions:
 
 *   How do we want to approach this filtering?
 
-## Rejected approaches
+## An idea
 
-*   Only support filtering by ID, and mint our own identifiers for unidentified subjects.
-    Then all subjects are identified and we can filter by them.
+We create synthetic source identifiers for unidentified subjects, which uses the label as the value, e.g.
 
-    e.g. such subjects could get a source identifier
+```json
+{
+  "identifierType": {
+    "id": "wellcome-catalogue-label",
+    "label": "Wellcome catalogue label",
+    "type": "IdentifierType"
+  },
+  "value": "Mental health",
+  "type": "Identifier"
+}
+```
 
-    ```json
-    {
-      "identifierType": {
-        "id": "wellcome-catalogue-label",
-        "label": "Wellcome catalogue label",
-        "type": "IdentifierType"
-      },
-      "value": "Mental health",
-      "type": "Identifier"
-    }
-    ```
+This would allow these subjects to get canonical IDs within the platform, but we don't want to put synthetic identifiers in public API responses.
+
+Instead, we suppress them in the API response but keep the canonical ID, so you'd get subjects like:
+
+```json
+"subjects": [
+  {
+    "id": "axbum98n",
+    "identifiers": [],
+    "label": "Mental health",
+    "concepts": [],
+    "type": "Subject"
+  }
+]
+```
+
+This means every subject can be filtered by identifier, and API clients (including the front-end) can find the filter URL.
+
+It adds some complexity to the pipeline, because we have to ensure these synthetic identifiers get scrubbed.
 
 ## See also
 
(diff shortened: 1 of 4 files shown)


rfcs/053-filtering-by-subject/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+4.  In the works API, there are filters and aggregations for subject/contributor/genre.
+
+5.  Given a single work in the works API, there should be an obvious way to construct a filter URL for works with the same subjects/contributors/genres as this work.
+
+6.  The catalogues are the source of truth for subject identifiers.
+    We can find equivalent identifiers, but we can't pick them from scratch.
+
+    e.g. If a Sierra record has a subject tagged with an LCSH identifier, we can find the Wikidata subject with that identifier.
+    If a Sierra record has a subject with no identifier, we can't choose an identifier, even if we could find a Wikidata subject with a matching label.
```
alexwlchan, line 30: These requirements might seem oddly specific; I added them as a way to negate some of the more wild ideas I had. (e.g. have "shadow identifiers" for unidentified concepts, but then it's impossible for anyone to work out how to filter!)

They did lead me towards the suggestion in "An idea".

rfcs/053-filtering-by-subject/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+*   We can't rely on filtering by ID, because not all subjects/genres/contributors are identified.
+
+From an API perspective, we could easily support both, but it's more complicated in the front-end.
+How do we choose which filter mechanism to use?
+How does an API client choose which filter mechanism to use?
+
+Questions:
+
+*   Are the requirements as stated correct?
```
jamieparkinson, line 55: A general response to the problem stated: it seems to me that the shortcomings of each approach (ID vs label) are kind of related to each other. That is to say - I think the existence of unidentifiable concepts in some sense implies an inability to disambiguate concepts by label. 

In reality it's slightly more complex than that - as you say, with family names it's realistic to end up with multiple ambiguous _and_ identified concepts. Still, I'm not sure there's really a problem per se...

> 1. On per-concept pages, there's a sample of matching images/works.

Concept pages are for identified concepts, this is fine with the ID search

> 1. On per-concept pages, there's a link to a filtered search for the given concept, directly below the sample results.

I'm not sure what this requirement means - is a "search for the given concept" a works/image search? Again, I feel that it's fine for this only to be for identified concepts.

> 3. On work pages, the list of subjects/contributors/genres to link to filtered searches for each subject.

I wonder if an approach here would be to link to a label search for an unidentified concept, and an ID search for an identified concept. Perhaps there could be UI signifiers that we were doing a text search rather than searching for some kind of canonical entity.

> 4. In the works API, there are filters and aggregations for subject/contributor/genre.

Think this is implied by the above 3 requirements.

> 5. Given a single work in the works API, there should be an obvious way to construct a filter URL for works with the same subjects/contributors/genres as this work.

As with 3 but agreed it isn't necessarily obvious. 

> 6. The catalogues are the source of truth for subject identifiers.

For sure.

**Appendix**

I think I might have talked myself into (via responses to 3 and 5) some kind of concept filter that accepts IDs _and_ labels? Is that an unacceptable deviation? It would behave much like the existing works search when searching for an ID.

jamieparkinson: I suppose it would be better named a "concepts query" rather than a "concepts filter" in this case. I'm going off it already tbh - I'm not sure how the aggregations would work, although I think that would be the expected behaviour in this case.

Re search for unidentified concept labels returning identified concepts, I think that's fine.

rfcs/053-filtering-by-subject/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+Strictly speaking we only need subjects and contributors for this work, but genres are so similar we should treat them in the same way.
+
+## Requirements
+
+1.  On per-concept pages, there's a sample of matching images/works.
+
+2.  On per-concept pages, there's a link to a filtered search for the given concept, directly below the sample results.
+
+3.  On work pages, the list of subjects/contributors/genres to link to filtered searches for each subject.
```
jtweed, line 20: On works pages, the list of subjects/contributors/genres to link to concept pages.

rfcs/053-filtering-by-subject/README.md
jtweed: I agree that if something is unidentified, we shouldn't make up an id. It should remain unidentified and that may mean that we have to treat it differently, eg the difference between identified and unidentified series on works pages. But it is being true to the data. We are more likely to get into a pickle if we don't go with the grain of what we have.

rfcs/053-filtering-by-subject/README.md
```diff
@@ -0,0 +1,101 @@
...
+If we add filtering/aggregations for subjects by ID, we already know how they'll be named: `subjects` and `source.subjects` for works and images, respectively.
+This is consistent with our existing API design.
+
+But do we add filtering/aggregations for subjects by ID?
+We find ourselves in a dilemma:
+
+*   We can't rely on filtering by label, because there may be concepts with similar/identical labels but which refer to different things.
+    e.g. two members of the same family.
+
+*   We can't rely on filtering by ID, because not all subjects/genres/contributors are identified.
```
jtweed, line 47: We can filter by id when we know it is an identified subject, I think I'm with Jamie and the direction he's heading below. Concept pages are for identified things. If they aren't identified, they can't have a page.

rfcs/053-filtering-by-subject/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+From an API perspective, we could easily support both, but it's more complicated in the front-end.
+How do we choose which filter mechanism to use?
```
jtweed, line 50: By whether the subject is identified or not.

rfcs/053-filtering-by-subject/README.md
jtweed: I agree with the general gist of what Jamie is saying above. Concept pages are for identified concepts, if it's not identified we don't give it a page. We tend towards more identified concepts over time (and it is the vast majority already, genres aside). For genres that means, no concept pages but a continuation of what we have today.

The caveat to the above is the potential mixing of user flows. If all genres do one thing and all subjects do another, that's probably ok. If it's less deterministic than that we may have a problem.

I would like @GarethOrmerod to be involved in this conversation, as I think we probably do need to treat linking to concept pages and back to searches for labels differently when we link to them from works pages, to make sure it's clear to user where's they're going to go in each case.

rfcs/053-filtering-by-subject/README.md
```diff
@@ -0,0 +1,101 @@
...
+How does an API client choose which filter mechanism to use?
+
+Questions:
+
+*   Are the requirements as stated correct?
+    Are there any missing or unnecessary requirements?
+
+*   How many identified/unidentified concepts are there in the catalogue?
```
jtweed, line 58: @harrisonpim may also want to weigh in here.

rfcs/053-filtering-by-subject/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+*   How do we want to approach this filtering?
+
+## An idea
+
+We create synthetic source identifiers for unidentified subjects, which uses the label as the value, e.g.
```
jtweed, line 64: Not convinced, I'd rather be honest about what the data is or find a way to make everything genuinely identified.

If we have to make a page for every concept, then maybe. But let's explore not doing that first.

rfcs/053-filtering-by-subject/README.md
alexwlchan: Clarified.

rfcs/053-filtering-by-subject/README.md
```diff
@@ -0,0 +1,85 @@
...
+2.  A user is on a search page.
+    They want to filter by subject.
+    They click the dropdown to see a list of available subject filters, and pick one.
+
+    Q: Is the list of available subjects based on ID or label?
+    Is it a mixture of both?
+
+    Q: How do we distinguish between a filter for the label "mental health" and the identified concept?
+
+    Q: How do we distinguish between two identified concepts with the same label?
```
paul-butcher, line 71: Regarding Concept pages, We've mentioned things like Date of Birth and a brief description being used as a disambiguation hint.  Perhaps such a thing could be short enough to be parenthesised, or that we also ensure that our Concept labels are unique by that mechanism: e.g. Nightingale (Bird) vs Nightingale (Statistician).

Wikipedia URLs are by name, rather than opaque id, and this is how they have solved the problem.

If we are linking via ID, then Concept labels can be mutable.

rfcs/053-filtering-by-subject/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+    Q: Should a user be able to discover this filter through the search UI?
+    If they remove the filter, can they re-add it without going via the concepts page?
+
+2.  A user is on a search page.
+    They want to filter by subject.
+    They click the dropdown to see a list of available subject filters, and pick one.
+
+    Q: Is the list of available subjects based on ID or label?
+    Is it a mixture of both?
```
paul-butcher, line 67: We might want to do some analysis on the degree to which this is a problem.  If there are only a handful of unidentified subjects, then we could endeavour to get them identified and work only on IDs.

If, OTOH, there are a substantial number of distinct unidentified subjects, then we may have to mix them.

rfcs/053-filtering-by-subject/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+    Q: How do we distinguish between a filter for the label "mental health" and the identified concept?
```
paul-butcher, line 69: I suspect (as long as everything is identified correctly) that what someone would be asking for when filtering by the label is more like a `contains` query than an `equals` query.  i.e. they don't remember exactly what the label is, but it was something like X.

If we are in a situation where we have to mix label and id in a UI, because there are too many unidentifiable labels, then that can only be seamlessly concealed from the user as long as there are no clashes (e.g. there are no label: "Mental Health" records that do not refer to the identified concept "Mental Health")

rfcs/053-filtering-by-subject/README.md
harrisonpim: I think concepts' preferred labels have to be mutable; the initial motivation for all of this work was a desire to update outdated language, and that's never going to stop being a problem

paul-butcher: Yes.  What I mean by that is that we have an advantage over Wikipedia in that no one elsewhere will be linking to a page like wellcomecollection.org/Florence_Nightingale and getting cross because we change it to  wellcomecollection.org/Nightingale,_Florence.  

Their solution is good enough for them even with their disadvantages (The Georgia Problem), it would (IMO) work even better for us.

harrisonpim (commented):
In response to general discussion above, I'm going to go against the grain and say that I think we _should_ be minting IDs for unidentified concepts, just like identified ones.

I think the proposed inconsistency in user experience between identified and unidentified concepts sounds Bad, and I can't really see a proper justification for it. We're introducing a lot of tricky technical work for ourselves through that distinction, and I can't see who it's helping. I feel like the same amount of work could be better spent on the careful minting/merging/redirecting etc of concept IDs, while providing a more consistent experience for users.

Even if the majority of subjects are identified, we still have _a lot_ of unidentified things (subjects, people, genres) and I don't think users should have to worry about whether we've managed to enrich the thing that they've clicked on - they'll expect the experience to be the same. Again, coming up with ways of hinting to users that some concepts are identified and some aren't through design just sounds like extra work for us and our users, and I can't see who it helps.

It's also been said elsewhere that we shouldn't implicitly introduce new data to what's in the catalogue, and I agree - I just don't think we're introducing anything new by IDing a string which already exists.  
If anything, I think consistency provides a _more_ honest reflection of what's in the catalogue.  
In that sense, a bare concepts page also feels like a better nudge for folks in collections info to identify those weird, dangling concepts, compared to a page of search results. It seems easier to me to imagine enriching/disambiguating an existing, sparse concept page than the leap from filtered search results to full concepts page.

rfcs/053-filtering-by-subject/README.md
alexwlchan: I went and had a look in a recent snapshot.

<table>
<tr><th></th><th>unique values</th><th>total occurrences</th></tr>
<tr><td>unidentified subjects;</br>label only</td><td>120,301</td><td>556,331</td></tr>
<tr><td>identified subjects;</br>canonical ID</td><td>104,481</td><td>1,540,795</td></tr>
</table>

paul-butcher:
> In that sense, a bare concepts page also feels like a better nudge for folks in collections info to identify those weird, dangling concepts, compared to a page of search results.

Also, finding something attached to the "wrong" Concept would be a better nudge to get it attached to the "right" one.  

> I feel like the same amount of work could be better spent on the careful minting/merging/redirecting etc of concept IDs, while providing a more consistent experience for users.

My main concern with the "mint everything" approach is that we might end up fragmenting Concepts due to subtle differences in labels (perhaps reflecting a change in practice over time).  If that is just down to a typo in a few records (e.g. Falliaze vs Fallaize), then that's a clear error that's simple enough to sort out.  OTOH, if half the records are "Florence Nightingale" and the other half are "Nightingale, Florence", and it's left unresolved for too long, users might start collecting bookmarks to the "wrong" one, and we'll need to do some redirecting when it is resolved.   As you say, focusing effort on merging/redirecting from the start might be a better approach, because it's probably inevitable.

kenoir:
This is still useful, but unlikely we'll merge as is or work directly from this branch, so closing this PR.

PR #84, opened 2022-07-06
RFC 051: LoC concepts ingest

jamieparkinson:
very much a wip

### DIFF
--- a/rfcs/051-concepts-adapters/README.md
+++ b/rfcs/051-concepts-adapters/README.md
@@ -1,0 +1,55 @@
+# RFC 051: Ingesting Library of Congress concepts
+
+As per the high-level design of [RFC 052](https://github.com/wellcomecollection/docs/pull/83), one of the key areas of the concepts pipeline topology is an ingest section. This RFC outlines how this will be implemented for the first work phase (attaching only LoC concepts to works).
+
+![High level outline of the architecture in this RFC](https://user-images.githubusercontent.com/4429247/177178393-8fd615f1-f178-4a0a-83b2-a1d5ee09634f.png)
+
+### Retrieving concepts from bulk exports
+LoC provide bulk exports of all [subject headings](https://id.loc.gov/authorities/subjects.html) (LCSH) and [names](https://id.loc.gov/authorities/names.html) (LCNAF). These come in variety of flavours: in both MADS and SKOS vocabularies; and in N-Triple, JSON-LD, Turtle, and XML formats. All of the bulk exports are gzipped.
+
+We can re-run the ingest process on a schedule to capture the (infrequent) changes to LoC concepts, which are almost always additions rather than changes/deletions. 
+
+In the architecture diagram above, the "adapter" and "transformer" are outlined as separate services. In practice, we can perform this mini-ETL operation more efficiently and simply as a streaming operation within one service when we use the gzipped JSON-LD files; these express "one concept per line" and so we needn't worry about traversing any trees or indeed caring much about the underlying RDF.
+
+![A more detailed view of how a concepts ingestor might work](https://user-images.githubusercontent.com/4429247/177571629-56f7dd7b-9c43-4f38-bb11-aae53ed0c2e6.png)
+
+The "transform" step here is (a) the most ill-defined and (b) the only step that needs to differ for LCSH and LCNAF ingestors.
+
+### Transforming external concepts 
+
+[RFC 050](https://github.com/wellcomecollection/docs/tree/main/rfcs/050-concepts-api) outlined the rough design of a concepts API: obviously, we need to store the external concepts in a way that is sufficient to construct these API documents. We also don't want to extraneous data, as the majority of the concepts won't be used in the catalogue. 
+
+Based on this, we might transform the LCSH documents to something like:
+
+```json
+{
+  "id": "<full LCSH URI>", // Necessary for indexing but not used externally
+  "identifier": {
+    "value": "sh12345678",
+    "identifierType": "lc-subjects",
+    "type": "Identifier"
+  },
+  "label": "<value of skos:prefLabel>",
+  "alternativeLabels": [
+    "<values of skos:altLabel>"
+  ]
+}
+```
+### Forwarding changes in external concepts to the concepts augmentation pipeline
+
+> Need to work this bit out
+
+
+
+## Questions
+
+- The format proposed above doesn't follow the existing schema for `identifierType` - rather than just the id we have previously used an object for `identifierType`, like
+  ```json
+  {
+    "id": "lc-subjects",
+    "label": "Library of Congress Subject Headings (LCSH)",
+    "type": "IdentifierType"
+  }
+  ```
+  Is this OK?
+- Is the unused `id` on the documents actually necessary? Could we just use the default ES IDs?


rfcs/051-concepts-adapters/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+- The format proposed above doesn't follow the existing schema for `identifierType` - rather than just the id we have previously used an object for `identifierType`, like
+  ```json
+  {
+    "id": "lc-subjects",
+    "label": "Library of Congress Subject Headings (LCSH)",
+    "type": "IdentifierType"
+  }
+  ```
+  Is this OK?
```
jtweed, line 54: Kind of depends what the docs are for. It’s possibly redundant data, though not exactly enough to be a problem, for internal service data, but something that we will want to add in the API.

rfcs/051-concepts-adapters/README.md
```diff
@@ -0,0 +1,55 @@
...
+- Is the unused `id` on the documents actually necessary? Could we just use the default ES IDs?
```
paul-butcher, line 55: When you wrote "necessary for indexing" on that field in the JSON, is that just "it needs an id", or do we want to look things up using this value that we have found elsewhere?

One thing I expect we will need to do is update existing records, which can only be done if you know the ids for them (unless we choose a throw-away-and-repopulate approach, but then we lose the opportunity for a quick exit when nothing changes, and it becomes a bit more tricky with multiple sources).

If we use an id that is present in the source data, then that is quite simple, because we can transform it into a _bulk body containing that value in _id, and let Elasticsearch do the rest.  Otherwise we have to do something to find out which ids ES assigned to which objects, then update them.

rfcs/051-concepts-adapters/README.md
paul-butcher: I think this is good.  There is plenty of opportunity to add in the redundant data later on, probably in the stage that populates the Index for the API.

The extra verbosity from storing the identifierType object all the way through isn't a problem for the computers, but if we are looking at it in logs or querying an interim DB for debugging, it adds clutter.  I'd rather just see it as a single property with the identifier for the type.

rfcs/051-concepts-adapters/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Transforming external concepts 
+
+[RFC 050](https://github.com/wellcomecollection/docs/tree/main/rfcs/050-concepts-api) outlined the rough design of a concepts API: obviously, we need to store the external concepts in a way that is sufficient to construct these API documents. We also don't want to extraneous data, as the majority of the concepts won't be used in the catalogue. 
+
+Based on this, we might transform the LCSH documents to something like:
+
+```json
+{
+  "id": "<full LCSH URI>", // Necessary for indexing but not used externally
```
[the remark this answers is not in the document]
jamieparkinson: I was thinking along these lines too but @paul-butcher has made a good point below https://github.com/wellcomecollection/docs/pull/84#discussion_r915631087

rfcs/051-concepts-adapters/README.md
jamieparkinson: I concur - perhaps rather than using the URIs straight from LCSH we can construct them in a predictable namespacey way that's transferable to any other identifier scheme, eg `lc-subjects/sh12345678`

rfcs/051-concepts-adapters/README.md
```diff
@@ -0,0 +1,60 @@
...
+}
+
+```
+### Ingesting changes 
+
+_This section is a bit speculative - we don't necessarily need the answers right now._
+
+We can start by just writing all of the source into an empty index - this is inefficient and we won't know what (if anything) changed, but it will never be wrong.
+
+To know what's changed, we have to start comparing to what's there already. This is of course very inefficient, but we can pass that inefficiency onto the ES cluster by using the [`doc_as_upsert`](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html#doc_as_upsert) feature of the `update` API alongside the [`noop`](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html#_detect_noop_updates) feature: with a single API call we'll be able to know whether anything was added.
```
alexwlchan, line 45: To double check I understand: we want to know what's changed because we want to trigger downstream activity from the ES index, right? e.g. _"We have received 500 concepts from LCSH, of which A/B/C have meaningfully changed, so we should do some additional work once they're indexed"_

Also: TIL that Elasticsearch supports [`upsert` as a bulk operation](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-update). I was initially concerned because I thought this meant one `upsert` per LCSH concept, but it sounds like we can do it much more efficiently. 🥳

rfcs/051-concepts-adapters/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+This leaves us with detecting deletions: the least likely kind of change. I don't think this needs to be addressed right now - if it's something we become aware of we can just write into an empty index again.
```
alexwlchan, line 47: Does LCSH remove concepts? Agree we should punt it for now, but I now wonder if it ever actually happens.

rfcs/051-concepts-adapters/README.md
```diff
@@ -0,0 +1,60 @@
...
+  "id": "lc-subjects/sh12345678", // QName-type identifier 
+  "identifier": {
+    "value": "sh12345678",
+    "identifierType": "lc-subjects",
+    "type": "Identifier"
+  },
+  "label": "<value of skos:prefLabel>",
+  "alternativeLabels": [
+    "<values of skos:altLabel>"
+  ]
```
alexwlchan, line 35: I was going to suggest we record some sort of provenance information, e.g. the modified date of the LCSH dump we're using, for debugging purposes – but it's questionably useful, and screws with the upsert logic.

Let's leave that for now.

rfcs/051-concepts-adapters/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+- The format proposed above doesn't follow the existing schema for `identifierType` - rather than just the id we have previously used an object for `identifierType`, like
+  ```json
+  {
+    "id": "lc-subjects",
+    "label": "Library of Congress Subject Headings (LCSH)",
+    "type": "IdentifierType"
+  }
+  ```
+  Is this OK? **Yes, we prefer this**
```
alexwlchan, line 59: 👍

rfcs/051-concepts-adapters/README.md
```diff
@@ -0,0 +1,60 @@
...
+- Is the unused `id` on the documents actually necessary? **It's useful to be able to construct queries for these concepts, and the identifier proposed is scalable to various source schemas**
```
alexwlchan, line 60: I did wonder if we need the `id` on the document itself; rather than as the `_id` of the Elastic document.

Presumably if we wanted to query by ID, we'd query on the structured `identifier` object instead?

I can think of scenarios where it's useful, and it does no harm, so I think it's fine.

alexwlchan (approved):
Looks fine. A couple of minor thoughts, but definitely not blocking and probably just because I haven't been in all these conversations; more curiosity than anything else.

rfcs/051-concepts-adapters/README.md
jamieparkinson: Correct, will make this crystal clear in the text

rfcs/051-concepts-adapters/README.md
jamieparkinson: Not sure to be honest!

rfcs/051-concepts-adapters/README.md
jamieparkinson: Will note this down as I think it might be achievable

rfcs/051-concepts-adapters/README.md
jamieparkinson: I did actually mean this to be the `_id` so will correct that

PR #85, opened 2022-11-25
Lambda logging RFC

jamieparkinson:
Rendered version: https://github.com/wellcomecollection/docs/blob/lambda-logging-rfc/rfcs/053-lambda-logging/README.md

### DIFF
--- a/rfcs/053-lambda-logging/README.md
+++ b/rfcs/053-lambda-logging/README.md
@@ -54,10 +54,10 @@ The work plan might look something like this:
 - Provision an Elastic serverless forwarder
 - Create a log group subscription and verify that the forwarder works
 - Create a Terraform module for Lambdas with log groups that subscribe to the forwarder
-- Roll this out across our Lambdas
+- Roll this out across our Lambdas - if we don't use a cross-account Kinesis stream, we'll want to have a log-streaming Lambda per account, like we have with Slack alerts.
 
 ## Caveats and questions
 
-- We've seen CloudWatch logs for Lambdas go missing before - is this a risk? We believe this is due to (mis)configuration of log retention periods, but we're not sure.
+- We've seen CloudWatch logs for Lambdas go missing before - is this a risk? We believe this might be due to (mis)configuration of log retention periods and/or IAM permissions, but we're not sure. Using a consistent Terraform module across our Lambdas will hopefully resolve these issues.
 - Should we use Kinesis or not? Some reasons for using it are given above: I think it would be a good idea.
-- What index should Lambda logs go into? The ECS logs go into `firelens-<yyyy.mm.dd>` indices - we probably shouldn't do the same for these, as they aren't using firelens and may have a different schema.
+- What index should Lambda logs go into? The ECS logs go into `firelens-<yyyy.mm.dd>` indices - we probably shouldn't do the same for these, as they aren't using firelens and may have a different schema. On the other hand, we would like all of our logs to fall under the same index pattern.
--- a/rfcs/053-lambda-logging/README.md
+++ b/rfcs/053-lambda-logging/README.md
@@ -1,0 +1,39 @@
+# RFC 053: Logging in Lambdas
+
+In [RFC 022](https://github.com/wellcomecollection/docs/tree/main/rfcs/022-logging), we identified that our application logs - which were then stored in Cloudwatch - were costing us money, were hard to query, and were inconsistent. We proposed (and went on to implement) an architecture in which ECS services contained a logging sidecar container that used Fluent Bit to stream logs directly to an Elasticsearch cluster. This continues to serve us well.
+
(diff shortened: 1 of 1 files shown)


alexwlchan (commented):
This looks like a good summary of the options; I wrote down a quick list of ideas before opening your RFC and you got all the things I thought of.

Is the intention that this eventually becomes the logging approach for all Lambdas? e.g. the window generator in the catalogue pipeline, or the snapshot Lambdas in the catalogue API repo. I can imagine we might not migrate them all in one go, but pick it up the next time we modify those functions – or would some function logs stay in CloudWatch forever?

There's an implicit requirement that logs are delivered reliably. I don't know if you think that's worth documenting; it sprung to mind because our current Lambda logging is definitely a bit hit-and-miss. We have functions that are dropping logs between CloudWatch and Lambda (because there's no consistency in how they're deployed 😬).

Seeing it laid out like this, I have opinions on which we should go for, but not sure if you want that level of commenting yet.

alexwlchan (commented):
Offering some opinions on the approaches:

## Stream CloudWatch logs to Elasticsearch via another Lambda

This feels like it would be the easiest to implement *if* we could sort out reliably delivery of logs from Lambda to CloudWatch Groups.

Currently every Lambda has a bespoke implementation which means some do and some don't; if we had a shared Lambda module that we could plug the function into and have it do all the IAM permissions for logs, that would be quite good.

I imagine we might have a single logging Lambda per account, which aggregates logs from all the log groups:

```mermaid
graph LR
   F1[λ function] --> L[shared logging Lambda]
   F2[λ function] --> L
   F3[λ function] --> L
```

This is a pattern we already have for Lambda alerting, which AFAIK works pretty well.

This would be configurable entirely in Terraform and doesn't need any changes to how we deploy Lambda application code.

## Using Lambda extensions/Docker images

This bypasses CloudWatch entirely and feels more "correct" – it's cutting out an unnecessary service and is what Lambda extensions are for. Possibly more complicated to implement; I don't know enough about how Lambda extensions work. It also avoids the Lambda&rarr;CloudWatch reliability snargle.

It's beyond the immediate scope of this work, but I wonder if it also shows a way to improve our Lambda app packaging story, which is a mess across different repos.

If we standardised on Docker for logging, then we'd need to use Docker to package our Lambdas – which as the concepts pipeline shows, isn't difficult and allows us to offload the heavy lifting to Buildkite and Docker Compose. That would be a good simplification to gradually apply.

In an ideal world, all our Lambdas would use the same approach to logging, but it's a lot of work to apply it in one go. I think we should leave anything that doesn't need active debugging as-is for now. e.g. the Sierra adapter doesn't need touching, but next time we need to modify it, we definitely get it into the new logging, and possibly Docker-ize it as well.

rfcs/053-lambda-logging/README.md
```diff
@@ -0,0 +1,63 @@
...
+*Option (1) does not preclude option (2) as further work*: we would already be providing a Terraform module for consumers, and if the Elastic Serverless Forwarder proved insufficient, we would already have written a module to stream the correct format of logs to Elasticsearch. 
+
+This fact clarifies that option (1) - streaming of logs from CloudWatch - is the better starting point. If it proves insufficient or impractical, the work we will have done already will mostly still be useful for option (2).
+
+The work plan might look something like this:
+
+- Provision an Elastic serverless forwarder
+- Create a log group subscription and verify that the forwarder works
+- Create a Terraform module for Lambdas with log groups that subscribe to the forwarder
+- Roll this out across our Lambdas
```
alexwlchan, line 57: Is this a single forwarder for all our accounts, or would we have a serverless forwarder per-account?

I'm envisioning something like https://github.com/wellcomecollection/platform-infrastructure/blob/main/monitoring/terraform/slack_alert_on_lambda_errors.tf but for forwarders.

rfcs/053-lambda-logging/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Caveats and questions
+
+- We've seen CloudWatch logs for Lambdas go missing before - is this a risk? We believe this is due to (mis)configuration of log retention periods, but we're not sure.
```
alexwlchan, line 61: I think this needs more investigation; if you get a working module that we use consistently, I think it will be okay. The reason Lambda logging is so flaky is because it's all copy-pasted between configurations.

I think log retention periods OR misconfigured IAM permissions are the cause, because I've definitely seen Lambdas that didn't have permissions to use their log groups.

rfcs/053-lambda-logging/README.md
```diff
@@ -0,0 +1,63 @@
...
+- Should we use Kinesis or not? Some reasons for using it are given above: I think it would be a good idea.
+- What index should Lambda logs go into? The ECS logs go into `firelens-<yyyy.mm.dd>` indices - we probably shouldn't do the same for these, as they aren't using firelens and may have a different schema.
```
alexwlchan, line 63: I don't mind, but I don't think it should include the word "Lambda".

Yesterday you mentioned using some sort of Elastic common schema, which isn't mentioned above. If we do use a properly generic schema that we might use for other, non-Lambda applications in the future, let's not box ourselves into a corner with naming.

rfcs/053-lambda-logging/README.md
jamieparkinson: That is sort of to-be-determined by whether we'd use Kinesis. I'll make that explicit.

rfcs/053-lambda-logging/README.md
jamieparkinson: Yeah, I would rather the ECS one wasn't firelens tbh!

PR #86, opened 2022-12-02
Prismic docs

melanierogan:
I don't have 'creator' access on GitBook, so I'm having trouble adding my documentation there. I'm not sure when the courier is arriving to take my laptop, so I want to be sure this documentation exists somewhere in Wellcome before then.


melanierogan:
I'm closing this as i've added a Prismic section to the Platform collection on GitBook. I have not yet made it public, to give people a chance to edit and approve.

PR #87, opened 2022-12-09
RFC 054: Authoritative ID Resolver

paul-butcher:
This shows the broad idea of a new catalogue pipeline stage to improve the way we handle authoritative ids coming out of the transformer. ([rendered](https://github.com/wellcomecollection/docs/blob/054-id-resolver/rfcs/054-improving-concept-source-ids/README.md))

TODO:
* Add some proper facts - not "some" and "a number" of things, put proper examples in
* Find an example of a compound constituent that matches a top-level concept

I'm currently rearranging this and expanding on it to focus on the problems it needs to fix, and to include/promote a solution that defocuses concept types in the id minter.

### DIFF
--- a/rfcs/054-id-resolver/README.md
+++ b/rfcs/054-id-resolver/README.md
@@ -34,7 +34,8 @@ This is due to a rule in the Sierra Transformer that exists because a Personal N
 a Person (if it has a $t subfield).
 
 ca. 2700 examples where a 700 with $t uses a code in $0 that refers to the Person in $a
-e.g. in https://wellcomecollection.org/works/hqqndr39 (b20241720) "Maimonides, Moses, 1135-1204. Maqālah fī sināʾat al-mantiq." has the id n78096039, which refers to the Person, and not 17115004, which refers to the text.
+e.g. in https://wellcomecollection.org/works/hqqndr39 (b20241720) "Maimonides, Moses, 1135-1204. Maqālah fī sināʾat al-mantiq." has the id n78096039, 
+which refers to the Person, and not 17115004, which refers to the text.
 ca 800 examples where a 700 with a $t uses a code in $0 that refers to the text.
 e.g. in https://wellcomecollection.org/works/nfnmnc6m (b20241781) (Karo, Joseph ben Ephraim, 1488-1575. Kesef mishneh) (not Karo, Joseph ben Ephraim)
 
@@ -61,6 +62,80 @@ In addition, only the lc-names version will be enhanced with Authority data.  Th
 It may be more correct to exclude any "wrong" identifiers as part of the Reccorder, but that would lead to the omission of all identifiers 
 whose identifier type has been incorrectly guessed.
 
+### Problems of homonymy
+
+#### The Aberdeen problem - homonyms with different types
+
+This is a hypothetical problem for which I have seen no evidence, but should be mentioned in case any solution to the above
+problems might cause it to arise.
+
+Unidentified Concepts derive their identifiers from their labels.  It may be the case that multiple genuinely different 
+referents have the same name, but are of different types (e.g. Aberdeen, the Granite City;  vs. Aberdeen, the now disemvowelled company founded there)
+
+#### The John Smith problem - homonyms with the same type
+
+There are evidently multiple authors by the name "John Smith" (https://wellcomecollection.org/concepts/t4jfabht), as they 
+have authored works roughly 340 years apart.  In the case of t4jfabht, this corresponds to lc-names:n86851637,
+but there may be some homonymous unidentified concepts.
+
+This is not something that can be resolved by the pipeline(s).  It must be fixed by 
+
+### Goals
+
+```gherkin
+Feature: Concepts presented in the Works and Concepts APIs
+```
+The scenario representing the end goal for new Concepts with authoritative ids is clear
+```
+Scenario: Concepts with the same authoritative id
+Given two Concepts with the same authoritative id
+When the Concepts reach the end of the Works and Concepts pipelines
+Then both Concepts have the same canonical identifier
+And both Concepts have the same type
+And the type accurately represents the real-world type of the referent
+```
+
+There is less clarity on non-authoritative ids (label-derived)
+```
+Scenario: Concepts with the same non-authoritative id
+Given two Concepts with the same label-derived id
+When the Concepts reach the end of the Works and Concepts pipelines
+Then (what happens? same as above?)
+```
+
+This is because of the hypothetical "Aberdeen Problem" - two concepts with the same name
+that legitimately have different types.  I suspect that in this case, it should be treated 
+in the same way as authoritative Concepts, with the Concept type not being used for differentiation.
+If the Aberdeen problem arises here, it should be handled in the source data by adding an
+authoritative id or differentiating the name.
+
+The `Concepts with the same authoritative id` scenario, above, applies to new and existing concepts alike.
+However, existing concepts have these further considerations:
+
+Wrongly-namespaced identifiers are to be fixed. There is no need to retain the "wrong" canonical id.
+
+```
+Scenario: Wrong identifier scheme
+Given a Concept that is currently in the wrong scheme
+Then the Concept's canonical id will not be the same as it currently is
+```
+
+Correctly-namespaced identifiers should ideally retain their existing canonical id, but this is not always possible.
+In instances of the Maimonedes problem, one of the Concepts will end up with a new canonical identifier.
+
+I don't know how the winner is to be chosen.  Possibly the most specific available?
+
+```
+Given a Concept that is currently in the right scheme
+And the Concept is already in the id minter database
+Then the Concept's canonical id should ideally be the same as it currently is
+
+Given two matching Concepts that are already in the id minter database
+And the two Concepts currently have different canonical ids.
+Then the Concepts' shared canonical id should be the canonical id that is already assigned to one of them.
+```
+
+
 ### Options
 
 Logically, there are three options for resolving this problem.
@@ -85,6 +160,7 @@ Non-LoC schemes (e.g. MeSH), can also be determined reliably from the combinatio
 TEI data also currently just distinguishes between LCSH (or any LoC) and MeSH in the same way. Other schemes are not used.
 
 #### Perfecting the extraction of Concept types
+
 This is impossible, for reasons outlined in the two "Problems", above.  
 There is no way to reliably determine the type of the referent from Sierra MARC data.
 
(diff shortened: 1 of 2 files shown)


rfcs/054-id-resolver/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+### LoC namespace
+
+MARC data from Sierra does not distinguish between LCSH and LCNames. They are both signalled by the use of `indicator2=0`.
+As with type, above, the namespace is inferred from the field.  In "Name" fields (`x00`, `x10`, `x11`), it is assumed to be
+LCNames. In "Term" fields (`648`, `650`, `651`), it is assumed to be LCSH.
+
+This is not reliable.  Many "Geographical Terms" are in fact Names of Places.
+
+I believe that this could be resolved by examining the first character of the identifier (n for LCNames, s for LCSH).
```
jamieparkinson, line 45: Could (should?) this specific function (ie inferring identifier type from prefix) be done in the transformer?

rfcs/054-id-resolver/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Compound Concepts
+
+Compound concepts in MARC data from Sierra bear an identifier referring to the whole concept, but the transformer
+also breaks them up into the individual constituent concepts.  However, the identifiers for these constituent concepts 
+are only "label-derived".  LoC and MeSH entries for compound concepts also contain the identifiers for the constituents,
+so a new stage could populate these properly.
```
jamieparkinson, line 52: Think this section could do with expanding a bit

rfcs/054-id-resolver/README.md
```diff
@@ -0,0 +1,155 @@
...
+* Some MeSH identifiers have been found wrongly assigned to the LCSH scheme.  
+* Two LCNames identifiers (n50034502 and n2001003970) do not exist
+* Some identifiers in the LCSH scheme have been parsed/written incorrectly, so they are not identifiers at all.
+* The wrong field was being extracted from some TEI records to find the LCSH id.
+
+In addition, there may be some that are very difficult to detect by eye.  A real identifier that refers to something different
+to the label of the field would only be detected when someone follows a link and finds themselves on a wholly unexpected Concept page.
+`n2001003970`, mentioned above, does not exist but `nr2001003970` does.  
+However, that identifer refers to the "Partnership for Governance Reform in Indonesia", whereas
+the associated label in the offending Work is "Pseudo Aristotele., De coloribus., Latin".
```
jamieparkinson, line 71: Interesting - would question whether we should be doing this as part of the pipeline, perhaps more appropriate to script some identifier validation so we can pass affected records on to collections info

rfcs/054-id-resolver/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+It may belong in the ID Minter.  There is some duplication between this task and the id minter, in the extraction of source
+identifiers from the input Work. The only reasons for keeping it separate are - 
+
+* better observability (weak, it's probably just as clear what has happened if resolution and minting are in the same stage)
+* the minter is fairly critical and may be best left alone (also weak, it's well covered in tests) 
+
+## Current Problem Manifestations
+
+### The Person/Agent (Maimonides) problem
```
jamieparkinson, line 128: I might be misunderstanding - is this a data problem (ie to be resolved with collections info) or is it fundamental to the problem ie unavoidable?

rfcs/054-id-resolver/README.md
```diff
@@ -0,0 +1,155 @@
...
+are correct and consistent before new Canonical Identifiers are minted for them.
+
+## How?
+
+### Concepts Pipeline/API
+The Concepts Pipeline Ingestor(s) should, where available, store the official types (maybe translated to our types) 
+and compositions of each Concept in the authoritative-concepts index.
+
+The Concepts API should have a service that can be called with one or more authoritative identifiers, and returns that information
+from the authoritative index.  This may be a separate private/internal API instance.
```
jamieparkinson, line 96: I think we could probably make the normal Concepts API do this, right?

jamieparkinson (commented):
Some initial thoughts - I think I am a bit confused about whether all 4 of the problems you've outlined do in fact require lookups

rfcs/054-id-resolver/README.md
paul-butcher: I think it could (and probably should), but without pairing it with determining the type, it would turn the Glasgow problem (of two different Glasgows) into another example of the Person/Agent problem (where one of the ids is missing)

paul-butcher: However, there is another solution for the Maimonides problem - which is not to use the type as part of the identifier.  I'll expand on that in another section

rfcs/054-id-resolver/README.md
paul-butcher: I think that would be simplest.

rfcs/054-id-resolver/README.md
paul-butcher: I think there are two modes for this 

- find all the broken ones and get someone to fix them
- don't let any new broken ones get through the pipeline

If we can add identifier validation even earlier, so that whoever is typing it in gets a warning immediately, then that would be nice.

rfcs/054-id-resolver/README.md
paul-butcher: This is a good point.  I had thought of it as needing a technological solution, due to the scale of all of this.  However, I should have a proper look at the scale of this particular one, and seek counsel from an information professional.

rfcs/054-id-resolver/README.md
alexwlchan: We've talked in the past about the pipeline having a way to signal errors in the data, of which this would definitely be an example.

alexwlchan (commented):
Thanks for writing this up, it's a tricky one and clearly our identifier model is a bit creaky.

I'm not convinced by the need for a new stage yet, but maybe I don't understand the problem well enough.

Some thoughts on the four problems:

1. Type inference – a few weeks ago you asked me why we include the ontologyType when deriving canonical IDs, e.g. why is `Agent sh1234` different to `Person sh1234`. We do need it for some cases, but I wonder if it should be the exception rather than the rule – e.g. Miro images/items/works

    Spitballing here:

    * You've suggested creating an "LC" identifier type that catches both lc-names and lcsh; what if we split the miro-image-number identifier type into works/images/items and then remove ontologyType from the ID minter.
    * In the ingestors, we collapse the miro-image-number-{works, images, items} types into the miro-image-number identifier type we display publicly.
    * The ID minter just mints IDs based on (source system, source identifier). If it gets an LCSH identifier as both Agent and Person, they get the same identifier.

    (You've hinted at this in a comment about the Maimonides problem, sorry if I'm repeating something said elsewhere.)

2. LoC namespace – I would leaning on inferring identifier type from prefix in the transformer, if we can solve the type inference problem.

3. Compound concepts – does this need a new stage, or could this part of the problem be solved by the concepts embedder?

4. Validation – this is something we can maybe help with, but I'm not sure if it belongs in the catalogue pipeline. Or if it does, we should consider solving the more general problem "how do we report errors" and not build something special for just this case.

paul-butcher:
> Thanks for writing this up, it's a tricky one and clearly our identifier model is a bit creaky.
> 
> I'm not convinced by the need for a new stage yet, but maybe I don't understand the problem well enough.
> 
> Some thoughts on the four problems:
> 
> 1. Type inference – a few weeks ago you asked me why we include the ontologyType when deriving canonical IDs, e.g. why is `Agent sh1234` different to `Person sh1234`. We do need it for some cases, but I wonder if it should be the exception rather than the rule – e.g. Miro images/items/works
>    Spitballing here:
>    
>    * You've suggested creating an "LC" identifier type that catches both lc-names and lcsh; what if we split the miro-image-number identifier type into works/images/items and then remove ontologyType from the ID minter.
>    * In the ingestors, we collapse the miro-image-number-{works, images, items} types into the miro-image-number identifier type we display publicly.
>    * The ID minter just mints IDs based on (source system, source identifier). If it gets an LCSH identifier as both Agent and Person, they get the same identifier.
>    
>    (You've hinted at this in a comment about the Maimonides problem, sorry if I'm repeating something said elsewhere.)

Yes, that's essentially what I was hinting at, but I haven't actually written it down, so you're not repeating anything.

> 2. LoC namespace – I would leaning on inferring identifier type from prefix in the transformer, if we can solve the type inference problem.

I think this combination is probably the way to go - infer the namespace from the prefix and don't include types as part of the key for identifiers, (authoritative ones at least, but although it may be theoretically valuable to include them in label derived ones, (what we have called the Aberdeen Problem in some conversations, not written about here at all), I don't think it would actually help in practice).  This would fix both the Maimonides and Glasgow problems.

I was reluctant to go that way because the id minter currently relies on them, so this would mean a whole set of new ids for all concepts, and it's a potentially significant change to the minter.

> 3. Compound concepts – does this need a new stage, or could this part of the problem be solved by the concepts embedder?

I think it could be done in the concepts embedder, but I think it is more efficient to do it before the id minter (or, as I have come to realise by writing all this down, as part of the id minter, just before minting new ids).

This could be a simple fact to determine between transformer and minter - sh2010112128 (School Sports -- Fiction) is made up of sh85118414 and sj2020050019.

If we leave it to the Concepts Embedder and the concepts-store, we need to use the knowledge graph to work out that sh85118414 and label-derived:schoolsports are synonyms.

> 4. Validation – this is something we can maybe help with, but I'm not sure if it belongs in the catalogue pipeline. Or if it does, we should consider solving the more general problem "how do we report errors" and not build something special for just this case.

I think the validation part of this RFC is the weakest, and I've probably caused confusion by including it.  My thought was that this is a problem that exists, and that if we did have the facility to check IDs against the official registry (to support fixing the other problems), it would, coincidentally, also include a bit of validation that we can use to report required changes back to the authors.

rfcs/054-id-resolver/README.md
```diff
@@ -0,0 +1,155 @@
...
+The Sierra Transformer should stop guessing whether to use LCNames or LCSubjects, and instead use a new scheme, "LC".
+
+The Catalogue Pipeline should have a new stage between the transformers and the id minter, the new stage:
+
+* extracts identifiers in the same way the id minter does, and requests the official type/composition for any that are in a scheme supported by the API.
+* replaces the type and identifiertype with the "right" value from the API
+* replaces the identifier values of any constituent concepts that have label-derived identifiers, with the official ones (based on similarity of label).
+* rejects and/or warns about documents containing any wrong identifiers (initially, just not known in that scheme, later, consider some comparison between labels).
+
+Everything downstream will behave exactly as it currently does.
```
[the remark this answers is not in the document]
paul-butcher: It would have just been for internal use whilst uncertain.  Having given it a bit more thought, it would be better to infer the scheme from the prefix.

alexwlchan:
> I was reluctant to go that way because the id minter currently relies on them, so this would mean a whole set of new ids for all concepts, and it's a potentially significant change to the minter.

Yes, we'd need to think carefully about how we do it, and I think we might do it in steps rather than all in one go. I can imagine breaking it into several RFCs.

rfcs/054-id-resolver/README.md
paul-butcher: Yes, and in expanding it, I have discovered that the MADS data is insufficient to fix this problem :(

PR #90, opened 2023-02-17
Content API RFC

harrisonpim:
[Rendered version of the RFC](https://github.com/wellcomecollection/docs/blob/prismic-api-rfc/rfcs/055-prismic-api/README.md) 

This PR closes https://github.com/wellcomecollection/platform/issues/5661

### DIFF
--- a/rfcs/054-prismic-api/README.md
+++ b/rfcs/054-prismic-api/README.md
@@ -1,0 +1,48 @@
+# RFC 054: Prismic API
+
+This RFC outlines a new set of API endpoints which will allow wellcomecollection.org users to search and filter content which is stored in Prismic.
+
+## Background
+
+We use Prismic as the CMS for exhibitions, events, stories, and other pieces of non-catalogue content on wellcomecollection.org.
+
+The MVP version of the site's prismic search functionality uses [Prismic's GraphQL endpoint](https://prismic.io/docs/graphql) to enable stories search. That MVP has demonstrated that prismic's search functionality isn't able to produce sufficiently relevant results on its own. We would like to replace it with something more configurable, akin to the system we have for the catalogue content.
+
+We're building a pipeline to ingest prismic content into elasticsearch (see separate RFC).
+
+We'd like to build a new corresponding API which will allow users to search for those prismic documents in elasticsearch.
+
+## Requirements, considerations, constraints
+
+- The API should only return enough information for users to figure out whether a result is relevant to them, and provide a link to the relevant page on wellcomecollection.org. The content of the pages themselves should still be fetched from prismic directly.
+- Though we're describing them as a separate project, the new endpoints should seamlessly fit into the rest of the wellcomecollection.org API suite from a user POV. All of the existing conventions should be followed.
+- The API url structure should also be consistent with what appears on the front end of the site. For example, an article on the site appears at `/articles/{id}`, so the API equivalent should be `/articles/{id}`. The same should be true for search results.
+- The new API service should be written in typescript, following patterns set by the [concepts API](../050-concepts-api/README.md) for filtering, pagination, error handling, etc.
+- We shouldn't create new IDs for exhibitions, events, stories, etc. There aren't any situations where we should need to merge content from multiple sources, so we should use the document IDs directly from prismic.
+- The elasticsearch index mapping should represent the contract between the pipeline and the API. The API shouldn't need to know anything about the structure of the data in prismic.
+
+## Proposed endpoints
+
+### `/articles`
+
+Filter for type eg stories, webcomics
+
+### `/exhibitions`
+
+Filters for date, location, status, etc
+
+### `/events`
+
+Filters for date, location, status, etc
+
+## Open questions
+
+- How do we represent functional content?
+- Should we include the full prismic API response at eg `/articles/{id}`?
+- If we want to avoid maintaining multiple versions of the same code, should this work be done as an extension to the concepts API?
+- Works exist at [api.wellcomecollection.org/catalogue/v2/works](api.wellcomecollection.org/catalogue/v2/works), concepts at [/catalogue/v2/concepts](api.wellcomecollection.org/catalogue/v2/concepts), images at [/catalogue/v2/images](api.wellcomecollection.org/catalogue/v2/images).  
+Stories etc aren't part of the catalogue - should the URL be different?
+
+## Next steps
+
+Get building?
--- a/rfcs/054-prismic-api/README.md
+++ b/rfcs/054-prismic-api/README.md
@@ -27,22 +27,84 @@ We'd like to build a new corresponding API which will allow users to search for 
 
 Filter for type eg stories, webcomics
 
+#### Different kinds of thing (are these 'labels'?)
+
+- story <https://wellcomecollection.org/articles/Y9vktREAACYA1v2U> at /articles
+- webcomic <https://wellcomecollection.org/articles/Y86rshEAACcAI4g4> at /articles
+- book <https://wellcomecollection.org/books/Y7QWrREAAHC43oH0> at /books
+- series <https://wellcomecollection.org/series/WfpEBCoAACrdd43y> composed of articles
+- in pictures <https://wellcomecollection.org/articles/Y-oAIhAAACAAix32> at /articles
+
+#### Filters
+
+- date
+- contributor
+- series
+- contentType
+
+#### Sort
+
+- date
+- relevance
+
 ### `/exhibitions`
 
-Filters for date, location, status, etc
+#### Filter
+
+- date
+- status
+- location
+- contributor?
+- format (ie permanent, exhibition, installation)
+- season
+
+#### Sort
+
+- date
+- relevance
 
 ### `/events`
 
-Filters for date, location, status, etc
+#### Filter
+
+- date
+- accessibility status (is it audio described, signed, etc)
+- location
+- format (eg discussion)
+- event-series
+- season
+- audience (eg schools)
+- contributor
+- isOnline?
+- availableOnline?
+
+#### Sort
+
+- date
+- relevance
+
+### `/books`
+
+?????????
 
 ## Open questions
 
 - How do we represent functional content?
 - Should we include the full prismic API response at eg `/articles/{id}`?
+    - should we be able to list all of the docs of a given type at once? what should that url look like?
 - If we want to avoid maintaining multiple versions of the same code, should this work be done as an extension to the concepts API?
 - Works exist at [api.wellcomecollection.org/catalogue/v2/works](api.wellcomecollection.org/catalogue/v2/works), concepts at [/catalogue/v2/concepts](api.wellcomecollection.org/catalogue/v2/concepts), images at [/catalogue/v2/images](api.wellcomecollection.org/catalogue/v2/images).  
 Stories etc aren't part of the catalogue - should the URL be different?
+- what are the top level types? is it articles? stories? series?
+- where do books go? they appear on the stories landing page, so should they appear in stories search results? or are they at a separate endpoint?
+- what is a serial? what is a series?
+- is 'In pictures' the same as 'Image gallery'?
+- Should we be able to filter by contributor role? is that something for the first pass or something we address later? which contributor roles are there? yes for stories? no for exhibitions and events?
+- can we filter exhibitions for accessibility, ie whether a digital guide exists? we can for events, but unclear for exhibitions
+- do we have any exhibition contributors who have contributed to more than one? is a filter for exhibition contributors really going to be useful?
 
 ## Next steps
 
-Get building?
+- Answer most of those questions ☝️
+- Write a glossary of terms
+- Get building?
--- a/rfcs/055-prismic-api/README.md
+++ b/rfcs/055-prismic-api/README.md
@@ -212,7 +212,7 @@ GET /events?query=foo&filters=bar&sort=baz
     "pageSize": 100,
     "totalResults": 1000,
     "totalPages": 10,
-    "nextPage": "/events?query=foo&filters=bar&sort=baz?page=2",
+    "nextPage": "/events?query=foo&filters=bar&sort=baz&page=2",
 }
 ```
 
--- a/rfcs/055-prismic-api/README.md
+++ b/rfcs/055-prismic-api/README.md
@@ -1,4 +1,4 @@
-# RFC 055: Prismic API
+# RFC 055: Content API
 
 This RFC outlines a new set of API endpoints which will allow wellcomecollection.org users to search and filter content which is stored in Prismic.
 
@@ -64,6 +64,3 @@ Read more about the structure of the [exhibitions endpoint](exhibitions.md)
 `https://api.wellcomecollection.org/content/v0/events`
 
 Read more about the structure of the [events endpoint](events.md)
-
-## Open questions
-
--- a/rfcs/055-prismic-api/README.md
+++ b/rfcs/055-prismic-api/README.md
@@ -1,5 +1,9 @@
 # RFC 055: Content API
 
+**Status:** Draft
+
+**Last updated:** 2023-03-06
+
 This RFC outlines a new set of API endpoints which will allow wellcomecollection.org users to search and filter content which is stored in Prismic.
 
 ## Background
--- a/rfcs/055-prismic-api/README.md
+++ b/rfcs/055-prismic-api/README.md
@@ -14,6 +14,7 @@ We'd like to build a new corresponding API which will allow users to search for 
 
 ## Requirements, considerations, constraints
 
+- We'll be considering articles, exhibitions, and events in this first pass. Books won't be included for now, but could be added at a later date.
 - The API should only return enough information for users to figure out whether a result is relevant to them, and provide a link to the relevant page on wellcomecollection.org. The content of the pages themselves should still be fetched from prismic directly.
 - Though we're describing them as a separate project, the new endpoints should seamlessly fit into the rest of the wellcomecollection.org API suite from a user POV. All of the existing conventions should be followed.
 - The API url structure should also be consistent with what appears on the front end of the site. For example, an article on the site appears at `/articles/{id}`, so the API equivalent should be `/articles/{id}`. The same should be true for search results.
@@ -235,7 +236,7 @@ GET /events?query=foo&filters=bar&sort=baz
 - How do we represent functional content?
 - If we want to avoid maintaining multiple versions of the same code, should this work be done as an extension to the concepts API?
 - Works exist at [api.wellcomecollection.org/catalogue/v2/works](api.wellcomecollection.org/catalogue/v2/works), concepts at [/catalogue/v2/concepts](api.wellcomecollection.org/catalogue/v2/concepts), images at [/catalogue/v2/images](api.wellcomecollection.org/catalogue/v2/images).
-Stories etc aren't part of the catalogue - should the URL be different?
+Stories etc aren't part of the catalogue - should the URL be different? Does `/content` work?
 - what are the top level types? is it articles? stories? series?
 - where do books go? they appear on the stories landing page, so should they appear in stories search results? or are they at a separate endpoint?
 - what is a serial? what is a series?
--- a/rfcs/055-prismic-api/README.md
+++ b/rfcs/055-prismic-api/README.md
@@ -6,22 +6,21 @@ This RFC outlines a new set of API endpoints which will allow wellcomecollection
 
 We use Prismic CMS to edit and store exhibitions, events, stories, and other pieces of non-catalogue content on wellcomecollection.org.
 
-The site currently allows users to [search for stories](https://wellcomecollection.org/search/stories) using [Prismic's GraphQL API](https://prismic.io/docs/graphql). That MVP implementation has demonstrated that prismic's search functionality isn't good enough to produce relevant results on its own. We would like to replace it with something more configurable, like the system we have for the catalogue.
+The site currently allows users to [search for stories](https://wellcomecollection.org/search/stories) using [Prismic's GraphQL API](https://Prismic.io/docs/graphql). That MVP implementation has demonstrated that Prismic's search functionality isn't good enough to produce relevant results on its own. We would like to replace it with something more configurable, like the system we have for the catalogue.
 
-We're building a pipeline to ingest prismic content into elasticsearch (see separate RFC).
-
-We'd like to build a new corresponding API which will allow users to search for those prismic documents in elasticsearch.
+We're building a pipeline to ingest Prismic content into elasticsearch (see separate RFC).  
+We also need a corresponding API which will allow users to search for those Prismic documents in elasticsearch.
 
 ## Requirements, considerations, constraints
 
 - We'll be considering articles, exhibitions, and events in this first pass. Books won't be included for now, but could be added at a later date.
-- The API should only return enough information for users to figure out whether a result is relevant to them, and provide a link to the relevant page on wellcomecollection.org. The content of the pages themselves should still be fetched from prismic directly.
+- The API should only return enough information for users to figure out whether a result is relevant to them, and provide a link to the relevant page on wellcomecollection.org. The content of the pages themselves should still be fetched from Prismic directly.
 - Though we're describing them as a separate project, the new endpoints should seamlessly fit into the rest of the wellcomecollection.org API suite from a user POV. All of the existing conventions should be followed.
-- Despite prioritising search and defaulting to the prismic API for display, We'll still include a way to fetch a single document by ID, eg `/articles/{id}`.
+- Despite prioritising search and defaulting to the Prismic API for display, We'll still include a way to fetch a single document by ID, eg `/articles/{id}`.
 - The API's URL structure should also be consistent with what appears on the front end of the site. For example, if article on the site appears at `/articles/{id}`, the API equivalent should be at `/articles/{id}`. The same should be true for search results.
 - The new API service should be written in typescript, following patterns set by the [concepts API](../050-concepts-api/README.md) for filtering, pagination, error handling, etc.
-- We shouldn't create new IDs for exhibitions, events, stories, etc. There aren't any situations where we should need to merge content from multiple sources, so we should use the document IDs directly from prismic.
-- The elasticsearch index mapping should represent the contract between the pipeline and the API. The API shouldn't need to know anything about the structure of the data in prismic.
+- We shouldn't create new IDs for exhibitions, events, stories, etc. There aren't any situations where we should need to merge content from multiple sources, so we should use the document IDs directly from Prismic.
+- The elasticsearch index mapping should represent the contract between the pipeline and the API. The API shouldn't need to know anything about the structure of the data in Prismic.
 - We won't include an endpoint for `/series`, nor will we include series nodes in the `/articles` endpoint. If an article is part of a series, we'll denormalise information from its parent onto the article itself to use for relevance and visual signposting in results.
 
 ## Proposed endpoints
@@ -241,6 +240,7 @@ GET /events?query=foo&filters=bar&sort=baz
 - If we want to avoid maintaining multiple versions of the same code, should this work be done as an extension to the concepts API?
 - Works exist at [api.wellcomecollection.org/catalogue/v2/works](api.wellcomecollection.org/catalogue/v2/works), concepts at [/catalogue/v2/concepts](api.wellcomecollection.org/catalogue/v2/concepts), images at [/catalogue/v2/images](api.wellcomecollection.org/catalogue/v2/images).
 Stories etc aren't part of the catalogue - should the URL be different? Does `/content` work?
+- Should we include a `/series` endpoint as part of this work?
 
 ## Next steps
 
--- a/rfcs/055-prismic-api/README.md
+++ b/rfcs/055-prismic-api/README.md
@@ -25,15 +25,48 @@ We'd like to build a new corresponding API which will allow users to search for 
 
 ### `/articles`
 
-Filter for type eg stories, webcomics
+#### Example response
 
-#### Different kinds of thing (are these 'labels'?)
+```json
+GET /articles/{id}
+{
+    "id": "",
+    "format": "",
+    "title": "",
+    "body": "", // not exposed by default, should require an ?includes=body
+    "standfirst": "",
+    "published": "",
+    "contributors": [
+        {
+            ... // reuse the Contributor model
+        },
+        ...
+    ],
+    "image": {
+        "thumbnail": {
+            ... // reuse the DigitalLocation model
+        },
+        "caption": ""
+    }
+}
+```
 
-- story <https://wellcomecollection.org/articles/Y9vktREAACYA1v2U> at /articles
-- webcomic <https://wellcomecollection.org/articles/Y86rshEAACcAI4g4> at /articles
-- book <https://wellcomecollection.org/books/Y7QWrREAAHC43oH0> at /books
-- series <https://wellcomecollection.org/series/WfpEBCoAACrdd43y> composed of articles
-- in pictures <https://wellcomecollection.org/articles/Y-oAIhAAACAAix32> at /articles
+```json
+GET /articles?query=foo&filters=bar&sort=baz
+{
+    "type": "ResultList",
+    "results": [
+        {
+            ... // as above
+        },
+        ...
+    ],
+    "pageSize": 100,
+    "totalResults": 1000,
+    "totalPages": 10,
+    "nextPage": "/articles?query=foo&filters=bar&sort=baz?page=2",
+}
+```
 
 #### Filters
 
@@ -49,12 +82,67 @@ Filter for type eg stories, webcomics
 
 ### `/exhibitions`
 
+#### Example response
+
+```json
+GET /exhibitions/{id}
+{
+    "id": "",
+    "format": "",
+    "title": "",
+    "status": "",
+    "instantiations": [
+        {
+            "start": "", // ISO 8601 date only (no time)
+            "end": "", // ISO 8601 date only (no time)
+            "place": "",
+            "guide": {
+                ... // reuse the DigitalLocation model
+            }
+        }
+    ],
+    "contributors": [ // behind ?includes=contributors
+        {
+            ... // reuse the Contributor model
+        },
+        ...
+    ],
+    "image": {
+        "thumbnail": {
+            ... // reuse the DigitalLocation model
+        },
+        "caption": ""
+    },
+    "seasons": [
+        "",
+        ...
+    ],
+}
+```
+
+```json
+GET /exhibitions?query=foo&filters=bar&sort=baz
+{
+    "type": "ResultList",
+    "results": [
+        {
+            ... // as above
+        },
+        ...
+    ],
+    "pageSize": 100,
+    "totalResults": 1000,
+    "totalPages": 10,
+    "nextPage": "/exhibitions?query=foo&filters=bar&sort=baz?page=2",
+}
+```
+
 #### Filter
 
 - date
 - status
-- location
-- contributor?
+- place
+- contributor
 - format (ie permanent, exhibition, installation)
 - season
 
@@ -65,13 +153,72 @@ Filter for type eg stories, webcomics
 
 ### `/events`
 
+#### Example response
+
+```json
+GET /events/{id}
+{
+    "id": "",
+    "format": "",
+    "title": "",
+    "status": "",
+    "isOnline": false,
+    "availableOnline": false,
+    "audience": "",
+    "instantiations": [
+        {
+            "start": "", // ISO 8601 datetime
+            "end": "", // ISO 8601 datetime
+            "place": ""
+        }
+    ],
+    "contributors": [ // behind ?includes=contributors
+        {
+            ... // reuse the Contributor model
+        },
+        ...
+    ],
+    "image": {
+        "thumbnail": {
+            ... // reuse the DigitalLocation model
+        },
+        "caption": ""
+    },
+    "series": [
+        "",
+        ...
+    ],
+    "seasons": [
+        "",
+        ...
+    ],
+}
+```
+
+```json
+GET /events?query=foo&filters=bar&sort=baz
+{
+    "type": "ResultList",
+    "results": [
+        {
+            ... // as above
+        },
+        ...
+    ],
+    "pageSize": 100,
+    "totalResults": 1000,
+    "totalPages": 10,
+    "nextPage": "/events?query=foo&filters=bar&sort=baz?page=2",
+}
+```
+
 #### Filter
 
 - date
 - accessibility status (is it audio described, signed, etc)
 - location
 - format (eg discussion)
-- event-series
+- series
 - season
 - audience (eg schools)
 - contributor
@@ -83,23 +230,15 @@ Filter for type eg stories, webcomics
 - date
 - relevance
 
-### `/books`
-
-?????????
-
 ## Open questions
 
 - How do we represent functional content?
-- Should we include the full prismic API response at eg `/articles/{id}`?
-    - should we be able to list all of the docs of a given type at once? what should that url look like?
 - If we want to avoid maintaining multiple versions of the same code, should this work be done as an extension to the concepts API?
-- Works exist at [api.wellcomecollection.org/catalogue/v2/works](api.wellcomecollection.org/catalogue/v2/works), concepts at [/catalogue/v2/concepts](api.wellcomecollection.org/catalogue/v2/concepts), images at [/catalogue/v2/images](api.wellcomecollection.org/catalogue/v2/images).  
+- Works exist at [api.wellcomecollection.org/catalogue/v2/works](api.wellcomecollection.org/catalogue/v2/works), concepts at [/catalogue/v2/concepts](api.wellcomecollection.org/catalogue/v2/concepts), images at [/catalogue/v2/images](api.wellcomecollection.org/catalogue/v2/images).
 Stories etc aren't part of the catalogue - should the URL be different?
 - what are the top level types? is it articles? stories? series?
 - where do books go? they appear on the stories landing page, so should they appear in stories search results? or are they at a separate endpoint?
 - what is a serial? what is a series?
-- is 'In pictures' the same as 'Image gallery'?
-- Should we be able to filter by contributor role? is that something for the first pass or something we address later? which contributor roles are there? yes for stories? no for exhibitions and events?
 - can we filter exhibitions for accessibility, ie whether a digital guide exists? we can for events, but unclear for exhibitions
 - do we have any exhibition contributors who have contributed to more than one? is a filter for exhibition contributors really going to be useful?
 
--- a/rfcs/055-prismic-api/README.md
+++ b/rfcs/055-prismic-api/README.md
@@ -21,6 +21,7 @@ We'd like to build a new corresponding API which will allow users to search for 
 - The new API service should be written in typescript, following patterns set by the [concepts API](../050-concepts-api/README.md) for filtering, pagination, error handling, etc.
 - We shouldn't create new IDs for exhibitions, events, stories, etc. There aren't any situations where we should need to merge content from multiple sources, so we should use the document IDs directly from prismic.
 - The elasticsearch index mapping should represent the contract between the pipeline and the API. The API shouldn't need to know anything about the structure of the data in prismic.
+- We won't include an endpoint for /series, nor will we include series nodes in the /articles endpoint. If an article is part of a series, we'll denormalise information from its parent onto the article itself to use for relevance and visual signposting in results.
 
 ## Proposed endpoints
 
@@ -48,7 +49,12 @@ GET /articles/{id}
             ... // reuse the DigitalLocation model
         },
         "caption": ""
-    }
+    },
+    "partOf": [
+        {
+            ... // reuse the partOf model for series
+        }
+    ]
 }
 ```
 
@@ -185,14 +191,11 @@ GET /events/{id}
         },
         "caption": ""
     },
-    "series": [
-        "",
-        ...
-    ],
-    "seasons": [
-        "",
-        ...
-    ],
+    "partOf": [
+        {
+            ... // reuse the partOf model for series and seasons
+        }
+    ]
 }
 ```
 
@@ -237,11 +240,6 @@ GET /events?query=foo&filters=bar&sort=baz
 - If we want to avoid maintaining multiple versions of the same code, should this work be done as an extension to the concepts API?
 - Works exist at [api.wellcomecollection.org/catalogue/v2/works](api.wellcomecollection.org/catalogue/v2/works), concepts at [/catalogue/v2/concepts](api.wellcomecollection.org/catalogue/v2/concepts), images at [/catalogue/v2/images](api.wellcomecollection.org/catalogue/v2/images).
 Stories etc aren't part of the catalogue - should the URL be different? Does `/content` work?
-- what are the top level types? is it articles? stories? series?
-- where do books go? they appear on the stories landing page, so should they appear in stories search results? or are they at a separate endpoint?
-- what is a serial? what is a series?
-- can we filter exhibitions for accessibility, ie whether a digital guide exists? we can for events, but unclear for exhibitions
-- do we have any exhibition contributors who have contributed to more than one? is a filter for exhibition contributors really going to be useful?
 
 ## Next steps
 
--- a/rfcs/055-prismic-api/README.md
+++ b/rfcs/055-prismic-api/README.md
@@ -9,13 +9,13 @@ We use [Prismic](https://prismic.io/) to edit and store information about our ex
 Recently, we've allowed users to [search for stories on wellcomecollection.org](https://wellcomecollection.org/search/stories) using [Prismic's GraphQL API](https://Prismic.io/docs/graphql).
 
 That MVP implementation has demonstrated that Prismic's search functionality isn't good enough to produce relevant results on its own.  
-See slack threads [about Prismic's GraphQL search](https://wellcome.slack.com/archives/C02ANCYL90E/p1666103130854219) and [attempts to augment Prismic's search results with third party libraries](https://wellcome.slack.com/archives/C3TQSF63C/p1667495781964079).
+See Slack threads [about Prismic's GraphQL search](https://wellcome.slack.com/archives/C02ANCYL90E/p1666103130854219) and [attempts to augment Prismic's search results with third party libraries](https://wellcome.slack.com/archives/C3TQSF63C/p1667495781964079).
 
-Primsic matches documents to a user's search terms using a very loose, fuzzy query on all text-like fields, but, unlike elasticsearch, does not assign each document a score corresponding to its relevance. Instead of sorting by relevance, users are limited to sorting the retrieved documents by date or by title, which often makes the results appear irrelevant (eg weak matches appearing at the top of the list due to recency). Primsic's GraphQL API is also unsuitable for filtering content by arbitrary fields, which further limits our users' ability to find the content they're looking for.
+Prismic matches documents to a user's search terms using a very loose, fuzzy query on all text-like fields, but, unlike Elasticsearch, does not assign each document a score corresponding to its relevance. Instead of sorting by relevance, users are limited to sorting the retrieved documents by date or by title, which often makes the results appear irrelevant (e.g. weak matches appearing at the top of the list due to recency). Prismic's GraphQL API is also unsuitable for filtering content by arbitrary fields, which further limits our users' ability to find the content they're looking for.
 
 We'd like to replace our queries to the Prismic API with something more configurable, like the system we have for the catalogue.
 
-We're building a pipeline which will ingest content from Prismic into a set of elasticsearch indices (see [RFC 056: Prismic to Elasticsearch ETL pipeline](https://github.com/wellcomecollection/docs/blob/rfc-056-prismic-etl/rfcs/056-prismic-etl-pipeline/README.md)). To allow users to search and filter  that content from Prismic, we also need a new set of API endpoints which will query those elasticsearch indices.
+We're building a pipeline which will ingest content from Prismic into a set of Elasticsearch indices (see [RFC 056: Prismic to Elasticsearch ETL pipeline](https://github.com/wellcomecollection/docs/blob/rfc-056-prismic-etl/rfcs/056-prismic-etl-pipeline/README.md)). To allow users to search and filter that content from Prismic, we also need a new set of API endpoints which will query those Elasticsearch indices. The priority purpose of these endpoints will be to serve our Search. We might use them at a later time for content list pages, but at this time the focus will solely be on making this useful for Search.
 
 This API will live at `https://api.wellcomecollection.org/content/v0/`, with endpoints for `/articles`, `/exhibitions`, and `/events`.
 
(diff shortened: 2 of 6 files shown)


rfcs/054-prismic-api/README.md
```diff
@@ -0,0 +1,48 @@
+# RFC 054: Prismic API
```
jamieparkinson, line 1: needs to be 055!

rfcs/054-prismic-api/README.md
```diff
@@ -0,0 +1,110 @@
...
+- If we want to avoid maintaining multiple versions of the same code, should this work be done as an extension to the concepts API?
+- Works exist at [api.wellcomecollection.org/catalogue/v2/works](api.wellcomecollection.org/catalogue/v2/works), concepts at [/catalogue/v2/concepts](api.wellcomecollection.org/catalogue/v2/concepts), images at [/catalogue/v2/images](api.wellcomecollection.org/catalogue/v2/images).  
+Stories etc aren't part of the catalogue - should the URL be different?
+- what are the top level types? is it articles? stories? series?
+- where do books go? they appear on the stories landing page, so should they appear in stories search results? or are they at a separate endpoint?
+- what is a serial? what is a series?
+- is 'In pictures' the same as 'Image gallery'?
+- Should we be able to filter by contributor role? is that something for the first pass or something we address later? which contributor roles are there? yes for stories? no for exhibitions and events?
+- can we filter exhibitions for accessibility, ie whether a digital guide exists? we can for events, but unclear for exhibitions
```
rcantin-w, line 103: @rcantin-w to investigate

rfcs/054-prismic-api/README.md
```diff
@@ -0,0 +1,110 @@
...
+- do we have any exhibition contributors who have contributed to more than one? is a filter for exhibition contributors really going to be useful?
```
rcantin-w, line 104: @harrisonpim to investigate

rfcs/054-prismic-api/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+### `/books`
+
+?????????
+
+## Open questions
+
+- How do we represent functional content?
+- Should we include the full prismic API response at eg `/articles/{id}`?
+    - should we be able to list all of the docs of a given type at once? what should that url look like?
```
rcantin-w, line 94: @alexwlchan @jamieparkinson @paul-butcher @jtweed we wanted to run these ones by you as it felt more of a group decision/technical. Fire away opinions :)

rfcs/054-prismic-api/README.md
```diff
@@ -0,0 +1,110 @@
...
+- If we want to avoid maintaining multiple versions of the same code, should this work be done as an extension to the concepts API?
```
rcantin-w, line 95: @alexwlchan @jamieparkinson @paul-butcher @jtweed same here

rfcs/054-prismic-api/README.md
```diff
@@ -0,0 +1,110 @@
...
+- Works exist at [api.wellcomecollection.org/catalogue/v2/works](api.wellcomecollection.org/catalogue/v2/works), concepts at [/catalogue/v2/concepts](api.wellcomecollection.org/catalogue/v2/concepts), images at [/catalogue/v2/images](api.wellcomecollection.org/catalogue/v2/images).  
+Stories etc aren't part of the catalogue - should the URL be different?
```
rcantin-w, line 97: @jtweed one for you

rfcs/054-prismic-api/README.md
```diff
@@ -0,0 +1,110 @@
...
+- what are the top level types? is it articles? stories? series?
+- where do books go? they appear on the stories landing page, so should they appear in stories search results? or are they at a separate endpoint?
```
rcantin-w, line 99: @jtweed @dannybirchall @jennpb wanted your thoughts on this one. Their url is under `/books` where all the other ones are under `/articles`, so we'd need to decide if this one strays from the URL convention.

rfcs/054-prismic-api/README.md
```diff
@@ -0,0 +1,110 @@
...
+- what is a serial? what is a series?
```
rcantin-w, line 100: @dannybirchall @davidpmccormick @gestchild @jtweed 

We'd like to get this one clarified, I think there was some legacy naming in the codebase here and we'd like to be using the correct language from the start.

rfcs/054-prismic-api/README.md
```diff
@@ -0,0 +1,110 @@
...
+- is 'In pictures' the same as 'Image gallery'?
```
rcantin-w, line 101: @dannybirchall @davidpmccormick @gestchild one for you?

rfcs/054-prismic-api/README.md
```diff
@@ -0,0 +1,110 @@
...
+- Should we be able to filter by contributor role? is that something for the first pass or something we address later? which contributor roles are there? yes for stories? no for exhibitions and events?
```
rcantin-w, line 102: @jtweed one for you. Considering authors & artists (which contain sub-categories such as photographers, illustrators, etc.), maybe there are others. 

We think it might make more sense for stories, which will often have more than one result per author, versus exh./events.

rfcs/054-prismic-api/README.md
alexwlchan: I would think `/articles`?

rfcs/054-prismic-api/README.md
rcantin-w: There is also "Photo Story"

rfcs/054-prismic-api/README.md
alexwlchan: I think we should definitely be sharing code between this new API and the concepts API, and we've discussed migrating the existing works/images APIs into TypeScript and that platform also.

rfcs/054-prismic-api/README.md
alexwlchan: If we allow filtering by contribution role in Prismic, do we also allow it for works?

rfcs/054-prismic-api/README.md
gestchild: I believe a serial is a particular type of series, i.e. one with a schedule which is a list of titles it contains and their corresponding publish date

rfcs/054-prismic-api/README.md
gestchild: where are these from?

rcantin-w: Labels!
See here for example, I think we might've gotten the wrong one [in search](https://wellcomecollection.org/search/stories?query=blue&page=2) where the card displays "Image Gallery" and then the [detailed page](https://wellcomecollection.org/articles/Ybi9_REAACAAcOrA) says "In Pictures"

rfcs/054-prismic-api/README.md
rcantin-w: Just had a chat with @dannybirchall, that they are very much a different thing from articles, and a book endpoint is not a required thing.

rfcs/054-prismic-api/README.md
gestchild: 'In Pictures' is one of the story formats and has been added to the format field of the story in the example you've posted. Not sure where 'Image Gallery' is coming from yet, just looking.
N.B. there isn't an 'Image Gallery' format set up in Prismic.

gestchild: > There is also "Photo Story"

This is another format that can be applied to stories:

Screenshot 2023-02-21 at 17 17 08

gestchild: Looks like 'Image Gallery' is being displayed on search results, because [it is mapped to the id of the 'In pictures' story format](https://github.com/wellcomecollection/wellcomecollection.org/blob/6500012429db624e2f9fc9392bc6a1be41b6806e/common/data/content-format-ids.ts#L2)

And that is the [source of the label in the search result](https://github.com/wellcomecollection/wellcomecollection.org/blob/6500012429db624e2f9fc9392bc6a1be41b6806e/catalogue/webapp/services/prismic/fetch/index.ts#L17)

Don't know why that is hard coded and not coming from Prismic?

gestchild: So yes, 'In pictures' and 'Image gallery' are the same thing and the label on the search result needs fixing

gestchild:
@rcantin-w @harrisonpim We've come up against some issues recently, outside of search, which this API could potentially resolve. Perhaps we could discuss tomorrow? cc @davidpmccormick

rfcs/054-prismic-api/README.md
rcantin-w: Thank you! Made a separate ticket for it: https://github.com/wellcomecollection/wellcomecollection.org/issues/9341

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+- relevance
+
+## Open questions
+
+- How do we represent functional content?
+- If we want to avoid maintaining multiple versions of the same code, should this work be done as an extension to the concepts API?
+- Works exist at [api.wellcomecollection.org/catalogue/v2/works](api.wellcomecollection.org/catalogue/v2/works), concepts at [/catalogue/v2/concepts](api.wellcomecollection.org/catalogue/v2/concepts), images at [/catalogue/v2/images](api.wellcomecollection.org/catalogue/v2/images).
+Stories etc aren't part of the catalogue - should the URL be different?
+- what are the top level types? is it articles? stories? series?
+- where do books go? they appear on the stories landing page, so should they appear in stories search results? or are they at a separate endpoint?
```
rcantin-w, line 240:
```suggestion
```

https://github.com/wellcomecollection/docs/pull/90#discussion_r1113347094 

We could also add it to considerations? "Books will not be considered as part of this RFC"

rfcs/054-prismic-api/README.md
rcantin-w: So I think we'll have 
`/articles`
`/articles/{id}` (even though that would rarely be used as "The content of the pages themselves should still be fetched from Prismic directly.", this ensures we follow "the new endpoints should seamlessly fit into the rest of the wellcomecollection.org API suite from a user POV. All of the existing conventions should be followed." )
and then we're wondering if we need to consider
`/articles/webcomics`

rfcs/054-prismic-api/README.md
rcantin-w: Right, so exhibition content types in Prismic do offer a BSL information and and Audio description information field, but after talking with Danny, I can confirm they are not being used (and not displayed when filled in; should we delete them?).

I know we'd thought about it, but as Exhibition Guides should be for every exhibition moving forward, he doesn't think it'd be a valuable filter. 

We might want to consider BSL tours and Dimmed lights times as more of events anyway, so adding an Accessibility filter to Exhibition doesn't seem necessary.

Loved the idea for events though 👍 

cc @harrisonpim

harrisonpim: Great, thanks @rcantin-w!

Even if accessibility info isn't a meaningful filter for exhibitions, do you think we should we aim to include a link to the digital guide in the API response? My instinct is no, given this goal

> The API should only return enough information for users to figure out whether a result is relevant to them

but I'm open to other arguments!

rcantin-w: The only reason we'd want that would be for a quick access, which I'm thinking we can determine later on if needed, based on the amount of people accessing it from the exhibition page itself?

rfcs/054-prismic-api/README.md
rcantin-w: We could also figure out later with data if people are trying to search for specific artists and take it from there?

harrisonpim: @alexwlchan made the point yesterday that filtering exhibitions and events by contributor is necessary if we want to include those results on a contributor's concept page. IMO that's more than enough reason to include it as a filter!

We should still clarify whether we want to be able to filter by specific contributor role though

rcantin-w: I absolutely agree, I meant the split by roles, sorry! 

So one filter for now and eventually, is this a nice one to have specific filters for roles?
I'm thinking especially of stories, where the author and the artist are different and people might want to know they're looking for the photographer/illustrator, etc.

rfcs/054-prismic-api/README.md
harrisonpim: We have results!

| Name                                              |   Count |
|:----------------------------------------------|----:|
| heart-n-soul-at-the-hub                       |   2 |
| ann-veronica-janssens                         |   1 |
| grace-ndiritu                                 |   1 |
| young-trans-people-from-gendered-intelligence |   1 |
| jason-barker                                  |   1 |
| dr-jana-funke                                 |   1 |
| maria-mckinney                                |   1 |
| professor-michael-doherty                     |   1 |
| professor-david-machugh                       |   1 |
| john-walter                                   |   1 |
| alison-rodger                                 |   1 |
| marion-coutts                                 |   1 |
| daria-martin                                  |   1 |
| dr-michael-banissy                            |   1 |
| jim-naughten                                  |   1 |
| martina-amati                                 |   1 |
| professor-kevin--fong                         |   1 |
| ranjit-kandalgaonkar                          |   1 |
| photobot.co                                   |   1 |
| heart-n-soul                                  |   1 |

Looks like Heart n Soul have contributed to multiple exhibitions, and given [this comment](https://github.com/wellcomecollection/docs/pull/90#discussion_r1114170705), i think it's worth including the filter!

rfcs/054-prismic-api/README.md
rcantin-w: That's right! Thank you.

rfcs/054-prismic-api/README.md
rcantin-w: After further chats, we're not going to do `/articles/[type]`, it'll be a filter instead.

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+- date
+- relevance
+
+## Open questions
+
+- How do we represent functional content?
+- If we want to avoid maintaining multiple versions of the same code, should this work be done as an extension to the concepts API?
+- Works exist at [api.wellcomecollection.org/catalogue/v2/works](api.wellcomecollection.org/catalogue/v2/works), concepts at [/catalogue/v2/concepts](api.wellcomecollection.org/catalogue/v2/concepts), images at [/catalogue/v2/images](api.wellcomecollection.org/catalogue/v2/images).
+Stories etc aren't part of the catalogue - should the URL be different? Does `/content` work?
+- Should we include a `/series` endpoint as part of this work?
```
harrisonpim, line 243: @davidpmccormick it would be great if you could re-state the problem behind the "Rearranging comics on the stories landing page" toggle in this thread!

It sounds like a `/series` endpoint is something we should aim to include, but creating a matching pair of `article` and `series` es indices could be a tricky bit of work for the prismic pipeline folks. I don't know whether it's something for the first pass, or something which should be left until after the basics have been implemented. Probably a question for @jtweed to answer? maybe @jamieparkinson has thoughts?

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+#### Example response
+
+```http
+GET /exhibitions/{id}
+{
+    "id": "",
+    "format": "",
+    "title": "",
+    "status": "",
+    "instantiations": [
```
jamieparkinson, line 101: showings? openings? not sure

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+    "results": [
+        {
+            ... // as above
+        },
+        ...
+    ],
+    "pageSize": 100,
+    "totalResults": 1000,
+    "totalPages": 10,
+    "nextPage": "/exhibitions?query=foo&filters=bar&sort=baz?page=2",
```
jamieparkinson, line 143:
```suggestion
    "nextPage": "/exhibitions?query=foo&filters=bar&sort=baz&page=2",
```

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+    "results": [
+        {
+            ... // as above
+        },
+        ...
+    ],
+    "pageSize": 100,
+    "totalResults": 1000,
+    "totalPages": 10,
+    "nextPage": "/articles?query=foo&filters=bar&sort=baz?page=2",
```
jamieparkinson, line 74:
```suggestion
    "nextPage": "/articles?query=foo&filters=bar&sort=baz&page=2",
```

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+    "results": [
+        {
+            ... // as above
+        },
+        ...
+    ],
+    "pageSize": 100,
+    "totalResults": 1000,
+    "totalPages": 10,
+    "nextPage": "/events?query=foo&filters=bar&sort=baz?page=2",
```
jamieparkinson, line 215:
```suggestion
    "nextPage": "/events?query=foo&filters=bar&sort=baz&page=2",
```

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+    "pageSize": 100,
+    "totalResults": 1000,
+    "totalPages": 10,
+    "nextPage": "/articles?query=foo&filters=bar&sort=baz?page=2",
+}
+```
+
+#### Filters
+
+- date
```
jamieparkinson, line 80: How would a date filter work in practice? Just filtering for whether 1 date was included or would we actually want a `fromDate` and a `toDate`?

rfcs/055-prismic-api/README.md
jamieparkinson: What is a `series`?

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+- The API's URL structure should also be consistent with what appears on the front end of the site. For example, if article on the site appears at `/articles/{id}`, the API equivalent should be at `/articles/{id}`. The same should be true for search results.
+- The new API service should be written in typescript, following patterns set by the [concepts API](../050-concepts-api/README.md) for filtering, pagination, error handling, etc.
+- We shouldn't create new IDs for exhibitions, events, stories, etc. There aren't any situations where we should need to merge content from multiple sources, so we should use the document IDs directly from Prismic.
+- The elasticsearch index mapping should represent the contract between the pipeline and the API. The API shouldn't need to know anything about the structure of the data in Prismic.
+- We won't include an endpoint for `/series`, nor will we include series nodes in the `/articles` endpoint. If an article is part of a series, we'll denormalise information from its parent onto the article itself to use for relevance and visual signposting in results.
+
+## Proposed endpoints
```
jamieparkinson, line 26: It looks like all the entities/endpoints have common fields:
- `id`
- `format`
- `title`
- `contributors`
- `image`

I wonder if this points towards a `type` field that might be `Article` or `Exhibition` or `Event` - like we use for concepts https://github.com/wellcomecollection/docs/tree/main/rfcs/050-concepts-api. On the other hand, is this subclassing of any use?

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+### `/articles`
+
+#### Example response
```
jamieparkinson, line 30: Can we get some more detail on the types of the values of these entities? Examples as we did in the concepts API RFC https://github.com/wellcomecollection/docs/tree/main/rfcs/050-concepts-api#endpoint-1-retrieve-a-single-concept might be a good way to achieve that

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+- status
+- place
+- contributor
+- format (ie permanent, exhibition, installation)
+- season
+
+#### Sort
+
+- date
+- relevance
```
jamieparkinson, line 159: Which is the default?

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+```http
+GET /events/{id}
+{
+    "id": "",
+    "format": "",
+    "title": "",
+    "status": "",
+    "isOnline": false,
+    "availableOnline": false,
```
jamieparkinson, line 173: What do these mean?

rfcs/055-prismic-api/README.md
jamieparkinson: What is a series?

rcantin-w: @dannybirchall might explain better but from I gathered, it's a grouping of articles/webcomics/other types. 
You'll find them under `/series/[id]` (https://wellcomecollection.org/series/WsSUrR8AAL3KGFPF) but there is no `/series` listing.
If they are scheduled (for every Thursday, for example), they are called a Serial, which is just a type of Series.

rfcs/055-prismic-api/README.md
rcantin-w: I'd see it the same as in `/works`, so from -> to.
Although here we might want to offer more than just years, having months and maybe days here might be more logical.
I know @gestchild has worked on a date picker in the past, is that something that we decided against in the end?

rfcs/055-prismic-api/README.md
rcantin-w: That's interesting, because without a query you'd think it would be date, but with one, relevance. I'm guessing the date would be highly considered here in terms of relevance @harrisonpim ?

rfcs/055-prismic-api/README.md
rcantin-w: From what I could gather; `isOnline` means the event takes place online
`availableOnline` means it's available for viewing afterwards (was recorded).

cc @gestchild @davidpmccormick for confirmation?

rfcs/055-prismic-api/README.md
rcantin-w: We don't have that behaviour elsewhere though, changing the sort order based on query parameters. I'm thinking it might have to be relevance by default? It is a search tool after all and `/whats-on` still exists for ordering by date?

cc @jtweed

rfcs/055-prismic-api/README.md
jamieparkinson: The works API uses `YYYY-MM-DD` fyi, the year resolution is just on the frontend

rfcs/055-prismic-api/README.md
jamieparkinson: I think there's also an argument for a blend ie scoring which considers recency as making content more relevant

harrisonpim: > I think there's also an argument for a blend ie scoring which considers recency as making content more relevant

yeah, all of the discussions we've had so far (and the tests i'm doing at the moment) have included date as secondary relevance feature by default. are there any situations where you wouldn't want to see events/exhibitions results which take the date into account?

jamieparkinson: Seems unlikely really, so long as the date doesn't obliterate other aspects of relevance!

rfcs/055-prismic-api/README.md
davidpmccormick: With the 'Rearranging comics on the stories landing page' [toggle](https://dash.wellcomecollection.org/toggles/) turned on we add cards linking to the three most recent comic series on the stories landing page

Screenshot 2023-02-23 at 11 31 29

This is the [most recent comic series page](https://wellcomecollection.org/series/Y8aThBEAAKREABhL) for example.

These series don't have knowledge of the articles that are contained in them (the link is specified on the articles themselves) so when we query for them we don't know if they contain any articles yet. The series have to be published in order for the articles to link to them, so there will always be a point (however brief) when they don't have any articles attached (this isn't something that can be done as an inline in Prismic when e.g. publishing the first article).

Article series that don't contain any articles will currently 404.

At present we're asking the content team to add a `delist` tag to the series until the first article is published, but 1. maybe `delist`ed things will still show up in search(?) and 2. it isn't great to put the onus on the content team to remember to do this.

rcantin-w: > 1. maybe delisted things will still show up in search(?)

I can confirm they do, we allow them to show up in search after confirming with the content team. They use `delist` to curate older content before rendering it in `/articles`, but believe they should still be searchable.

rfcs/054-prismic-api/README.md
```diff
@@ -0,0 +1,110 @@
...
+- date
+- relevance
+
+### `/books`
+
+?????????
+
+## Open questions
+
+- How do we represent functional content?
```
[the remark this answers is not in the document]
jtweed: Interesting question. I think it's worth bottoming it out a bit, same for exhibitions and events, to make sure that we don't build ourselves into a corner. I would assume it becomes another endpoints, `/pages`?

rfcs/054-prismic-api/README.md
jtweed: Code sharing is one thing, but this should be built as a separate service. Single responsibility principle and all that.

rfcs/054-prismic-api/README.md
jtweed: IMO this is a separate API that should live at `/content/v1`, so stories would be at `/content/v1/articles`.

jtweed:
> @rcantin-w @harrisonpim We've come up against some issues recently, outside of search (and may be relevant to search too), which this API could potentially resolve. Perhaps we could discuss tomorrow? cc @davidpmccormick

It would be good to understand what these are and whether it makes sense to do now. Whilst I don't see us using this API to render content pages, I do think that we may end up using it for section pages as well as for search. We need to prioritise its use for search, but if there are cases that we can knock off at the same time, it would be good to know about them now so that we can make sure we incorporate.

rfcs/055-prismic-api/README.md
jtweed: Agree that this is out of scope until we have search covered off. @jamieparkinson is also correct as to how to think about this - if we want a list of `x`, then we generate that by filtering `/content/v1/x`. This is consistent with how the catalogue api works and allow us to inherit pagination etc. Cf hierarchical works, ie archives and library series.

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+- How do we represent functional content?
+- If we want to avoid maintaining multiple versions of the same code, should this work be done as an extension to the concepts API?
+- Works exist at [api.wellcomecollection.org/catalogue/v2/works](api.wellcomecollection.org/catalogue/v2/works), concepts at [/catalogue/v2/concepts](api.wellcomecollection.org/catalogue/v2/concepts), images at [/catalogue/v2/images](api.wellcomecollection.org/catalogue/v2/images).
+Stories etc aren't part of the catalogue - should the URL be different? Does `/content` work?
+- Should we include a `/series` endpoint as part of this work?
+
+## Next steps
+
+- Answer most of those questions ☝️
```
rcantin-w, line 247: > > @rcantin-w @harrisonpim We've come up against some issues recently, outside of search (and may be relevant to search too), which this API could potentially resolve. Perhaps we could discuss tomorrow? cc @davidpmccormick
> 
> It would be good to understand what these are and whether it makes sense to do now. Whilst I don't see us using this API to render content pages, I do think that we may end up using it for section pages as well as for search. We need to prioritise its use for search, but if there are cases that we can knock off at the same time, it would be good to know about them now so that we can make sure we incorporate.

I never know how to deal with comments that aren't threadable, so putting it here as I know we can have a thread. It was about two things:

1) https://github.com/wellcomecollection/docs/pull/90#discussion_r1115559913
2) An issue with the sorting order of Stories, where older stories coming from the old platform would never have the correct publishing date as they were published in Prismic later on. Therefore there's a field for a custom date that's used in the UI (`published_date`(?)) but I believe cannot be used for Sorting. They would like to have that considered as part of the API and maybe take advantage of it (@gestchild please correct me if I'm wrong).

rfcs/055-prismic-api/README.md
jennpb: I don't know what `showings, `openings` or`instantiations` means here or how that relates to exhibitions. I'm probably missing context here. Is this to tie in somehow with touring exhibitions?

rfcs/055-prismic-api/README.md
jtweed: There should be a `type` here in any case, for consistency with the catalogue api. I do think it's useful as potentially we will have endpoints with mixed types, like we do in a couple of cases in the catalogue.

rfcs/055-prismic-api/README.md
jtweed: Agreed, it would be good to see a couple of examples using real data here.

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+#### Example response
+
+```http
+GET /articles/{id}
+{
+    "id": "",
+    "format": "",
+    "title": "",
+    "body": "", // not exposed by default, should require an ?includes=body
```
jtweed, line 38: Do we need this? Not sure we do? If it is there, what format is it in?

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+### `/articles`
+
+#### Example response
+
+```http
+GET /articles/{id}
+{
+    "id": "",
+    "format": "",
```
jtweed, line 36: Is this a controlled list, like work formats? If so, we need to see that represented as an object here with its own type.

rfcs/055-prismic-api/README.md
jtweed: I think all date filters should be ranges.

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+    "nextPage": "/articles?query=foo&filters=bar&sort=baz&page=2",
+}
+```
+
+#### Filters
+
+- date
+- contributor
+- series
+- contentType
```
jtweed, line 83: What is this?

rfcs/055-prismic-api/README.md
jtweed: What is it called in the Prismic model? That should be our starting point here and we should have a _very_ good reason for not using the same names for things.

rfcs/055-prismic-api/README.md
rcantin-w: I'm tempted to say no _for the UI_, in `transformPrismicResponse` there is no mention of it.

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+    "format": "",
+    "title": "",
+    "status": "",
+    "instantiations": [
+        {
+            "start": "", // ISO 8601 date only (no time)
+            "end": "", // ISO 8601 date only (no time)
+            "place": "",
+            "guide": {
+                ... // reuse the DigitalLocation model
```
jtweed, line 107: As above, I'm not sure we do reuse things like that here. We should be following how things are modelled in Prismic for the most part

rfcs/055-prismic-api/README.md
rcantin-w: I think it's meant to be changed to `format` (webcomic, for example) after a chat we had yesterday, @harrisonpim right?

rfcs/055-prismic-api/README.md
jtweed: Elsewhere the default order is undefined. I think there is an argument for the default order to be "useful" for listings etc, but I don't think it make sense to worry about that until and if we are using these endpoints to generate section pages. Even at that point there are arguments that it is better to be specific, as the order required might be different on different pages or we might want to AB test different algorithmic orders. So I'd go with undefined as the easiest option, then relevance if it's a full text query.

rcantin-w: Just to check what we mean by undefined? The current stories search is currently ordered by latest published by default but we remove it from the URL query as it's the default order. Same for works and "relevance".

rfcs/055-prismic-api/README.md
jtweed: What are they in Prismic? I'm assuming it helps to render links to events without having to look at all that, much like the availabilities on a work? I can also see us wanting to filter events in a simple way like this.

jtweed: Another example of why we need to see some real examples here.

rfcs/055-prismic-api/README.md
jtweed: 2 is interesting and the sort of problem we could eventually solve with this. I would think that we should use the content pipeline to identify these articles and then correct the date to keep things like filtering consistent.

rfcs/055-prismic-api/README.md
rcantin-w: Ah sorry if we meant straight from the API, I was still under the impression it was relevance by default?

rfcs/055-prismic-api/README.md
jamieparkinson: Argument would be that if our query includes the article body then for explainability the API should expose it

rfcs/055-prismic-api/README.md
jamieparkinson: I agree for this field, but for images (which also have licensing information) I think it seems quite sensible to use DigitalLocations. Was briefly discussed and I know @alexwlchan agrees, but definitely up for debate

rfcs/054-prismic-api/README.md
jamieparkinson: I spent a while wrestling with this thought yesterday (thought I'd written a comment but can't find it) - I very much agree with it living at `/content` (probably `v0` at first, though), but I think it's worth noting that having a `/content/v0/articles|events|exhibitions` endpoint does preclude us from having a [mixed-type list](https://github.com/wellcomecollection/docs/pull/90#discussion_r1115861763) at the root (`/content/v0/`). Not a problem, but I found it useful to think about, anyway.

rfcs/055-prismic-api/README.md
alexwlchan: **General comment about events:**

I think the way events are modelled in Prismic is somewhat sub-optimal and not something we should necessarily seek to emulate; I know Jenn is thinking about this also.

see e.g. https://github.com/wellcomecollection/wellcomecollection.org/issues/8902 and https://github.com/wellcomecollection/wellcomecollection.org/issues/8830 for some thorny examples

I agree with sticking to Prismic in general, but this might be a case where we want to make changes with a view to backporting those back into Prismic later.

Before you can model this in the API, I think we need to think about what behaviour we want out of the front-end. e.g. the "Lights Up" session in those tickets that runs on three different occasions but has a single summary page on the website. How does that show up when I use the website search? How does it interact with filters? Does the appearance of the search result change based on my filters? etc.

rfcs/055-prismic-api/README.md
jennpb: I could imagine wanting to filter content that is of or has a certain type, what we currently call `interpretation` in events, e.g. `BSL video` `Audio description`. Is that what you mean by "what sorts of guides are available" @alexwlchan ?

alexwlchan: > Is that what you mean by "what sorts of guides are available"?

yes, exactly

dannybirchall: hmm, interpretation in events is generally provided live; the guides are static digital content.

rfcs/055-prismic-api/README.md
rcantin-w:
```suggestion
- format
```

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+    "totalPages": 10,
+    "nextPage": "/articles?query=foo&filters=bar&sort=baz?page=2",
+}
+```
+
+#### Filters
+
+- date
+- contributor
+- series
```
rcantin-w, line 82: Not for v0?

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,249 @@
...
+#### Example response
+
+```http
+GET /articles/{id}
+{
+    "id": "",
+    "format": "",
+    "title": "",
+    "body": "", // not exposed by default, should require an ?includes=body
+    "standfirst": "",
```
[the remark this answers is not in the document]
harrisonpim: [answer confirmed in slack](https://wellcome.slack.com/archives/C3N7J05TK/p1677149280740959?thread_ts=1677145404.623039&cid=C3N7J05TK) - we won't use the standfirst in the search results, so we don't need to include it in the api response either

rfcs/055-prismic-api/README.md
rcantin-w: So one for @jamieparkinson's ETL side of considerations, and resolved on here.

rfcs/055-prismic-api/exhibitions.md
```diff
@@ -0,0 +1,101 @@
...
+  DD-MM-YYYY
+- **instantiations.end.to**
+  DD-MM-YYYY
+- **place.label**
+  List of physical locations, would also include "Online".
+- **contributor.agent.label**
+  e.g. Filmmaker, Curator
+- **format.label**
+  [`Permanent Exhibition`, `Season`, `Installation`]
+- **season**
```
rcantin-w, line 88:
```suggestion
```

As there are only two right now, I don't think we want this as a filter.

rfcs/055-prismic-api/articles.md
```diff
@@ -0,0 +1,74 @@
...
+        ...
+    ],
+    "pageSize": 100,
+    "totalResults": 1000,
+    "totalPages": 10,
+    "nextPage": "/articles?query=foo&filters=bar&sort=baz&sortOrder=asc&page=2",
+}
+```
+
+## Filters
```
rcantin-w, line 55: If useful

```
- **publicationDate.from**
  DD-MM-YYYY
- **publicationDate.to**
  DD-MM-YYYY
- **contributor.agent.label**
  Includes author, artists, photographers&hellip;
- **partOf**
  List of series (e.g. serials - which is a scheduled list of articles, or webcomic series)
- **format.label**
  [`Comic`, `Long read`, `Prose poem`, `Podcast`, `In pictures`, `Article`, `Photo story`, `Interview`, `Book extract`]
  ```

File: rfcs/054-prismic-api/README.md (hunk 22 above)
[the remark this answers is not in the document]
Comment (rcantin-w) [reply rfcs/054-prismic-api/README.md]: We've put this in the `v1` list, closing this thread.

File: rfcs/055-prismic-api/README.md (hunk 20 above)
Comment (rcantin-w) [reply rfcs/055-prismic-api/README.md]: @harrisonpim closing this after our conversation re sorting - feel free to reopen but I think you've taken it in for the newer versions of the files.

File: rfcs/055-prismic-api/README.md (hunk 27 above)
Comment (rcantin-w) [reply rfcs/055-prismic-api/README.md]: After discussing with @dannybirchall we're going with interpretations, but not the static digital content which might go into v1. 

Is there anywhere I could find example of DigitalLocations to further my understanding?

File: rfcs/055-prismic-api/README.md (hunk 18 above)
Comment (rcantin-w) [reply rfcs/055-prismic-api/README.md:46]: `type: "Article" | "Event" | "Exhibition"` has been added to the example responses

File: rfcs/055-prismic-api/README.md (hunk 12 above)
Comment (davidpmccormick) [reply rfcs/055-prismic-api/README.md]: Wondering about the `delist` issue and I’m thinking even if I go with a different approach that doesn’t involve requiring the content team to add that tag, the comic series will necessarily still have to be published before any comics are attached to it, so it will (presumably) still show up in search results?

File: rfcs/055-prismic-api/articles.md [hunk 32]
```diff
@@ -0,0 +1,114 @@
+# Prismic API: articles endpoint
```
jtweed, line 1: Content API

rfcs/055-prismic-api/articles.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Example response
+
+```
+GET /articles/{id}
+{
+   {
```
jtweed, line 8: Spurious `{`

rfcs/055-prismic-api/articles.md
```diff
@@ -0,0 +1,114 @@
...
+    "type": "Article",
+    "id": "XAEEIhQAACsA4eEl",
+    "format": {
+        "type": "Format",
```
jtweed, line 12: Is this `Format` or `StoryFormat`?

rfcs/055-prismic-api/articles.md
```diff
@@ -0,0 +1,114 @@
...
+        "id": "W7TfJRAAAJ1D0eLK",
+        "name": "article"
```
jtweed, line 14: What is this called in the Prismic API? Let's not confuse labels and human readable slugs here.

rfcs/055-prismic-api/articles.md
```diff
@@ -0,0 +1,114 @@
...
+    },
+    "title": "Sick of being lonely",
+    "publicationDate": "2018-12-20T10:44:15+0000",
+    "caption": "When his relationship ended, Thom James first withdrew from the world, then began to suffer from illnesses with no apparent physical cause.",
+    "contributors": [
```
jtweed, line 19: Has this been cribbed from the Catalogue API or is it modelled like this in Prismic?

rfcs/055-prismic-api/articles.md
```diff
@@ -0,0 +1,114 @@
...
+
+- relevance
+- publicationDate
+
+Default sort should be by relevance, with a fallback to id if no query is provided or where documents have the same score.
+
+## Aggregations
+
+- contributor.agent.label
+- format.label
```
jtweed, line 114: Seems wrong to me. We'll want to restrict by ID in the API, not label. Which is called `name` above in any case (assuming that's correct).

rfcs/055-prismic-api/events.md
```diff
@@ -0,0 +1,108 @@
+# Prismic API: Events endpoint
```
jtweed, line 1: Content API

rfcs/055-prismic-api/events.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Example response
+
+```
+GET /events/{id}
+{
+    "type": "Event",
+    "id": "Y8aBHhEAAKRE_9AF",
+    "format": {
```
jtweed, line 10: Make sure this is consistent with ArticleFormat.

rfcs/055-prismic-api/events.md
```diff
@@ -0,0 +1,108 @@
...
+        ...
+    ],
+    "instantiations": [
+        {
+            "start": "2018-12-20T10:44:15+0000",
+            "end": "2018-12-20T16:44:15+0000",
+            "place": {
+                "id": "Wn1fvyoAACgAH_yG",
+                "label": "Reading Room",
+                "type": "Place"
```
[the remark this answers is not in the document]
jtweed: Do we want to reuse the "catalogue" `Place` here? Is it really the same? I think the wider `type` question is not one for a PR comment, but will need a possibly in person chat to explain the wider context and why the APIs are like that. (Spoiler: the reason was JSON-LD, but I think it's a pattern that should stay).

rfcs/055-prismic-api/events.md
```diff
@@ -0,0 +1,108 @@
...
+            },
+        }
+    ],
+    "contributors": [ // behind ?includes=contributors
+        {
+            ... // reuse the Contributor model (without primaryContributor)
```
jtweed, line 49: Are we sure about that? I can see arguments either way, but I'm not sure it's where I would start.

rfcs/055-prismic-api/exhibitions.md
```diff
@@ -0,0 +1,86 @@
+# Prismic API: exhibitions endpoint
```
jtweed, line 1: All the same comments as for events.

jtweed (changes requested):

rfcs/055-prismic-api/articles.md
harrisonpim: should be an `ArticleFormat`, right?

rfcs/055-prismic-api/articles.md
harrisonpim: tbh I originally thought the same thing, but [the catalogue api uses contributor.agent.label](https://api.wellcomecollection.org/catalogue/v2/works?contributor.agent.label=Columbario,%20Christopher.)... Happy to switch to ID if you think that makes more sense though?

rfcs/055-prismic-api/articles.md
harrisonpim: you're right, this should be an `Article`

rfcs/055-prismic-api/articles.md
harrisonpim: it's a minimal version of both really, but closer to how things look in the catalogue API. In the raw prismic response we get:

```json
{
    "contributors": [
        {
            "role": {
                "id": "Wux6DyIAAO5n3lzk",
                "type": "editorial-contributor-roles",
                "tags": [],
                "lang": "en-gb",
                "slug": "artist",
                "first_publication_date": "2018-05-04T15:19:48+0000",
                "last_publication_date": "2019-08-14T14:31:04+0000",
                "link_type": "Document",
                "isBroken": false
            },
            "contributor": {
                "id": "Y8aT1REAAKREABkW",
                "type": "people",
                "tags": [],
                "lang": "en-gb",
                "slug": "alex-brenchley",
                "first_publication_date": "2023-01-17T12:27:22+0000",
                "last_publication_date": "2023-01-25T11:54:21+0000",
                "link_type": "Document",
                "isBroken": false
            },
            "description": []
        }
    ]
}
```

By making a couple of requests in the pipeline for those individual `people` and `editorial-contributor-roles`, i think we can tidy that up into something like

```json
{
    "contributors": [
        {
            "agent": {
                "id": "Y8aT1REAAKREABkW",
                "label": "Alex Brenchley",
                "type": "Person"
            },
            "roles": [
                {
                    "id": "Wux6DyIAAO5n3lzk",
                    "label": "Artist",
                    "type": "EditorialContributorRole"
                }
            ],
            "type": "Contributor"
        }
    ]
}
```

What do you think?

rfcs/055-prismic-api/events.md
harrisonpim: [see the comment above](https://github.com/wellcomecollection/docs/pull/90#discussion_r1121561757)

rfcs/055-prismic-api/articles.md
harrisonpim: - single role instead of an array
- `contributor` instead of `agent`

rfcs/055-prismic-api/articles.md
harrisonpim: use `id` instead

rfcs/055-prismic-api/README.md
rcantin-w: Resolving thread as it was addressed in [David's PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/9362)

rfcs/055-prismic-api/events.md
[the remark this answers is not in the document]
rcantin-w: It's a "Place" content type in Prismic, so it's different from the Catalogue's. If there is potential for confusion, we could always think up a different name (`Location`?), but we were also wanting to stick with the Prismic model.

Screenshot 2023-03-02 at 10 28 56

rfcs/055-prismic-api/future-versions.md
```diff
@@ -0,0 +1,86 @@
...
+- **format.label**
+  [`Session`, `Game`, `Installation`, `Discussion`, `Performance`, `Workshop`, `Chill out`, `Shopping`, `Festival`, `Screening`, `SEND workshop`, `Late`, `Symposium`, `Gallery tour`, `Seminar`, `Study day`, `Walking tour`]
+- **partOf**
+  Part of a series of events (I'm not sure this is possible?)
+- **audience**
+  The public this is geared towards, e.g. Schools
+- **contributor.agent.id**
+  e.g. Facilitator, Host
+- **availableOnline**
+  Was recorded and the video is made available for a rewatch online.
```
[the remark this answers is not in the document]
harrisonpim: i've added a note about this in `future-versions.md`!

rfcs/055-prismic-api/future-versions.md
```diff
@@ -0,0 +1,86 @@
...
+### Filter
+
+- **instantiations.start.from**
+  DD-MM-YYYY
+- **instantiations.start.to**
+  DD-MM-YYYY
+- **instantiations.end.from**
+  DD-MM-YYYY
+- **instantiations.end.to**
+  DD-MM-YYYY
```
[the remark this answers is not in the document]
harrisonpim: done!

rfcs/055-prismic-api/future-versions.md
```diff
@@ -0,0 +1,86 @@
...
+- **interpretation**
+  Interpretations are useful accessibility tools for event searching. They are, for example: Captioned, BSL, Wheelchair friendly
+- **place.label**
+  List of physical locations, would also include "Online".
+- **format.label**
+  [`Session`, `Game`, `Installation`, `Discussion`, `Performance`, `Workshop`, `Chill out`, `Shopping`, `Festival`, `Screening`, `SEND workshop`, `Late`, `Symposium`, `Gallery tour`, `Seminar`, `Study day`, `Walking tour`]
+- **partOf**
+  Part of a series of events (I'm not sure this is possible?)
```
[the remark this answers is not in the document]
harrisonpim: afaik it's a bit of both... there are workarounds with the current data, but we'd like to change how things are structured in prismic anyway. either way, not something we need to worry about for v0 😁

rfcs/055-prismic-api/articles.md
```diff
@@ -0,0 +1,109 @@
...
+}
+```
+
+## Filters
+
+- **publicationDate.from**
+  DD-MM-YYYY
+- **publicationDate.to**
+  DD-MM-YYYY
+- **contributor.agent.id**
```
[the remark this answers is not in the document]
harrisonpim: good point! i've made that change

rfcs/055-prismic-api/articles.md
```diff
@@ -0,0 +1,109 @@
...
+            "type": "Contributor",
+        }
+    ],
+    "image": {
+        "type": "PrismicImage",
+        "dimensions": {
+            "width": 4000,
+            "height": 2670
+        },
+        "alt": "Photograph of a gallery installation showing a bench with yellow cushions facing a large video projection. The projection shows the inside of a McDonalds restaurant looking down from a high viewpoint in which the restaurant floor is beginning to flood with water.',"copyright": "Flooded McDonalds by Superflex | | Wellcome Collection | | | |","url": "https: //images.prismic.io/wellcomecollection/1c4a250965a3f9bd17e636f5fc008b88e1e4c649_ep_000832_027.jpg?auto=compress,format",
```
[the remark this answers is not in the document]
harrisonpim: there were a couple of problems with the json examples in all of the docs - they're all a bit tidier now 😁

rfcs/055-prismic-api/articles.md
```diff
@@ -0,0 +1,109 @@
...
+
+```
+GET /articles/{id}
+{
+    "type": "Article",
+    "id": "XAEEIhQAACsA4eEl",
+    "format": {
+        "type": "Format",
+        "id": "W7TfJRAAAJ1D0eLK",
+        "name": "Article"
```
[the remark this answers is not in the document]
harrisonpim: good point! done!

rfcs/055-prismic-api/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+We'd like to replace our queries to the Prismic API with something more configurable, like the system we have for the catalogue.
+
+We're building a pipeline which will ingest content from Prismic into a set of Elasticsearch indices (see [RFC 056: Prismic to Elasticsearch ETL pipeline](https://github.com/wellcomecollection/docs/blob/rfc-056-prismic-etl/rfcs/056-prismic-etl-pipeline/README.md)). To allow users to search and filter that content from Prismic, we also need a new set of API endpoints which will query those Elasticsearch indices. The priority purpose of these endpoints will be to serve our Search. We might use them at a later time for content list pages, but at this time the focus will solely be on making this useful for Search.
+
+This API will live at `https://api.wellcomecollection.org/content/v0/`, with endpoints for `/articles`, `/exhibitions`, and `/events`.
+
+We won't consider the way that documents are scored as part of this RFC. Relevance requirements should be developed iteratively and independently from the development of the API.
```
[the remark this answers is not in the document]
harrisonpim: really glad to hear it! thanks for all the ideas/guidance along the way

rfcs/055-prismic-api/articles.md
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+## Filters
+
+- **publicationDate.from**
+  YYYY-MM-DD
+- **publicationDate.to**
+  YYYY-MM-DD
+- **contributor.agent**
```
jamieparkinson, line 96: There is no `contributor.agent` field in the model given in the example above - as per [the faceting principles](https://github.com/wellcomecollection/docs/tree/main/rfcs/037-api-faceting-principles), I think this field should be `contributors.contributor` instead?

rfcs/055-prismic-api/articles.md
```diff
@@ -0,0 +1,111 @@
...
+## Sort options
+
+- relevance
+- publicationDate
+
+Default sort should be by relevance, with a fallback to id if no query is provided or where documents have the same score.
+
+## Aggregations
+
+- contributor.agent
```
jamieparkinson, line 110: Ditto

rfcs/055-prismic-api/future-versions.md
```diff
@@ -0,0 +1,86 @@
...
+  Interpretations are useful accessibility tools for event searching. They are, for example: Captioned, BSL, Wheelchair friendly
+- **place.label**
+  List of physical locations, would also include "Online".
+- **format.label**
+  [`Session`, `Game`, `Installation`, `Discussion`, `Performance`, `Workshop`, `Chill out`, `Shopping`, `Festival`, `Screening`, `SEND workshop`, `Late`, `Symposium`, `Gallery tour`, `Seminar`, `Study day`, `Walking tour`]
+- **partOf**
+  Part of a series or season of events
+- **audience**
+  The public this is geared towards, e.g. Schools
+- **contributor.agent**
```
jamieparkinson, line 36: Ditto

rfcs/055-prismic-api/future-versions.md
```diff
@@ -0,0 +1,86 @@
...
+  YYYY-MM-DD
+- **instantiations.start.to**
+  YYYY-MM-DD
+- **instantiations.end.from**
+  YYYY-MM-DD
+- **instantiations.end.to**
+  YYYY-MM-DD
+- **place.label**
+  List of physical locations, would also include "Online".
+- **contributor.agent**
```
jamieparkinson, line 70: Ditto

rfcs/055-prismic-api/articles.md
jamieparkinson: Alternative would be renaming the field itself to agent, although that's a divergence from the prismic data

jamieparkinson (changes requested):

rfcs/055-prismic-api/articles.md
harrisonpim: good point!

jamieparkinson (approved):

rfcs/055-prismic-api/articles.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Filters
+
+- **publicationDate.from**
+  YYYY-MM-DD
+- **publicationDate.to**
+  YYYY-MM-DD
+- **contributors.contributor**
+  Includes author, artists, photographers, etc.
+- **format.label**
```
jtweed, line 98: Should this and the corresponding aggregation not be on `id` rather than `label`? Ie, `?format={id}&aggregations=format`?

jtweed (approved):
Approved subject to single comment about format id vs label filtering.

rfcs/055-prismic-api/articles.md
harrisonpim: done!

PR #91, opened 2023-02-22
Add internal terms to glossary

rcantin-w:
In an effort to ensure everyone uses the same language and help with onboarding, here is a incomplete glossary of the terms used in relation to the website content and context. Please contribute to them as I don't hold all the necessary knowledge (yet)!

A term might also be used differently on the UI, in Prismic and on the Backend, so we could make that distinction.

### DIFF
--- a/GLOSSARY.md
+++ b/GLOSSARY.md
@@ -55,6 +55,7 @@ Book is a content type in Prismic, displayed on the [Books listing page](https:/
 One of the _Location_ options, closed stores are places where only staff can go. A a reader has to make a request for items in closed stores.
 
 ##### Collection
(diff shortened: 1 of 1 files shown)


GLOSSARY.md
```diff
@@ -32,3 +33,141 @@
...
+
+##### Season
+A season is a grouping of most any _content types_, but primarily events, exhibitions, stories. The idea is that it's a time-limited, curated programme of activities, and it gets promoted as one programme. e.g.: [What does it mean to be human, now?](https://wellcomecollection.org/seasons/X84FvhIAACUAqiqp). They are displayed on `/seasons/[id]`.
+
+
+##### Serial
+A serial is a type of _series_, in which _articles_ are scheduled to be published.
+
+
+##### Series
```
jennpb, line 148: Events can be grouped into event series: a grouping of events under a regularly occurring singular title/theme (e.g. Exploring Research, Study Days)

Articles can be grouped into story series: a grouping of articles connected thematically. These may be guest edited and include articles written by multiple contributors (unlike a _serial_ which tends to have the same contributor(s) across all articles in the serial).

PR #92, opened 2023-02-23
RFC: Prismic to Elasticsearch ETL pipeline 

jamieparkinson:


rfcs/056-prismic-etl-pipeline/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+**Last updated:** 23/2/2023
+
+## Context
+In order to make our editorial content - including stories, comics, exhibitions and events - more discoverable, we want to be able to search it via an API as described in [RFC 055](https://github.com/wellcomecollection/docs/tree/main/rfcs/055-prismic-api).
+
+While Prismic does provide some search functionality of its own (which we have been using for initial versions of unified search), we want more control and fewer limitations. To achieve this we want to use Elasticsearch, as we do for our other search services. As such, we need to get data from Prismic into Elasticsearch: this RFC will propose the mechanism by which we achieve that.
+
+#### Desiderata
+- Changes (including additions and edits but possibly not deletions - see BEEPB OPOSDF) in Prismic are reflected promptly in the Elasticsearch index
```
alexwlchan, line 13: > See BEEPB OPOSDF

I thought this read [OSPF](https://en.wikipedia.org/wiki/Open_Shortest_Path_First) (the routing algorithm) because I am old. 👵 

What do you actually mean here?

rfcs/056-prismic-etl-pipeline/README.md
```diff
@@ -0,0 +1,45 @@
...
+## Proposal
+
+The Prismic corpus is smaller than our other corpora, and this ETL pipeline is very linear and self-contained. This points us towards a simple architecture, which helps with (a) speed/ease of reindexing and (b) maintainability across the team.
+
+![Proposed architecture](https://user-images.githubusercontent.com/4429247/220966188-097cac58-56f8-4c6d-afd8-48a17e50720c.png)
+
+1. A 'window generator' Lambda triggered on schedule, which generates a payload representing a time period/window that is sent to (2).
+2. A 'Prismic ETL' Lambda which consumes time periods (potentially half-bounded or unbounded, for complete reindexes) and then:
+   1. Queries using the `@prismicio/client` for documents with a [`last_publication_date`](https://prismic.io/docs/rest-api-technical-reference#q) within the specified window.
+   2. Transforms these documents to JSON objects with `display` and `query` fields corresponding to the entities/queries described in [RFC 055](https://github.com/wellcomecollection/docs/tree/main/rfcs/055-prismic-api).
```
alexwlchan, line 35: I think this may prove to be more complicated than it sounds, because of the way documents are linked in Prismic – I don't know if modified dates get propagated.

e.g. if you have a contributor who's written three articles:

```mermaid
graph TD
    A[article 1] --> contributor
    B[article 2] --> contributor
    C[article 3] --> contributor
```

we wouldn't have a document for the contributor in Elasticsearch, but we would have documents for each of the three articles.

If the contributor is updated in Prismic, they'd be fetched on the next window – how do we make sure the article documents get updated in Elasticsearch?

I don't think contributors know where they're linked from; the link exists on the article (or exhibition, or event).

rfcs/056-prismic-etl-pipeline/README.md
```diff
@@ -0,0 +1,45 @@
...
+   3. Indexes these into an Elasticsearch cluster using the Elasticsearch JS client's bulk helpers.
+   
+For complete reindexes, it would be straightforward to trigger the Prismic ETL lambda from a local script, with a payload that covers all documents.
+
+#### Technical implementation points
+- The intention is that the Prismic ETL Lambda will be written in TypeScript for maintainability, but one disadvantage of this is that we lose some of the patterns/tools that Scala gives us for reactive streaming data pipelines. Suggest we try [RxJS](https://rxjs.dev/) for this purpose.
+
+
+## Questions and potential issues
+- What do we do about deletions? Do we know if they happen? We could write something similar to the [CALM deletion checker](https://github.com/wellcomecollection/docs/tree/main/rfcs/032-calm-deletions) if necessary.
```
alexwlchan, line 45: I think the documents just disappear from the Prismic API.

alexwlchan (commented):
I’ll wait until this is more fleshed-out to discuss in detail.

It'd be good to explain why we're not using the Prismic web hooks – I remember we had a few issues in the past; would it be worth giving them another try?

rfcs/056-prismic-etl-pipeline/README.md
jamieparkinson: Interesting, will investigate this further

rfcs/056-prismic-etl-pipeline/README.md
jamieparkinson: Indeed, question is whether we actually delete stuff all that often. Will clarify.

alexwlchan: To be more exact: I think documents go to some sort of Prismic archive (or it looks that way in the GUI).

I don’t know if that’s accessible through the API.

I know documents disappear from a “get me an article with id X” article POV, but maybe they’re accessible through another API.

rfcs/056-prismic-etl-pipeline/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Context
+In order to make our editorial content - including stories, comics, exhibitions and events - more discoverable, we want to be able to search it via an API as described in [RFC 055](https://github.com/wellcomecollection/docs/tree/main/rfcs/055-prismic-api).
+
+While Prismic does provide some search functionality of its own (which we have been using for initial versions of unified search), we want more control and fewer limitations. To achieve this we want to use Elasticsearch, as we do for our other search services. As such, we need to get data from Prismic into Elasticsearch: this RFC will propose the mechanism by which we achieve that.
+
+#### Desiderata
+- Changes (including additions and edits but possibly not deletions - see [questions](#questions-and-potential-issues)) in Prismic are reflected promptly in the Elasticsearch index
+- Full reindexes are easy, quick and cheap to perform
+- Changes to the data mapping (and the index mapping) can be made easily by any developer
```
alexwlchan, line 15: I would expand this slightly, to cover further back in the process – it should be possible for any developer who's updating the Prismic model to update this data mapping *if necessary*, i.e.

- it should be possible to change the Prismic model definitions and leave this pipeline as-is if you're changing unmapped fields
- it should be possible to change the Prismic model definitions and coordinate your change with this pipeline

rfcs/056-prismic-etl-pipeline/README.md
```diff
@@ -0,0 +1,91 @@
...
+   1. Queries Prismic for all documents (including denormalised data on primary documents) updated within the time window.
+   2. For all secondary documents, query the ES index for already-indexed primary documents that contain them.
+   3. Queries Prismic for all the documents (including denormalised data) from (ii) that are not part of the list returned by (i).
+   4. Transforms the resultant primary documents into JSON objects as described above.
+   5. Indexes these into an Elasticsearch cluster using the Elasticsearch JS client's bulk helpers. 
+
+For complete reindexes, it would be straightforward to trigger the Prismic ETL lambda from a local script, with a payload that covers all documents. In this case, it would also be an easy optimisation to disable steps (ii) and (iii), as all documents would be being fetched regardless.
+
+#### Technical implementation points
+- The intention is that the Prismic ETL Lambda will be written in TypeScript for maintainability, but one disadvantage of this is that we lose some of the patterns/tools that Scala gives us for reactive streaming data pipelines. Suggest we try [RxJS](https://rxjs.dev/) for this purpose.
```
alexwlchan, line 82: There's a stronger argument for TS: sharing code/dev expertise between the existing front-end apps.

alexwlchan (commented):
This looks good; a few notes inline. I want to think about it a bit more before approving.

rfcs/056-prismic-etl-pipeline/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Last updated:** 23/2/2023
+
+## Context
+In order to make our editorial content - including stories, comics, exhibitions and events - more discoverable, we want to be able to search it via an API as described in [RFC 055](https://github.com/wellcomecollection/docs/tree/main/rfcs/055-prismic-api).
+
+While Prismic does provide some search functionality of its own (which we have been using for initial versions of unified search), we want more control and fewer limitations. To achieve this we want to use Elasticsearch, as we do for our other search services. As such, we need to get data from Prismic into Elasticsearch: this RFC will propose the mechanism by which we achieve that.
+
+#### Desiderata
+- Changes (including additions and edits but possibly not deletions - see [questions](#questions-and-potential-issues)) in Prismic are reflected promptly in the Elasticsearch index
```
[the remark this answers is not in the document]
jamieparkinson: Yeah, agreed, will add some more context about the actual requirements here

rfcs/056-prismic-etl-pipeline/README.md
jamieparkinson: Ah yes, good point (fortunately don't think it makes much difference to implementation?)

rfcs/056-prismic-etl-pipeline/README.md
jamieparkinson: Indeed, that's what I meant by maintainability - will clarify.

rfcs/056-prismic-etl-pipeline/README.md
alexwlchan: > fortunately don't think it makes much difference to implementation?

yeah, I think it's more a scenario to bear in mind rather than imposing big requirements

(I suppose the only way it might affect implementation is if we can create a consistent approach to doing this sort of task, but that could just as easily be "existing Prismic tasks copy the new way" as "the new way copies the existing Prismic tasks")

rfcs/056-prismic-etl-pipeline/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+1. A 'window generator' Lambda triggered on schedule, which generates a payload representing a time period/window that is sent to (2).
+2. A 'Prismic ETL' Lambda which consumes time periods (potentially half-bounded or unbounded, for complete reindexes) and then:
+   1. Queries Prismic for all documents (including denormalised data on primary documents) updated within the time window.
+   2. For all secondary documents, query the ES index for already-indexed primary documents that contain them.
+   3. Queries Prismic for all the documents (including denormalised data) from (ii) that are not part of the list returned by (i).
+   4. Transforms the resultant primary documents into JSON objects as described above.
+   5. Indexes these into an Elasticsearch cluster using the Elasticsearch JS client's bulk helpers. 
+
+For complete reindexes, it would be straightforward to trigger the Prismic ETL lambda from a local script, with a payload that covers all documents. In this case, it would also be an easy optimisation to disable steps (ii) and (iii), as all documents would be being fetched regardless.
```
paul-butcher, line 79: I wonder if this might be a job for Step Functions or Airflow (or maybe just a pair of similar-but-different queue-driven lambdas similar to how we split the bulk and SQS-driven paths in the Concepts pipeline).

Both paths could be independently triggered from the same window generator, then we could have one path that starts with primary documents and another that starts with secondary documents.

That way, rather than thinking of the optimisation to disable steps 2.2 and 2.3 for a full reindex, it would simply be running the normal primary document path over all documents since time immemorial.

Only the secondary document path would include steps 2.2 and 2.3

A further optimisation in 2.2 might be to query for already indexed primary documents that have not been updated since the latest secondary document changed.

jamieparkinson: This is very useful - I think what I might suggest is that we start with a single Lambda which is modularised in a way informed by this comment, with a view that - if the pipeline needs to expand - we might want to split up the monolith into DAG-y microservices further down the line.

alexwlchan (approved):

PR #93, opened 2023-04-13
Add an RFC about restructuring the catalogue pipeline

alexwlchan:
Rendered version: https://github.com/wellcomecollection/docs/blob/restructuring-the-catalogue-pipeline/rfcs/057-restructuring-the-catalogue-pipeline/README.md

Highlighting this bit from the introduction:

> I'm not saying we should definitely do this, but I'm writing the RFC to help me clarify some thoughts.

Although I'm not committing to this work, I'd be interested in thoughts – does this make sense? Does it seem sensible? What have I missed? And I think an RFC is a better home for that discussion than a Slack thread.


kenoir:
This is still useful, but unlikely we'll merge as is or work directly from this branch, so closing this PR.

PR #96, opened 2024-01-10
Add terms to glossary

rcantin-w:
## What does this change?
Adds Audience, Interpretation types and Place as well as adds to Location.

People understand the definitions

It confuses people further 😄


jcateswellcome (commented):
All good

PR #97, opened 2024-01-19
RFC 060: Offsite requesting

jamieparkinson:
https://github.com/wellcomecollection/docs/blob/rfc-offsite-requesting/rfcs/060-offsite-item-requesting/README.md


kenoir:
This all makes sense, nothing here I disagree with.

> Is it really OK to have no machine-readable method for discriminating offsite items in our APIs?

I suppose we could always add a `closed-stores-offsite` type if we found we needed it?

agnesgaroux:
Just to make sure I understand it correctly:
the `DisplayAccessMethod` is the unique place where we'll hold information about where an item is and where it can be viewed, with a different value for every case:
- in our building and viewable in our building 
- in an off-site location and viewable in our building, providing for (eventually) multiple off-site locations 
- in an off-site location and viewable at this off-site location, providing for (eventually) multiple off-site locations

jamieparkinson:
Not quite - there wouldn't be a new access method for offsite-but-viewed-onsite, as the way in which these items are accessed would still be the same.

Good question re multiple offsite locations - it's hard to know what we'll need, but the intention at least initially is that there'd be one access method for all offsite-viewed items, and any differences between processes would be determined internally in our applications (kind of comparable to the offsite-but-viewed-onsite items, where specifics are also hidden from users).

agnesgaroux:
@jamieparkinson as discussed last week, we need a way to infer the viewing venue (to get the opening times) and its lead time (to trim the list of opening times so it starts at the 1st available viewing date) 

>  there wouldn't be a new access method for offsite-but-viewed-onsite, as the way in which these items are accessed would still be the same.

Assuming that "offsite-but-viewed-onsite" items have a longer lead time than "onsite-and-viewed-onsite" items (since they have to be transported from Deepstore), for an item that has a `DisplayAccessMethod` of "online-request", how do we know whether it can be viewed from tomorrow (= "onsite-and-viewed-onsite") or say, from next week (= "offsite-but-viewed-onsite") 
As I said, that's assuming that deepstore items have a longer lead time than stuff that is stored in our building. If it's the same then no problemo 
Same thing with the off-side locations btw, if they all have the same lead time then easy peasy. Otherwise we need a way to distinguish between venues, and different lead times, in order to know when the `AvailabilitySlots` list should start

jamieparkinson:
Ah OK I see, I believe this is covered by _"Items API to use data direct from Sierra (as it does currently) to discriminate offsite items and to return appropriate availability slots."_.

ie because the items API is already getting the Sierra location data it can "know" the internal details even when the externally-presented model doesn't (this will also apply for the viewed-offsite stuff). Is there something I've missed?

agnesgaroux:
Ha yes, I remember now getting confused about that when we talked about it 🙃 sorry my bad

PR #108, opened 2024-10-10
RFC for collection graph work

Amvla:
The RFC provides an overview of graphs and how they might be useful, a detailed summary from a deep dive into the archived [knowledge-graph](https://github.com/wellcomecollection/knowledge-graph/tree/main) repo, and some considerations for future development.


kenoir:
Preview link: https://github.com/wellcomecollection/docs/blob/rfc-knowledge-graph/rfcs/062-knowledge-graph/README.md

rfcs/062-knowledge-graph/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+**Pipeline**
+
+Apply learnings from the Wellcome Academic Graph's pipeline design where applicable. For example, consider using the Neo4j driver with Cypher queries to load data to Neo4j and removing the neomodel dependency for better performance, leveraging existing expertise from the machine learning team. Refactoring changes to increase pipeline performance may also include preprocesssing all source data into nodes and edges before bulk loading them to Neo4j, instead of enriching and loading each work iteratively. Finally, ensure the code meets machine learning team standards by improving Python import readability, adding correctly formatted docstrings, enabling auto-formatting, and adding unit tests.
+
+**Infrastructure**
+
+Consider deploying the Neo4j database, which was initially only set up locally with the official Docker image in v1 of the graph, which would enable broader access by the platform and ML teams. If the graph is used to enrich concepts for search, plan its integration with existing infrastructure and processes. As above, apply learnings from the Wellcome Academic Graph's setup where applicable.
```
kenoir, line 226: Would [AWS Neptune](https://aws.amazon.com/neptune/) be a useful tool in this instance?

kenoir:
There might be some interesting overlap with other places in the catalogue pipeline where we are required to extract and append graph-like relationships from Work data unrelated to concepts, for example constructing `collectionPath` and appending relations to works for navigation where there is a hierarchical relationship between works. Mostly this is for archives where there is an explicit hierarchy in the domain model, but [also between some library records](https://github.com/wellcomecollection/docs/tree/main/rfcs/046-transitive-sierra-hierarchies) (e.g. the Fallaize collection).

kenoir (approved):
Great overview.

rfcs/062-knowledge-graph/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Overview of knowledge graph v1 (archived)
+
+Prototype pipelines were developed in 2022 to construct a [Neo4j](https://neo4j.com/) graph for the Collection. The source code (written in Python) can be viewed in the archived [knowledge-graph](https://github.com/wellcomecollection/knowledge-graph) repository on github. The code is not well-documented and does not include tests. Specific pipline logic can only be deduced from reading the code itself. The purpose of this graph was to enrich existing concepts with data from Library of Congress, Medical Subject Headings (MeSH), and Wikidata. These sources reference each other via fields such as "closely matching" identifiers, and this was used to create edges between concepts in the graph. This graph was only used for enrichment of existing concepts, not to extract/infer new concepts. Its main purpose was to surface this data in an elasticsearch index. As far as it is possible to tell from the repository code, the graph database itself was not associated with any dedicated infrastructure and only created locally.
```
rcantin-w, line 47:
```suggestion
Prototype pipelines were developed in 2022 to construct a [Neo4j](https://neo4j.com/) graph for the Collection. The source code (written in Python) can be viewed in the archived [knowledge-graph](https://github.com/wellcomecollection/knowledge-graph) repository on GitHub. The code is not well-documented and does not include tests. Specific pipeline logic can only be deduced from reading the code itself. The purpose of this graph was to enrich existing concepts with data from Library of Congress, Medical Subject Headings (MeSH), and Wikidata. These sources reference each other via fields such as "closely matching" identifiers, and this was used to create edges between concepts in the graph. This graph was only used for enrichment of existing concepts, not to extract/infer new concepts. Its main purpose was to surface this data in an Elasticsearch index. As far as it is possible to tell from the repository code, the graph database itself was not associated with any dedicated infrastructure and only created locally.
```

typo to "pipeline" and capitalised some tool names while I was at it but it's very much a np; enjoying my read so far 👀 😊

rfcs/062-knowledge-graph/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+**Graph model**
+
+Re-evaluate the graph model and make sure it represents internal and external data effectively. Ensure the graph model is efficient and supports downstream tasks like indexing, machine learning analyses, and visualisation. Verify that edge directionality aligns with the edge labels, such as reversing HAS_CONCEPT edges to reflect semantic directionality. Consider whether undirected edges should exist between source concept nodes.
+
+**Pipeline**
+
+Apply learnings from the Wellcome Academic Graph's pipeline design where applicable. For example, consider using the Neo4j driver with Cypher queries to load data to Neo4j and removing the neomodel dependency for better performance, leveraging existing expertise from the machine learning team. Refactoring changes to increase pipeline performance may also include preprocesssing all source data into nodes and edges before bulk loading them to Neo4j, instead of enriching and loading each work iteratively. Finally, ensure the code meets machine learning team standards by improving Python import readability, adding correctly formatted docstrings, enabling auto-formatting, and adding unit tests.
```
rcantin-w, line 222:
```suggestion
Apply learnings from the Wellcome Academic Graph's pipeline design where applicable. For example, consider using the Neo4j driver with Cypher queries to load data to Neo4j and removing the neomodel dependency for better performance, leveraging existing expertise from the machine learning team. Refactoring changes to increase pipeline performance may also include preprocessing all source data into nodes and edges before bulk loading them to Neo4j, instead of enriching and loading each work iteratively. Finally, ensure the code meets machine learning team standards by improving Python import readability, adding correctly formatted docstrings, enabling auto-formatting, and adding unit tests.
```

jcateswellcome:
Sorry this is more of a question than a comment - the RFC is a great overview! Do we understand the rationale for using a spreadsheet as the source for stories? Is this as a proof of concept graph just to include them so that these could be associated with people/concepts? How would this be done? Makes sense to exclude in the firs instance and build this with just the catalogue? Interesting!

rfcs/062-knowledge-graph/README.md
Amvla: That's a good point, it's definitely worth exploring other options including Neptune (I've added a note in this section). There's also Neo4j AuraDB but I'm not sure how exactly that fits into their pricing structure.

Amvla:
> There might be some interesting overlap with other places in the catalogue pipeline where we are required to extract and append graph-like relationships from Work data unrelated to concepts, for example constructing `collectionPath` and appending relations to works for navigation where there is a hierarchical relationship between works. Mostly this is for archives where there is an explicit hierarchy in the domain model, but [also between some library records](https://github.com/wellcomecollection/docs/tree/main/rfcs/046-transitive-sierra-hierarchies) (e.g. the Fallaize collection).

This is very interesting! This is actually answering a question I had noted down during my exploration around whether works can be connected to each other as well. One initial idea I had was whether we could add (weighted) edges representing visual/semantic similarity (I'm aware visual similarity has already been implemented in search, but having it in the graph could still be useful for other ML or visualisation approaches).

Amvla:
> Sorry this is more of a question than a comment - the RFC is a great overview! Do we understand the rationale for using a spreadsheet as the source for stories? Is this as a proof of concept graph just to include them so that these could be associated with people/concepts? How would this be done? Makes sense to exclude in the firs instance and build this with just the catalogue? Interesting!

I'm also curious about this. I was a bit surprised to find this mysterious spreadsheet referenced in the code when I looked through the pipeline.

PR #109, opened 2024-10-18
RFC 063: Catalogue Pipeline services from ECS to Lambda 

kenoir:
## What does this change?

Part of https://github.com/wellcomecollection/platform/issues/5789

This change adds an RFC to discuss proposed pipeline changes to move to using AWS Lambda for pipeline services, instead of autoscaling ECS services.

## How to test

- [ ] [Read the documentation,](https://github.com/wellcomecollection/docs/blob/rk-ag/ecs-lambda-pipeline/rfcs/063-catalogue_pipeline_lambdas/README.md) does it make sense?

Clear reasons for making this infrastructure change, and consensus on the direction of travel.

We should be clear about the risks involved and how best to to determine success.

### DIFF
--- a/rfcs/063-catalogue_pipeline_lambdas/README.md
+++ b/rfcs/063-catalogue_pipeline_lambdas/README.md
@@ -147,7 +147,9 @@ Some suggestions for future improvements that could be looked at as part of this
 
 - Can we simplify batcher service processing logic?
 
-- Given that the path_concatenator only processes Sierra record trees, would it be better for this to happen earlier in the pipeline? Could be just after the Sierra transformer business, which removes the need for the router to route the relevant works to have their paths concatenated.  
+- Can we remove the use of [Pekko Streams](https://pekko.apache.org/docs/pekko/current/general/stream/stream-design.html) in some pipeline services? The logic can be  complex and difficult to understand.
+
+- Given that the `path_concatenator` only processes Sierra record trees, would it be better for this to happen earlier in the pipeline? Could be just after the Sierra transformer business, which removes the need for the router to route the relevant works to have their paths concatenated.  
 
 - If we were to wait at the matcher/merger stage for every work to be processed, could we then do without the batcher? We would be able to scan the store populated by the matcher/merger and send all members of a tree together to have their relations embedded, removing the need for a mechanism that limits the “explosion” we have now that is caused by only sending partial trees to the relation_embedder.  
 
--- a/rfcs/063-catalogue_pipeline_lambdas/README.md
+++ b/rfcs/063-catalogue_pipeline_lambdas/README.md
@@ -155,4 +155,4 @@ Some suggestions for future improvements that could be looked at as part of this
 
 - Should we look into SNS features that allows conditional subscription based on the message body as a replacement for the router?
 
-- Can we consider replacing intermediate indices with S3? We don’t search them; they are acting as a simple JSON stores. 
+- Can we consider replacing intermediate indices with S3? We don't search them; they are acting as a simple JSON stores. 
--- a/rfcs/063-catalogue_pipeline_lambdas/README.md
+++ b/rfcs/063-catalogue_pipeline_lambdas/README.md
@@ -2,76 +2,138 @@
 
 ## Context
 
-Catalogue-pipeline services run on ECS. The infrastructure includes scaling to account for resource needs during a reindex (scale up) or day-to-day running of the pipeline (scale down). Scaling is achieved using ECS auto scaling rules driven from CloudWatch metrics on queue depth.  
-
-ECS services 
-
----
+Catalogue-pipeline services run on [Elastic Container Service (ECS)](https://aws.amazon.com/ecs/). The infrastructure includes scaling to account for resource needs during a reindex (scale up) or day-to-day running of the pipeline (scale down). Scaling is achieved using ECS auto scaling rules driven from CloudWatch metrics on queue depth. Other newer projects leverage AWS Lambda to run similar tasks, and there is an opportunity to remove complexity in the catalogue pipeline and align with newer projects.
 
 ## Background
 
-Why we want to make the change
+The following points cover in more detail the reasoning behind moving from ECS based compute to AWS Lamdba:
 
-1. Reduction in infra complexity by chucking scaling
+- **Reduction in infra complexity by removing ECS auto-scaling logic:**
 
-Scaling the pipeline up or down is achieved in several ways:  
--> a manual step is performed to terraform the pipeline into "reindexing_state" by setting a bunch of variables to `true` [here](pipeline/terraform/2024-08-15/main.tf). Search for `reindexing_state` in the catalogue-pipeline repo to find all the ways these variables are used.  
--> every service also has a high/low message count alarm on its input queue. By [default](pipeline/terraform/modules/stack/variables.tf) services scale up to 12 instances as long as there are messages on the queue, and will scale down when the queue is clear. However some services (eg. ingestors) have specially configured `max_capacity` usually to account for data stores' read/write limits.
+   Scaling the pipeline up or down is achieved in several ways:
+   
+   - A manual step is performed to terraform the pipeline into "reindexing_state" by setting a bunch of variables to `true` [here](pipeline/terraform/2024-08-15/main.tf). Search for `reindexing_state` in the catalogue-pipeline repo to find all the 
+     ways these variables are used.  
+   - Every service also has a high/low message count alarm on its input queue. By [default](pipeline/terraform/modules/stack/variables.tf) services scale up to 12 instances as long as there are messages on the queue, and will scale down when the queue 
+     is clear. However some services (eg. ingestors) have specially configured `max_capacity` usually to account for data stores' read/write limits.
 
-2. Reduction in app complexity by chucking akka/pekko batching
+   This change aims to remove the need for this scaling logic.
 
-Several (SQS driven) pipeline services implement “SQSStream” (see https://github.com/wellcomecollection/scala-libs/blob/main/messaging/src/main/scala/weco/messaging/sqs/SQSStream.scala ) that batches SQS messages so that they can be processed in bulk. This logic lives inside the service and makes it difficult (impossible?) to run locally.  
-This model means the messaging and business/processing logic part of the services are entangled in sometimes confusing ways, and we're losing queue management features of AWS SQS (eg. having to keep track of success and failure to delete or retry messaging accordingly).  
-This change is also an opportunity to reassess and possibly refactor out some error handling that may be too particular, over-optimising for performance at the expense of clarity. 
+- **Reduction in app complexity by removing akka/pekko batching:**
 
-3. Potential reductions in cost 
+   Several (SQS driven) pipeline services implement “SQSStream” (see https://github.com/wellcomecollection/scala-libs/blob/main/messaging/src/main/scala/weco/messaging/sqs/SQSStream.scala ) that batches SQS messages so that they can be processed in 
+   bulk. This logic lives inside the service and makes it difficult (impossible?) to run locally.  
+
+   This model means the messaging and business/processing logic part of the services are entangled in sometimes confusing ways, and we're losing queue management features of AWS SQS (eg. having to keep track of success and failure to delete or retry 
+   messaging accordingly).  
  
-It is worth noting that the ECSs are configured to scale down to 0 (effectively shutting down) when the input is low, which means we are not, in fact, paying for continously-running tasks. One the other hand, this often means starting a task and stopping it repeatedly as messages trickle down the pipeline.  
-Costs for 2023-10-01 to 2024-09-30:  
-EC2 instances: $3,266.79  
-EC2 - Other: $6,639.06
+   This change is also an opportunity to reassess and possibly refactor out some error handling that may be too particular, over-optimising for performance at the expense of clarity. 
 
-4. Potential increase in speed (deployment and processing)
+- **Potential reductions in cost:**
+ 
+   ECS services are configured to scale down to 0 when the input is low, which means we are not paying for continously-running tasks, but this results in startinf and stopping tasks repeatedly as messages trickle down the pipeline.  
 
-Given that our services shut down when not in use, lambda would probably provide faster processing. Even accounting for cold starts lambda would be ready to process requests much faster than a EC2 service can. Using provisioned concurrency would further increase processing speed, but is likely not necessary, at least for most services. 
+- **Potential increase in speed (deployment and processing):**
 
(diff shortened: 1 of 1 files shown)


rfcs/063-catalogue_pipeline_lambdas/README.md
```diff
@@ -0,0 +1,139 @@
...
+- **Reduction in infra complexity by removing ECS auto-scaling logic:**
+
+   Scaling the pipeline up or down is achieved in several ways:
+   
+   - A manual step is performed to terraform the pipeline into "reindexing_state" by setting a bunch of variables to `true` [here](pipeline/terraform/2024-08-15/main.tf). Search for `reindexing_state` in the catalogue-pipeline repo to find all the 
+     ways these variables are used.  
+   - Every service also has a high/low message count alarm on its input queue. By [default](pipeline/terraform/modules/stack/variables.tf) services scale up to 12 instances as long as there are messages on the queue, and will scale down when the queue 
+     is clear. However some services (eg. ingestors) have specially configured `max_capacity` usually to account for data stores' read/write limits.
+
+   This change aims to remove the need for this scaling logic.
```
paul-butcher, line 20: Would it be truer to say "reduce", or "reduce and possibly remove" (although that latter sentiment is probably conveyed by "aims" before)?

It's possible that the `max_capacity` limitation, and the need to embiggen the datastores may still need to be covered by something resembling the scaling logic.

rfcs/063-catalogue_pipeline_lambdas/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+- **Reduction in app complexity by removing akka/pekko batching:**
+
+   Several (SQS driven) pipeline services implement “SQSStream” (see https://github.com/wellcomecollection/scala-libs/blob/main/messaging/src/main/scala/weco/messaging/sqs/SQSStream.scala ) that batches SQS messages so that they can be processed in 
+   bulk. This logic lives inside the service and makes it difficult (impossible?) to run locally.  
+
+   This model means the messaging and business/processing logic part of the services are entangled in sometimes confusing ways, and we're losing queue management features of AWS SQS (eg. having to keep track of success and failure to delete or retry 
+   messaging accordingly).
```
paul-butcher, line 28: 🔥

rfcs/063-catalogue_pipeline_lambdas/README.md
```diff
@@ -0,0 +1,139 @@
...
+- **Reduction in app complexity by removing akka/pekko batching:**
+
+   Several (SQS driven) pipeline services implement “SQSStream” (see https://github.com/wellcomecollection/scala-libs/blob/main/messaging/src/main/scala/weco/messaging/sqs/SQSStream.scala ) that batches SQS messages so that they can be processed in 
+   bulk. This logic lives inside the service and makes it difficult (impossible?) to run locally.  
+
+   This model means the messaging and business/processing logic part of the services are entangled in sometimes confusing ways, and we're losing queue management features of AWS SQS (eg. having to keep track of success and failure to delete or retry 
+   messaging accordingly).  
+ 
+   This change is also an opportunity to reassess and possibly refactor out some error handling that may be too particular, over-optimising for performance at the expense of clarity. 
+
```
paul-butcher, line 31: Yes, but I'm not sure this work is necessary to facilitate that.  The weird batcher error handling could be rewritten even if we left the rest as-is.

rfcs/063-catalogue_pipeline_lambdas/README.md
```diff
@@ -0,0 +1,139 @@
...
+- **Potential reductions in cost:**
+ 
+   ECS services are configured to scale down to 0 when the input is low, which means we are not paying for continously-running tasks, but this results in startinf and stopping tasks repeatedly as messages trickle down the pipeline.  
+
+- **Potential increase in speed (deployment and processing):**
+
+   ECS services take a few minutes to start tasks as the auto-scaling rules rely on reported CloudWatch metrics, whereas Lambda invocations are [much faster](https://docs.aws.amazon.com/lambda/latest/operatorguide/execution-environments.html#cold-start-latency). Using [provisioned concurrency](https://docs.aws.amazon.com/lambda/latest/dg/provisioned-concurrency.html) may further increase processing speed, but is likely not necessary, at least for most services.
+
+  Service deployment at present relies on the ECS deployment APIs which take some time to determine that a service is "stable", there is no such process for AWS Lambda and deployments are almost instant.
+
```
paul-butcher, line 41: I think another speed aspect is that we currently try to compromise between normal running and reindexing with the way we gather messages.  This, along with the way we pull messages into an internal queue until the application decides it's got enough messages/waited long enough means that we can see artificial delays during normal running, and towards the end of a reindex (e.g. because multiple ECS instances have pulled some of the messages, and now there aren't enough to go around)

rfcs/063-catalogue_pipeline_lambdas/README.md
```diff
@@ -0,0 +1,139 @@
...
+   This allows for most efficient use of the infrastructure/resources and avoids one of the oft-cited complexity/downside in the catalogue-pipeline, namely that the application tries to handle two quite different use cases.  
+
+   The different versions of the aggregator and recorder are also configured for their purpose, thus making the infrastructure code easier to read and understand.
+
+## Suggested solution
+
+We could incrementally refactor the catalogue-pipeline services to use lambdas instead of ECS. Some considerations are:
+
+ - We currently build container images to packages our code for deployment, we may wish to continue doing so for simplicity. See the documentation around [packaging and deploying container based Java Lambdas](https://docs.aws.amazon.com/lambda/latest/dg/java-image.html) that we would need to use to achieve this.
```
paul-butcher, line 55:
```suggestion
 - We currently build container images to package our code for deployment, we may wish to continue doing so for simplicity. See the documentation around [packaging and deploying container based Java Lambdas](https://docs.aws.amazon.com/lambda/latest/dg/java-image.html) that we would need to use to achieve this.
```
minor typo

rfcs/063-catalogue_pipeline_lambdas/README.md
```diff
@@ -0,0 +1,139 @@
...
+  could also leverage SQS lambda features to protect the data stores (eg. elasticsearch index at the ingestor stage) from overloading by setting a maximum concurrency on the event source.
+   
+- We can leverage SQS event source and lambda features to batch up to 10000 messages before invoking the function (useful in the context of the batcher service). This will make it possible to remove the batching mechanism that exists inside the 
+  services, making it easier to run them locally using a [lambda runtime emulator](https://github.com/aws/aws-lambda-runtime-interface-emulator).
+
+  Error handling could use [Partial batch responses](https://docs.aws.amazon.com/lambda/latest/dg/services-sqs-errorhandling.html#services-sqs-batchfailurereporting).
+  
+- Lambda is generally considered to be cost-effective for workloads with intermittent or unpredictable traffic patterns (1ms for 1024MB: $0.0000000167).
+
+- Processing speed could be slowed down by [AWS Lambda cold starts](https://aws.amazon.com/blogs/compute/operating-lambda-performance-optimization-part-1/) but this can be optimised by provisioning concurrency.
```
paul-butcher, line 67: It's good to mention this as a potential mitigation, but it may be unnecessary and this may not be the best place to start looking in order to reduce overall latency.
 
During normal running, this model is likely to be at least as fast as the current approach (because an ECS is also quite likely to have spun down and now needs to come up again), and real time processing is not expected anyway. 

When it's busy, the Lambdas are likely to be reused.

We may get some good results from looking at what we can hold onto in memory between runs.  Similarly, we may be able to do away with Akka in some places, which would reduce both the package size and the internal startup time.

rfcs/063-catalogue_pipeline_lambdas/README.md
```diff
@@ -0,0 +1,139 @@
...
+  
+- The concepts-pipeline has demonstrated the value of using AWS Lambdas in a data pipeline.
+
+
+### Transition from ECS to Lambda
+
+In terms of transition from ECS to lambdas to prevent downtime and allow a smooth transition:
+
+- Update deployment to allow deployment of existing container images to both ECS & Lambda runtime environments.
+- Move services over one at a time, testing that a new Lambda service works by dual-running it alongside its ECS counterpart without publishing messages downstream.
```
paul-butcher, line 77: We'll need to do something to ensure that messages can be pulled by both systems.  Are we going to have a duplicate input queue populated by the same upstream SNS?

For stages that store data, rather than Batcher/Router/Matcher that just push messages downstream, do we want to consider in this RFC  how that is to be done? (just write to S3, regardless of whether it's a stage that needs to output an index?)

rfcs/063-catalogue_pipeline_lambdas/README.md
```diff
@@ -0,0 +1,139 @@
...
+- When satisfied AWS Lambda behaviour for a service has reached parity we can remove the ECS service and infrastructure.
+
+## Potential issues
+
+AWS Lambda has a different execution model than ECS, with SQS messages causing Lambda invocations triggered by an AWS Lambda Trigger (managed invocations) and configuration options to dictate how many messages a Lambda invocation handles, how many Lambdas can run at once, and how long any single Lambda can run for. In ECS, tasks are launched by autoscaling rules, those tasks then poll SQS for messages processing and deleting them as they are received. In ECS when there is no more work to do tasks are terminated based on autoscaling rules.
+
+Potential issues may arise around this change in exectution model, specifically:
```
paul-butcher, line 84:
```suggestion
Potential issues may arise around this change in execution model, specifically:
```
minor typo

rfcs/063-catalogue_pipeline_lambdas/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Concurrency
+
+With ECS concurrency can be limited by the number of running tasks which provides a _rough_ restriction on the number of messages processed at once. With AWS Lambda concurrency can be specified exactly with ["reserved concurrency"](https://docs.aws.amazon.com/lambda/latest/dg/configuration-concurrency.html), and in combination with [batch size](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#sqs-polling-behavior) can tightly control the number of messages a Lambda processes.
+
+As we are moving from one model to another, we will need to calibrate concurrency for each service by estimation and experiment. 
+
+Concurrency is of particular concern where we are making network requests to external services that may not be configured to meet demand, for example Elasticsearch or RDS databases.
```
paul-butcher, line 92: ✨

rfcs/063-catalogue_pipeline_lambdas/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Batching
+
+Batching describes the number of messages handed to a Lambda to process in one invocation. For the catalogue pipeline we will be invoking Lambdas from non-FIFO SQS messages, and this has [specific restrictions around batch size](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html).
```
paul-butcher, line 96: There is an interesting potential conflict here.

FIFO queues allow Content Based Deduplication, which could be an improvement on the current setup (because of the explosion downstream of the relation embedder), but if they can't provide enough messages for efficient bulk updates in ES, then we won't be able to take advantage of that feature.

rfcs/063-catalogue_pipeline_lambdas/README.md
```diff
@@ -0,0 +1,139 @@
...
+- Length of batch window, (how long to wait before passing on a batch), [up to 5 minutes](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html).
+
+Along with concurrency batching dictates how much work is processed at one time. 
+
+A specific concern around batch size: 
+
+> The maximum number of records in each batch that Lambda pulls from your stream or queue and sends to your function. Lambda passes all of the records in the batch to the function in a single call, up to the payload limit for synchronous invocation 
+  (6 MB). The maximum batchsize for lambda is 10 000. This is far less than the batch size that the batcher is currently able to process (up to 120000 `collectionPath`). This could mean fewer nodes are being matched in a batch, reducing the 
+  beneficial effect of the batcher on the relation_embedder load. As of 2024-10-23 there are 271791 documents with a collectionPath in the merged index, ie. as many messages that the batcher needs to process as part of a full reindex
+
```
paul-butcher, line 111: It might be worth mentioning here that even in the current system,  `flushInterval` and `maxBatchSize` are set to try to balance between "wait until you have loads of messages" and "don't wait forever", so it does not always process 120000.

During a reindex, I'm pretty sure there are more than three invocations of the batcher.

rfcs/063-catalogue_pipeline_lambdas/README.md
```diff
@@ -0,0 +1,139 @@
...
+When deciding on batching configutation, maximum execution time must be considered in order that we do not attempt to process work that exceeds that limit.
+
+While running on ECS batch size was less of a concern as there is no upper execution time, and messages are received by polling SQS for more messages while a task is executing.
```
paul-butcher, line 114: This is an insightful point.  I have run afoul of this problem in the editorial photography stuff.

rfcs/063-catalogue_pipeline_lambdas/README.md
kenoir: I'll change this to be less specific: "an opportunity to reassess and possibly reduce incidental complexity".

rfcs/063-catalogue_pipeline_lambdas/README.md
kenoir: Yes, i'll rephrase for the document and add this point.

rfcs/063-catalogue_pipeline_lambdas/README.md
kenoir: Yes, good point that there are other optimisations besides provisioned concurrency. How to hold onto things in-memory for containerised lambdas is not clear to me, and is further mudied by these being JVM based applications ... I guess it may be as easy as putting variables in global scope?

Besides the options around [static initialization](https://docs.aws.amazon.com/lambda/latest/operatorguide/static-initialization.html#:~:text=The%20INIT%20code%20runs%20when,uses%20a%20warm%20execution%20environment.) we could also use /tmp?

I'll add this to the doc.

rfcs/063-catalogue_pipeline_lambdas/README.md
kenoir: > Are we going to have a duplicate input queue populated by the same upstream SNS?

Yes, that's what I imagined we'd do. Where services are stateful we'll have to take a decisions on how to handle state.

rfcs/063-catalogue_pipeline_lambdas/README.md
kenoir: I'll add this point to the doc.

PR #112, opened 2024-11-07
"All" search RFC

rcantin-w:
Relates to https://github.com/wellcomecollection/content-api/issues/148

Preview link: https://github.com/wellcomecollection/docs/blob/all-search-rfc/rfcs/062-content-api-all-search/README.md


rfcs/062-content-api-all-search/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+<img src="./assets/prototype.png" alt="Design prototype for the 'All' search page" />
+
+[Consult the design prototype here](https://www.figma.com/design/qssPpJy1lOWSFtuACajkZr/Global-search?node-id=4656-12994&node-type=canvas&m=dev).
+
+There is to be no filtering nor sorting feature on this page. Therefore, we aim to build something with minimalism in mind, allowing us to have query performance at the forefront of our concerns.
+
+We will do so by creating a new endpoint: https://api.wellcomecollection.org/content/v0/all.
+
+Its response will return [an ordered list of Addressable content types](#api-response-addressable-content-types-list), as well as [the relevant results for the Collection search](#api-response-collection-search). (**TODO: figure out what that looks like**)
```
kenoir, line 22: It might make sense for the catalogue results to be loaded from a different endpoint, and potentially for them to be loaded async on the client.

rfcs/062-content-api-all-search/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+This document is a special case, in that it is one Prismic document that needs to be indexed as two documents: "Audio with transcripts" and "British sign language with subtitles", as they are two different pages on the website ([Audio with transcripts](https://wellcomecollection.org/guides/exhibitions/jason-and-the-adventure-of-254/audio-without-descriptions) and [British sign language with subtitles](https://wellcomecollection.org/guides/exhibitions/jason-and-the-adventure-of-254/bsl)).
+
+### Have all fields in the "query" object align
+
+Something that will help the search performance would be to have as little fields to look through as possible, and have their names match across content types. We still want certain fields to have more weight than others in terms of relevance, so an option would be to list all the Prismic fields we want to have for each content type, and order them in terms of importance. Then, in the index, the query object could look like:
+
+```
+query: {
+    weight100: "[title]. [subtitle].",
```
kenoir, line 61: This pattern of naming indexed query fields by weight is not something we seem to do elsewhere and might not be slicing things in a useful way. I think it makes more sense to give these names related to their content and leave any reference of weighting in search to the search query.

rfcs/062-content-api-all-search/README.md
rcantin-w: Could/should the result be returned from the Content API then, or should it be a new endpoint built in the Catalogue API?

rfcs/062-content-api-all-search/README.md
rcantin-w: Works for me, so it's just about making sure that all 10 content types are using the same term for what serves the same purpose (like mentioned in the Captions, standfirsts and intro texts" of this document) then.

rfcs/062-content-api-all-search/README.md
kenoir: It's probably worth checking in with @dana-saur, but allowing these results to be loaded async means this page can gracefully degrade, i.e. if the catalogue search is offline, content can still be searched for.

rfcs/062-content-api-all-search/README.md
kenoir: > so it's just about making sure that all 10 content types are using the same term for what serves the same purpose

👍

rfcs/062-content-api-all-search/README.md
rcantin-w: Absolutely, I don't think she'll have any objections as it's just about being useful to the user to the extent of our capacities, but can confirm.

I thought the returned results could/should be minimal ([WIP here](https://github.com/wellcomecollection/docs/pull/112/files#diff-389a05c08a4a5c8541620f19bbab9b36b3d6561c5a0466a4f519e8adf59cac8d)), we really don't need that much, which is why I thought of a different endpoint.

rfcs/062-content-api-all-search/README.md
rcantin-w: If we do what we said and add the strings of other fields in "description", maybe. Or add a fourth one ("contributors") as I feel like that's the main outlier that should have influence (e.g. if you search by artist name)

rfcs/062-content-api-all-search/README.md
kenoir: > I thought the returned results could/should be minimal

The catalogue API output when unadorned by `include` returns minimal information about works & images (though images is quite a bit more wordy than works).

kenoir:
To finalize this RFC I recommend updating to include:

- Indicate that it will use existing catalogue API responses, but load them asynchronously
- Include a proposed elasticsearch mapping for the all index
- Include an architecture diagram, showing which services are loading documents into the "all" index

rfcs/062-content-api-all-search/README.md
rcantin-w: [Now addressed](https://github.com/wellcomecollection/docs/pull/112/commits/a2f2201a4e59530214fce9b9b5cdcad731f238da)

rfcs/062-content-api-all-search/README.md
rcantin-w: [Addressed here ](https://github.com/wellcomecollection/docs/pull/112/commits/6c44091273c59ea57372f91ebb7a853f0ec56a35)

rfcs/062-content-api-all-search/README.md
```diff
@@ -54,19 +54,20 @@ This document is a special case, in that it is one Prismic document that needs t
...
 }
 ```
 
-This is of course a very rough example that hopefully explains the theory.
+Most content types will allow this to fit, with one note on the `decription` field: 
 
-#### Captions, standfirsts and intro texts
+#### Description, captions, standfirsts and intro texts
 
 We have built our content types to use an array of fields to serve the same purpose; what could be called a "description" of the document gets called "Promo caption", "standfirst" (which is a slice, so part of the body), or "Intro text". There is [a ticket which aims to address the case of the Standfirst slices](https://github.com/wellcomecollection/wellcomecollection.org/issues/10753), but in the meantime, I suggest we use only one name for these in the index: "`description`". We will need to determine which content type should use which field as a description, but once that gets indexed, it becomes much easier to reference it by one name, at least in the "display" object.
```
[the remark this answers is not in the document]
rcantin-w: Do you mean using both `caption` and `standfirst`, for example?
If so, if we wanted to we could append both of those in the `description` string.

rfcs/062-content-api-all-search/README.md
```diff
@@ -3,7 +3,10 @@
 <!-- TODO add relevant sections -->
 
 - [Background information](#background-information)
-- ["All" index](#all-index)
+- [Elasticsearch "All" index](#elasticsearch-all-index)
+- [Indexation/Updates](#indexationupdates)
```
kenoir, line 7: Should this be "Indexing"?

rcantin-w: Probably!

rcantin-w: Amended, will close this

rcantin-w:
There is a diagram yet to do, but I'll start opening up the ticket for review.

kenoir (approved):

PR #113, opened 2024-11-18
Renumber All search RFC

rcantin-w:
Just realised I hadn't pulled in a while and rcf/62 was already taken (so was 63), so switched it to 64, if that's something we care about? Maybe not. We have a few of those doubles in that repo.


rcantin-w:
We don't seem to mind so I'll close this

PR #116, opened 2025-01-06
RFC 066: Graph pipeline

StepanBrychta:
RFC describing development considerations for the graph pipeline. This includes describing:

* Our general approach for loading concepts (and the relationships between them) into the graph
* Our approach for loading source concepts from each concept source (LoC, MeSH, Wikidata)
* Our architecture and implementation plans


Amvla:
Preview link: https://github.com/wellcomecollection/docs/blob/rfc-066-graph-pipeline/rfcs/066-graph_pipeline/README.md

rfcs/066-graph_pipeline/README.md
```diff
@@ -0,0 +1,244 @@
+# RFC 066: Catalogue Graph pipeline
+
+This RFC outlines considerations for the development of the cagalogue-graph pipeline. The first iteration of the graph will be focused on concepts and their enrichment with data from external ontologies, as discussed below.
```
kenoir, line 3: Nit:

```suggestion
This RFC outlines considerations for the development of the catalogue-graph pipeline. The first iteration of the graph will be focused on concepts and their enrichment with data from external ontologies, as discussed below.
```

rfcs/066-graph_pipeline/README.md
```diff
@@ -0,0 +1,244 @@
...
+  graph.
+
+## Architecture
+
+The catalogue graph pipeline will run on a serverless event-driven architecture. All processing will be done by two
+Lambda functions, an `extractor` Lambda function, and an `indexer` Lambda function.
+
+The `extractor` will be responsible for:
+
+1. Communicating with concept sources and extracting a stream of raw source concepts via `Source` classes.
```
kenoir, line 200: Can a single lambda execution carry out a full extraction from the sources? I feel like it may take longer than 15 minutes to scroll through all the LoC concepts for example?

rfcs/066-graph_pipeline/README.md
```diff
@@ -0,0 +1,244 @@
...
+Lambda function. Execution will be orchestrated via AWS Step Functions. Execution order will follow the ordering in the
+diagram above — top to bottom, and nodes before edges.
+
+The full pipeline will be configured to run regularly but on an infrequent schedule (once every few months or similar)
+due to source concepts not being updated frequently. The sub-pipeline processing Wellcome catalogue concepts can be
+scheduled to run more frequently.
+
+## Revised architecture
+
+Adding entities to the graph via the `indexer` Lambda function (as described in the previous section) has performance
```
kenoir, line 241: I wouldn't put this as an after-thought but state is as part of the implementation above and keep this as a note that we explored adding entities in the less performant way, so chose the bulk loader.

rfcs/066-graph_pipeline/README.md
```diff
@@ -0,0 +1,244 @@
...
+   is not a concern as the serverless cluster should automatically scale based on traffic.)
+3. The event source mapping can be configured to insert a query back into the queue in case of error, providing an
+   automatic solution to transient faults. Additionally, queries which repeatedly fail will be placed in the DLQ, where
+   they can be investigated.
+
+The `extractor` could also be separated into several smaller microservices, but this would add a cost and performance
+overhead, as processed items would have to be serialised and deserialised between individual steps.
+
+Each invocation of the `extractor` Lambda function will only execute one Transformer, based on the input passed to the
+Lambda function. Execution will be orchestrated via AWS Step Functions. Execution order will follow the ordering in the
```
kenoir, line 232: Step-functions interactions (IIIRC) lend themselves to being diagrammed nicely, which might be a useful thing to add here.

rfcs/066-graph_pipeline/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+* Each concept source (e.g. MeSH, LoC, Wellcome catalogue) will have a corresponding `Source` class, which will
+  include logic for downloading raw concepts and stream them to subsequent components for processing.
+* Each combination of concept source and concept type (e.g. LoC name, LoC location, MeSH concept) will have a
+  corresponding `Transformer` class, which will include logic for extracting nodes and edges (separately) from each raw
+  concept and converting them into the desired format for indexing into the Neptune database.
+* Transformed concepts will be streamed in batches to the `CypherQueryBuilder`, which will be responsible for creating
+  openCypher upsert queries.
+
+![implementation.png](figures%2Fimplementation.png)
```
kenoir, line 158: This is great.

I wonder if there is  utility in thinking about how we might orchestrate running each step in order here?

rfcs/066-graph_pipeline/README.md
```diff
@@ -0,0 +1,244 @@
...
+ensure that at the end of each pipeline run, the graph will be correct and complete. Not adhering to this ordering
+would result in missing nodes or edges.
+
+`SAME_AS` edges (connecting SourceConcept nodes) will be extracted in two stages. The first stage will involve running a
+SPARQL query against Wikidata to extract all linked LoC and MeSH subjects. The second stage will involve downloading the
+external links file from LoC and extracting all linked Wikidata subjects. The ordering of these two stages does not
+matter, and in the diagram they are both labeled `3`. In the first version of the graph, only one of these stages might
+be implemented.
+
+`HAS_SOURCE_CONCEPT` edges (connecting Concept nodes to SourceConcept nodes) will also be extracted in two stages. In
```
kenoir, line 171: I'm guessing the order here, with `SAME_AS` and `HAS_SOURCE_CONCEPT` being the last stage after loading all the large external datasets allow us to switch to a mode where we listen to catalogue pipeline updates for new concepts?

rfcs/066-graph_pipeline/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+### Library of Congress
+
+`SourceConcept` nodes from LCSH, as well as within-LoC edges will be loaded in full from https://id.loc.gov/download/authorities/subjects.skosrdf.jsonld.gz. `SourceName` nodes from LCNAF (including additional fields for birth date, death date, and birth place), as well as within-LoC edges will be loaded in full from https://id.loc.gov/download/authorities/names.skosrdf.jsonld.gz.
+
+Given that location terms appear in both LCSH and LCNAF, there needs to be some pipeline logic to extract `SourceLocation` nodes from both (for instance, via their grouping into `MADS/RDF Geographic`). This is to avoid cross-links between `SourceConcept` and `SourceName` nodes via Wikidata and MeSH. It may be desirable to create new source nodes for other entities such as organisations in the future, so we can collect relevant metadata on these from Wikidata. For example, organisations are currently part of LCNAF and will therefore be `SourceName` nodes with empty birth/death dates. However, unlike geographic locations, they are not affected by any cross-linking issues and will thus be deprioritised in the first iteration of the graph pipeline.
```
kenoir, line 17: I think some examples would be useful here in grasping the specific differences around locations being discussed here.

rfcs/066-graph_pipeline/README.md
StepanBrychta: That's a great question. 

At the moment, the largest source we are planning to stream in full is the LoC Names dataset, with 10+ million source concepts. I tested running the `extractor` Lambda function on the full dataset and streaming the results into S3 and it took approximately 10 minutes when I allocated 4 GB of memory to the Lambda function. 

So while the answer is technically _yes_, we don't have a lot of room to expand, which is not ideal.

In my testing, I discovered that the bottleneck does not come from Lambda networking (downloading the full file only takes a minute or so) but from the Python processing. With this in mind, I can think of a few relatively simple fixes:
1) We could introduce new arguments to the `extractor` Lambda function, specifying a minimum and a maximum index to process. For example, we could set it up so that each Lambda invocation processes up to a million items — the first invocation would process items 0 to 999,999 and then stop. The second invocation would skip the first million items (by streaming them from the ZIP file but discarding them without processing them, or by utilising HTTP range requests where supported) and then process items 1,000,000 to 1,999,999, etc. Each Lambda invocation could write into a separate S3 file and we could run all of them in parallel.
2) We could split the `extractor` processing into two Lambda functions — one for downloading the raw ZIP and extracting the raw dictionaries, and one for transforming the raw dictionaries into nodes/edges for Neptune bulk upload. (I haven't tested this approach, but I would expect at least a 30% speed-up in both Lambdas.)

Neither of these solutions would involve a major architecture change, which means that we don't necessarily have to choose and implement one of them now. Instead, we could focus on implementing a fix when/if we start getting closer to the 15 minute Lambda execution limit. We are planning to run the `extractor` Lambda very infrequently (once every few months), so even in the worst case where we suddenly exceed the 15 minute limit without anticipating it, we could just run the `extractor` locally once, and then we would have a few months to implement a proper fix before the next run.

rfcs/066-graph_pipeline/README.md
StepanBrychta: Good point, I will rewrite the architecture section to make it more clear that we are planning to use the Neptune bulk loader.

rfcs/066-graph_pipeline/README.md
StepanBrychta: The plan is to use a combination of state machines and Terraform for orchestration.

There is a draft implementation on [this branch](https://github.com/wellcomecollection/catalogue-graph/tree/initial-pipeline-work). Terraform dynamically generates individual steps of the state machines based on a variable stored [here](https://github.com/wellcomecollection/catalogue-graph/blob/initial-pipeline-work/terraform/variables.tf).

rfcs/066-graph_pipeline/README.md
StepanBrychta: Yes, after bulk loading all the source concepts (and Wellcome catalogue concepts/works), we can periodically run an incremental version of the pipeline whenever changes to catalogue concepts/works are detected. 

It wouldn't make sense to use Neptune bulk upload for this purpose, so maybe we could utilise the SNS topic with the `indexer` Lambda function.

rfcs/066-graph_pipeline/README.md
kenoir: Looking [over the infra PR](https://github.com/wellcomecollection/catalogue-graph/pull/4) I can see this is thoroughly covered 👍

rfcs/066-graph_pipeline/README.md
kenoir: > It wouldn't make sense to use Neptune bulk upload for this purpose, so maybe we could utilise the SNS topic with the indexer Lambda function.

👍 this makes sense.

rfcs/066-graph_pipeline/README.md
Amvla: Good idea, I've updated the README with some examples.

rfcs/066-graph_pipeline/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+Instead, if we load source nodes and edges from LoC and MeSH in full, we only need to add an edge between the new concept and the source it is connected to. Source node and edge extraction can happen independently, and their frequency can be scheduled around updates to LoC and MeSH. To make sure there are no dependencies between different sources, the data will be loaded in multiple passes, whereby nodes and within-vocabulary edges will be loaded first, followed by edges between different vocabularies.
+
+### Library of Congress
+
+`SourceConcept` nodes from LCSH, as well as within-LoC edges will be loaded in full from https://id.loc.gov/download/authorities/subjects.skosrdf.jsonld.gz. `SourceName` nodes from LCNAF (including additional fields for birth date, death date, and birth place), as well as within-LoC edges will be loaded in full from https://id.loc.gov/download/authorities/names.skosrdf.jsonld.gz.
```
paul-butcher, line 15: A shortcoming of the skos downloads is that they don't have information to help us resolve compound terms.

Perhaps we can consider changing to use the mads downloads instead as part of this work.

e.g. if you look at [Stomach--Cancer--Pathophysiology](http://id.loc.gov/authorities/subjects/sh2010010359), in the SKOS version, it's atomic.  If you look in MADS, it contains a component list, consisting of Stomach, Cancer, and Pathophysiology:
```
...
"madsrdf:componentList": {
        "@list": [
          {
            "@id": "http://id.loc.gov/authorities/subjects/sh85128254"
          },
          {
            "@id": "http://id.loc.gov/authorities/subjects/sh99004861"
          },
          {
            "@id": "http://id.loc.gov/authorities/subjects/sh00006929"
          }
        ]
      },
...
```

rfcs/066-graph_pipeline/README.md
```diff
@@ -0,0 +1,244 @@
...
+MERGE (a)-[r:REL]-(b)
+```
+Furthermore, it is undesirable in terms of query performance to run undirected queries, as explained [here](https://docs.aws.amazon.com/neptune/latest/userguide/best-practices-opencypher-directed-edges.html).
+
+However, the current graph data model does include some edges which do not have any inherent directionality and should be queryable from both sides (for example, `SAME_AS` edges between two source concepts from different ontologies). To avoid introducing an artificial hierarchy between ontologies, and given that this will not add too much storage overhead, these relationships will be represented with two directed edges.
+
+### List properties
+
+It is not possible to create array type properties in AWS Neptune (see [here](https://docs.aws.amazon.com/neptune/latest/userguide/bulk-load-tutorial-format-opencypher.html) for an overview of possible data types). Any list properties (such as `alternative_labels`) will be concatenated to a string, with a `||` separator between individual items. While this can be seen as a limitation, we may benefit from better query performance resulting from indexing of these strings in the graph database.
```
paul-butcher, line 139: Would it be valuable to have labels as nodes in their own right, rather than as part of a property of a node?

It could offer an opportunity to deal with homonyms, e.g.
```mermaid
graph
    A[n79045467]
    B[no95051310]
    C(John Nash)


    B -->|has label| C

    A -->|has label| D(Nash, John, 1752-1835)
    A -->|has label| C
    B -->|has label| E(Nash, John P., 1915-)
```

rfcs/066-graph_pipeline/README.md
Amvla: Yep, we're actually in the process of switching to the mads file. (I initially thought we could get away with the smaller skos version, but as we're also interested in the component lists we will switch to this to implement as part of the NARROWER_THAN relationship.) I will update the url above.

rfcs/066-graph_pipeline/README.md
Amvla: This is an interesting possibility to consider for the future. I think the plan is to include alternative labels in a label-matching or other more sophisticated algorithms either way, but I'm definitely curious in a comparison e.g. from a query performance perspective, or whether this might improve linkage between sources (although we need to be careful about mixing names and concepts).

rfcs/066-graph_pipeline/README.md
kenoir: It's not entirely clear to me how creating SourceLocation nodes helps avoid cross-linking and ambiguity with Wikidata nodes? I think it's probably my reading that's the issue rather than the description though so not blocking for merge.

kenoir (approved):

rfcs/066-graph_pipeline/README.md
paul-butcher: Thinking about it, homonyms are a pretty niche requirement and would probably be better served by a deliberate consideration of reused terms.

Most labels and alternative labels from the sources will be unique to the concept in question, even for relatively common names like John Nash (because most personal or place names have a disambiguator such as their occupation, country, years alive or active etc.).

Amvla: > Most labels and alternative labels from the sources will be unique to the concept in question

That's interesting... something I've been curious about

PR #117, opened 2025-01-08
Add RFC 066 Prismic API ID casing

davidpmccormick:
☝️

### DIFF
--- a/rfcs/066-prismic-api-ids
+++ b/rfcs/066-prismic-api-ids
@@ -1,0 +1,39 @@
+# RFC 066: Prismic API ID casing
+
+## TL;DR
+- Use kebab-case for Custom type API IDs (plural/singular for reusable/single respectively)
+- Use camelCase for all (Custom type and Slice) field API IDs
+- Use snake_case for Slice API IDs
+
+## Background
+We’d like to make sure our Prismic API IDs are written with consistent casing and that we’re following defaults from Prismic so that we’re going with the grain as far as possible and not having to manually override auto-generated SliceMachine files.
+
+## Custom type API IDs
+These should be kebab-cased and plural when the type is ‘reusable’ as opposed to ‘single’ (e.g. the reusable ‘Exhibition highlight tour’ Custom type has the API ID `exhibition-highlight-tours`, whereas the single ‘Global alert’ Custom type has the API ID `global-alert`).
+
+We have done this fairly consistently although a couple of reusable content types have been given singular API IDs – `card` and `collection-venue`.
+
+Should we try to update these two to `cards` and `collection-venues` or just live with it?
+
+## Field API IDs
+We have a mixture of snake_case and camelCase field API IDs. SliceMachine defaults to snake_case (based on the label), although this is readily overridable in the GUI at the point of adding the field.
+
+We exclusively work with camelCased variables in TypeScript and while snake_cased object properties aren’t a problem (e.g. `data.some_property`), we wouldn’t be able to destructure those properties off the `data` object without it being a linting error:
+
+```
+const { some_property } = data; // <-- linting error
+```
+
+Whereas this isn’t a problem:
+
+```
+const { someProperty } = data; // <-- ✨
+```
+
+So perhaps the preference should be for updating the field names at point of creation to be camelCased to remove the need for an extra renaming variable step in the TypeScript.
+
+## Slice IDs
+Prior to our use of SliceMachine, we gave our Slices kebab-cased IDs. However, SliceMachine requires that Slices have snake_cased IDs (it doesn’t let you override it in the GUI). When we migrated to SliceMachine we updated the Slice IDs in the files that SliceMachine generated to be camelCased in order to be able to use our existing code consistently. But any subsequent change to the Slice through SliceMachine will revert the ID to snake_case and we have to remember to override it back to camelCase. For newer Slices we have kept the default (snake_case) ids rather than override them to camelCase.
+
+I don’t think we deal directly with the Slice IDs. Instead, we send all of them (in [a `components` object](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/slices/index.ts#L5-L29)) along to SliceMachine, so the linting problems mentioned above don’t arise. As such, I propose that we move to using snake_case for all Slice IDs.
+


rcantin-w (commented):
(np? unsure)
Can we put it in a folder so the order doesn't get messed up when other RFCs are created 🙏

rfcs/067-prismic-api-ids.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Custom type API IDs
+These should be kebab-cased and plural when the type is ‘reusable’ as opposed to ‘single’ (e.g. the reusable ‘Exhibition highlight tour’ Custom type has the API ID `exhibition-highlight-tours`, whereas the single ‘Global alert’ Custom type has the API ID `global-alert`).
+
+SliceMachine makes these snake_case by default but allows this to be overridden easily in the GUI - we've done this fairly consistently although a couple of reusable content types have been given singular API IDs – `card` and `collection-venue`. I don't think there's any mileage in updating everything to snake_case since *none* of them are in that format currently. I'm not sure if we should try to update to `cards` and `collection-venues` or just live with those as they are?
+
+## Field API IDs
+We have a mixture of snake_case and camelCase field API IDs. SliceMachine defaults to snake_case (based on the label), although this is readily overridable in the GUI at the point of adding the field.
```
rcantin-w, line 17: Think the acronym would ideally be spelled out fully on first mention (think onboarding etc)

rfcs/067-prismic-api-ids.md
```diff
@@ -0,0 +1,37 @@
...
+- Use camelCase for all (Custom type and Slice) field API IDs
+- Use snake_case for Slice API IDs
+
+## Background
+We’d like to make sure our Prismic API IDs are written with consistent casing and that we’re following defaults from Prismic so that we’re going with the grain as far as possible and not having to manually override auto-generated SliceMachine files.
+
+## Custom type API IDs
+These should be kebab-cased and plural when the type is ‘reusable’ as opposed to ‘single’ (e.g. the reusable ‘Exhibition highlight tour’ Custom type has the API ID `exhibition-highlight-tours`, whereas the single ‘Global alert’ Custom type has the API ID `global-alert`).
+
+SliceMachine makes these snake_case by default but allows this to be overridden easily in the GUI - we've done this fairly consistently although a couple of reusable content types have been given singular API IDs – `card` and `collection-venue`. I don't think there's any mileage in updating everything to snake_case since *none* of them are in that format currently. I'm not sure if we should try to update to `cards` and `collection-venues` or just live with those as they are?
```
rcantin-w, line 14: Think the acronym would ideally be spelled out fully on first mention (think onboarding etc) and then the acronym in parentheses for future reference (e.g. Amazon Web Services (AWS))

rcantin-w, line 14: This addresses both `snake_case` - > `kebab-case` and singular vs plural, is that correct? Should we split them up in two paragraphs for separate conversations?

rfcs/067-prismic-api-ids.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Field API IDs
```
rcantin-w, line 16: What's an example of a Field API? Is it single fields like "title"?

rfcs/067-prismic-api-ids.md
```diff
@@ -0,0 +1,37 @@
...
+const { some_property } = data; // <-- linting error
+```
+
+Whereas this isn’t a problem:
+
+```
+const { someProperty } = data; // <-- ✨
+```
+
+So perhaps the preference should be for updating the field names at point of creation to be camelCased to remove the need for an extra renaming variable step in the TypeScript.
```
rcantin-w, line 31: That could be nice indeed. What about legacy fields?

rfcs/067-prismic-api-ids.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Slice IDs
+Prior to our use of SliceMachine, we gave our Slices kebab-cased IDs. However, SliceMachine requires that Slices have snake_cased IDs (it doesn’t let you override it in the GUI). When we migrated to SliceMachine we updated the Slice IDs in the files that SliceMachine generated to be camelCased in order to be able to use our existing code consistently. But any subsequent change to the Slice through SliceMachine will revert the ID to snake_case and we have to remember to override it back to camelCase. For newer Slices we have kept the default (snake_case) ids rather than override them to camelCase.
+
+I don’t think we deal directly with the Slice IDs. Instead, we send all of them (in [a `components` object](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/slices/index.ts#L5-L29)) along to SliceMachine, so the linting problems mentioned above don’t arise. As such, I propose that we move to using snake_case for all Slice IDs.
```
rcantin-w, line 36: Sounds good to me! Including legacy ones?

rcantin-w (commented):
Thanks for doing this, much needed conversations on there. Feels like details but really it amount to some level of frustration every time we try to do the right thing when creating new fields/modifying them, so I think it's very valuable in the long-term.

rfcs/067-prismic-api-ids.md
davidpmccormick: it is. I'll add one (or two) examples

rfcs/067-prismic-api-ids.md
davidpmccormick: I don't know! The completionist in me would like it all to be neat, but I'm not sure the ends justify the means. I suppose this is what the RFC is for?

rfcs/067-prismic-api-ids.md
davidpmccormick: See [above](https://github.com/wellcomecollection/docs/pull/117#discussion_r1908549467)

kenoir:
Preview link: https://github.com/wellcomecollection/docs/blob/rfc-066-prismic-api-ids/rfcs/067-prismic-api-ids/README.md

rfcs/067-prismic-api-ids/README.md
```diff
@@ -9,12 +9,19 @@
...
 
-SliceMachine makes these snake_case by default but allows this to be overridden easily in the GUI - we've done this fairly consistently although a couple of reusable content types have been given singular API IDs – `card` and `collection-venue`. I don't think there's any mileage in updating everything to snake_case since *none* of them are in that format currently. I'm not sure if we should try to update to `cards` and `collection-venues` or just live with those as they are?
+SliceMachine makes these snake_case by default but allows this to be overridden easily in the Graphical User Interface (GUI) and to-date we have made *all* of them kebab-case, so there probably isn't a good case for changing these to snake_case now.
+
+A couple of reusable content types have been given singular API IDs – `card` and `collection-venue` – should we consider changing these two to `cards` and `collection-venues`?
```
rcantin-w, line 16: If the effort level is low, I think it would be nice, so might be a question of weighing time/effort? With the migration script, is this something we can consider doing easily?

rfcs/067-prismic-api-ids/README.md
```diff
@@ -9,12 +9,19 @@
...
+When we started using Prismic we decided that these should be kebab-cased. Also, if the type was ‘reusable’ it should be plural, if it was 'single' it should be singular. For example, the reusable 'Exhibition highlight tour' type has an API ID of `exhibition-highlight-tours` whereas the single 'Global alert' type has an API ID of `global-alert`.
 
-SliceMachine makes these snake_case by default but allows this to be overridden easily in the GUI - we've done this fairly consistently although a couple of reusable content types have been given singular API IDs – `card` and `collection-venue`. I don't think there's any mileage in updating everything to snake_case since *none* of them are in that format currently. I'm not sure if we should try to update to `cards` and `collection-venues` or just live with those as they are?
+SliceMachine makes these snake_case by default but allows this to be overridden easily in the Graphical User Interface (GUI) and to-date we have made *all* of them kebab-case, so there probably isn't a good case for changing these to snake_case now.
```
rcantin-w, line 14: 👍 
Let's consider this a documented decision and keep an eye should new content be created to ensure it's kebab-cased!

rfcs/067-prismic-api-ids.md
rcantin-w: I think again it's a question of how easy it is. Is it low effort and high satisfaction? etc.

rfcs/067-prismic-api-ids/README.md
```diff
@@ -0,0 +1,46 @@
...
+const { some_property } = data; // <-- linting error
+```
+
+Whereas this isn’t a problem:
+
+```
+const { someProperty } = data; // <-- ✨
+```
+
+So perhaps the preference should be for updating the field names at point of creation to be camelCased to remove the need for an extra renaming variable step in the TypeScript.
```
rcantin-w, line 38: I feel like since it's "readily overridable in the GUI at the point of adding the field.", we should enforce camelCase for sure, again, as a documented agreed upon decision.

Could we also consider aligning existing ones with the migration script?

davidpmccormick: I think we should consider it – I added a new vs legacy para at the end to wave my hands in the air about whether or not we want to do it based on cost/benefit

rfcs/067-prismic-api-ids/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+I don’t think we deal directly with the Slice IDs. Instead, we send all of them (in [a `components` object](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/slices/index.ts#L5-L29)) along to SliceMachine, so the linting problems mentioned above don’t arise. As such, I propose that we move to using snake_case for all Slice IDs.
```
rcantin-w, line 43: Agreed. Since it gets overwritten every time, we don't know to revert camelcasing anywhere do we?

rfcs/067-prismic-api-ids/README.md
```diff
@@ -0,0 +1,46 @@
+# RFC 067: Prismic API ID casing
+
+## TL;DR
+- Use kebab-case for Custom type API IDs (plural/singular for reusable/single respectively)
+- Use camelCase for all (Custom type and Slice) field API IDs
+- Use snake_case for Slice API IDs
```
rcantin-w, line 6: maybe add here what needs to be changed manually and which ones are default?

rfcs/067-prismic-api-ids/README.md
rcantin-w: Actually, quick thought, what about changing the decision to camelCase to reduce the amount of different casings we use?

rfcs/067-prismic-api-ids/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+I don’t think we deal directly with the Slice IDs. Instead, we send all of them (in [a `components` object](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/slices/index.ts#L5-L29)) along to SliceMachine, so the linting problems mentioned above don’t arise. As such, I propose that we move to using snake_case for all Slice IDs.
+
+## New vs. legacy
+Whatever pattern we decide to go with in future, we also need to decide whether we should apply it to legacy content. Doing it for all content old and new would obviously be good for consistency but needs to be weighed against the effort required and risks associated with migrating the content.
```
rcantin-w, line 46: I'd suggest adding a spike/discovery ticket in the next sprint to see how much effort we think is required.

rfcs/067-prismic-api-ids/README.md
davidpmccormick: …or snake_case so that they're aligned with what we're going to (have to) go for with the Slice API IDs?

rcantin-w: Are these IDs used in our code/will cause TS pain?

rfcs/067-prismic-api-ids/README.md
gestchild: Creating the new (pluralised) types and migrating the content should be straightforward. I think the riskier part would be finding where the old (singular types) are linked from and updating those too. e.g. where a card is used on a page, we'd need to link it to the new version of the card

rfcs/067-prismic-api-ids/README.md
davidpmccormick: They aren't used as TS variables, no (they are currently kebab-case and that would already have been a problem for use in JS). They can only be used in quotes when querying Prismic.

rfcs/067-prismic-api-ids/README.md
davidpmccormick: I'll document this.

rfcs/067-prismic-api-ids/README.md
rcantin-w: Right! So snake_case works too 👍

rfcs/067-prismic-api-ids/README.md
rcantin-w: Migration doesn't allow for overwriting, right? We always have to re-create a new document?

rcantin-w: Nice, that'd be great if it worked for IDs

rcantin-w (approved):

PR #118, opened 2025-02-11
RFC 068: Exhibitions in Events index; Content API

rcantin-w:
## What does this change?

https://github.com/wellcomecollection/content-api/issues/208

Read it here: https://github.com/wellcomecollection/docs/blob/68-exhibitions-content-api/rfcs/068-exhibitions-content-api/README.md

Starting the discussion on how we want Exhibitions to be indexed/returned in view of Events & Exhibitions search work.

I have not mocked up the API responses in the end as I determined indexing exhibitions separately should be a different piece of work for the future, and the current `/events` response should not change.

## How to test

Read up and comment!


rfcs/068-exhibitions-content-api/README.md
```diff
@@ -0,0 +1,58 @@
+# RFC 068: Exhibitions in Content API
+Exhibitions are to be added to Events search, becoming Events & Exhibitions search. We'll therefore be working on indexing Exhibitions in a more intentional manner. 
+That indexing and subsquent API endpoint will power the Events & Exhibitions search as well as, eventually, the existing listing pages.
+
+## Indexing Exhibitions
+I can think of two ways to go about indexing Exhibitions for ranking against Events to make sense and be easier to handle.
```
kenoir, line 6: I think there's a 3rd possibility: add exhibitions to events search by adding exhibitions as a new format to the events index with a subset of fields tailored to searching as an event AND index exhibitions as in a new index for exhibitions listing pages where more fields and different search mappings are required.

This way you keep the /events endpoint, add an /exhibitions endpoint and the amount of work required is similar as in both cases you need to decide which subset of exhibition fields need to be searchable with events, with the  advantage that you don't need to pollute your exhibitions index with the concern that it also has to be searchable as an event.

rcantin-w: Does this suggest that events listing pages are rendered with the /events endpoint filtering out exhibition format? (so kind of a midway between both solutions in that one endpoint still filters out but not the other?)

kenoir: > Does this suggest that events listing pages are rendered with the /events endpoint filtering out exhibition format?

Yes I think so.

kenoir: So the relationship is something like, all Exhibitions are Events, not all Events are Exhibitions - what shows on /events is a subset of all events, specifically excluding exhibitions. /exhibitions is only exhibitions.

rcantin-w: OK! I don't mind that, especially as is it's easy to remove them should our needs completely change and that currently it _is_ the logic in search. Curious for @gestchild's opinion if he can think of any scenario I can't that this wouldn't cover, but otherwise will move forward with this suggestion.

gestchild: This sounds good to me ✅

rcantin-w: I've documented our discussion from this morning @kenoir, https://github.com/wellcomecollection/docs/pull/118/files#diff-eb58d89e102638d8dcda788407322d9e31c3fa86938afffd79ed1430547b31a7R24

rfcs/068-exhibitions-content-api/README.md
```diff
@@ -0,0 +1,88 @@
...
+```
+  display: {
+    format: {
+      type: 'EventFormat',
+      id: "Xk5QHxEAACUAe4gV",
+      label?: "Season"
+    };
+  },
+  query: {
+    format: ['Exhibition', 'Season'] // This way it's queryable under both identifiers
```
rcantin-w, line 51: @paul-butcher I believe this would match your suggestion?

rcantin-w: (said yes in person)

PR #119, opened 2025-02-13
RFC 069: Catalogue Graph Ingestor

kenoir:
## What does this change?

Part of https://github.com/wellcomecollection/platform/issues/5913

This change proposes an implementation for a service that ingests concepts data from the catalogue-graph in a way that allows us to replace the existing concepts-pipeline.

## How to test

- [ ] [Read the RFC](https://github.com/wellcomecollection/docs/blob/rk/rfc-68-grapg-ingestor/rfcs/069-catalogue_graph_ingestor/README.md), does it make sense?

Clear direction for implementation of next steps for catalogue-graph.


rfcs/069-catalogue_graph_ingestor/README.md
```diff
@@ -0,0 +1,185 @@
...
+{
+  "id": "avkn7rq3",
+  "identifiers": [
+    {
+      "identifierType": {
+        "id": "nlm-mesh",
+        "label": "Medical Subject Headings (MeSH) identifier",
+        "type": "IdentifierType"
+      },
+      "value": "D001255",
```
Amvla, line 49: There is a very small number of `Person` concepts in the catalogue with a reference to [viaf](https://viaf.org/en) or [fihrist](https://www.fihrist.org.uk/). These name authorities are currently not in the graph. We could add these as `SourceName`, but this wasn't a priority initially because there are very few of them and we don't get any onward links or descriptions from the sources. However, in the context of reproducing the response from the Concepts API, it means that we currently don't have the source identifiers for these concepts in the graph and may need to add these in.

Amvla (approved):
Great overview!

rfcs/069-catalogue_graph_ingestor/README.md
StepanBrychta: Good point, we might want to add them for completeness. In the meantime, we could index `viaf` and `fihrist`-derived concepts into the ES index as if they were label-derived, and the corresponding concept pages should look and behave in the same way they do now.

StepanBrychta (approved):

rfcs/069-catalogue_graph_ingestor/README.md
kenoir: I've incorporated these comments into the RFC for completeness when we implement.

PR #120, opened 2025-02-26
New concepts API RFC (#5934)

StepanBrychta:
## What does this change?

Preview [here](https://github.com/wellcomecollection/docs/tree/070-new-concepts-api/rfcs/070-concepts-api-changes).


rfcs/070-concepts-api-changes/README.md
```diff
@@ -0,0 +1,324 @@
...
+Some label-derived concepts are currently linked to multiple source concepts from the same source ontology via `HAS_SOURCE_CONCEPT` edges. 
+For example, the label-derived concept `Cleanliness` is connected to three separate LoC subject headings — `Sanitation`, `Hygiene`, and `Baths`.
+All of these connections were created by matching the label of the label-derived concept to alternative labels of the LoC source concepts.
+This and similar connections can pose several issues:
+  * By linking all of these concepts together, we would be merging the concept pages for 'Sanitation', 'Hygiene', and 'Baths'
+    into a single page. 
+  * In cases where a concept is connected to a given source concept **and** the parent of the source concept, we would end up
+    listing a given concept as a related concept on its own theme page (i.e. a concept would be related to, or part of, itself).
+
+To avoid these issues, we should limit ourselves to creating at most one `HAS_SOURCE_CONCEPT` node from a label-derived concept node.
```
kenoir, line 315: I would favour limiting to a most one `HAS_SOURCE_CONCEPT` _within a known ontology_. This is to allow matching across MeSH & LoC, but reduce the risk of a concept being connected to itself through a proliferation of LoC label / alternative label to a label derived concept though a text match.

rfcs/070-concepts-api-changes/README.md
```diff
@@ -0,0 +1,324 @@
...
+
+## Merging concepts across sources
+
+Due to storing `HAS_SOURCE_CONCEPT` edges between catalogue concepts and source concepts, and `SAME_AS` edges
+between source concepts, the catalogue graph allows us to 'merge' identical concepts into a single unified theme page.
+For example, if a work references a LoC subject heading `Sanitation` and a different work references a MeSH term `Sanitation`,
+both of these works would link to (and be listed on) the same `Sanitation` theme page.
+
+To enable this functionality, all 'merged' concepts will be stored in the Elasticsearch index as a single document,
+which will include all of their canonical IDs. This is similar to the `sameAs` field
```
kenoir, line 150: It may be preferable to keep seperate elasticsearch documents for each concept ID; doing so would allow us to deal with the lowered reliability of "correct" relationships to other works through concepts inherent with label-derived concepts. 

Put another way: If you arrive at a label derived concept page, the actual meaning is necessarily ambiguous so it makes sense the works on that page might have a broader / less accurate semantic relationship. However if you arrive at a MeSH derived concept page, the meaning is quite specific and the connection to other works through a label-derived concept match is less useful.

kenoir: We may want to implement a difference in the way these pages are constructed, either through how attributes are merged from a related concept onto the indexed document, or via how queries for works and images based on matched concepts in these pages are weighted.

For example on an explicitly matched (NOT label derived) "LoC" concept page, we may want to reduce the visibility of works that are matched through a label-derived concept. You can imagine writing an elasticsearch query which weights matches to works with label derived concepts as less relevant than matches through concepts in a known ontology.

rfcs/070-concepts-api-changes/README.md
```diff
@@ -0,0 +1,324 @@
...
+| yz3xs9c9, cqpmracr                     | Communicable Disease Control |
+| zc9y7m45, kx79h6jm, xcdy58xp, mz9syutk | Sanitary Engineering         |
+
+Putting everything together, we can construct a full JSON response returned to the front-end:
+
+```
+GET /concepts/eva7r2dw
+{
+  // The Wellcome IDs of all 'merged' concepts are included
+  "id": ["eva7r2dw", "d8z89dv6", "mwyumfq2", "jubdg55b"],
```
kenoir, line 209: If we choose to represent "semantic cognates" (?) i.e. concepts that share a common meaning, as separate documents, we'll need an attribute that represents those matches here, similar to the "sameAs" field.

For example:

```json
{
  "id": "vk6nj4mt",
  "identifiers": [
    {
      "identifierType": {
        "id"   : "label-derived"                                    ,
        "label": "Identifier derived from the label of the referent",
        "type" : "IdentifierType"
      },
      "value": "church of england",
      "type": "Identifier"
    }
  ],
  "label": "Church of England",
  "alternativeLabels": [],
  "type": "Organisation",
  "matchedConcepts": [
    {
      "id": "d8z89dv6",
      "matchType": {"id": "label"},
      "identifiers": [
        {
          "identifierType": "lc-subjects",
          "value"         : "sh85117296" ,
          "type"          : "Identifier"
        }
      ]
    }
  ]
}

```

rfcs/070-concepts-api-changes/README.md
```diff
@@ -0,0 +1,324 @@
...
+      "type": "Identifier"
+    },
+    {
+      "identifierType": "wikidata",
+      "value": "Q949149",
+      "type": "Identifier"
+    }
+  ],
+   
+  // The MeSH label takes priority over any other labels
```
kenoir, line 236: If we do not merge concepts into one document, I would still expect us to try to enhance concept pages by merging fields from matched concepts, using a hierarchy as we do in the works merger.

rfcs/070-concepts-api-changes/README.md
```diff
@@ -0,0 +1,324 @@
...
...similarTo         | Can be obtained from the graph by filtering for people who are linked to a given field of work and whose birth date is in a given range. In some cases many people might be returned and choosing the most 'notable' ones might be difficult. One approach would be to determine notability based on how many works reference a given person as a concept (i.e. counting the number of edges).                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
```
rcantin-w, line 62: >  In some cases many people might be returned and choosing the most 'notable' ones might be difficult. One approach would be to determine notability based on how many works reference a given person as a concept (i.e. counting the number of edges).

Just yesterday, we offered up this section ("Notable people") to users as part of our design sprint prototype, and the use of the term "Notable" raised the point of "are we as an organisation stating facts OR are we weighing in on who WE consider to be notable?". And like you said that's a pretty difficult thing to do.

I thought it was interesting to raise here, I don't really a suggestion or an idea to offer, but maybe this section needs to be re-explored.

rfcs/070-concepts-api-changes/README.md
```diff
@@ -0,0 +1,324 @@
...
+The screenshot below shows all concept metadata exposed on this page type:
+
+> <img src="img/example_subject_theme_page.png" alt="example_subject_theme_page" width="600"/>
+
+This page only has one additional piece of information not shown on other theme pages (refer to tables above for all other fields):
+
+
+| Field name   | Details                                                                            |
+|--------------|------------------------------------------------------------------------------------|
+| narrowerThan | Stored in the graph as SourceConceptNarrowerThan and SourceConceptHasParent edges. |
```
rcantin-w, line 94: `narrowerThan` is not in the image, is it `partOf`?

rfcs/070-concepts-api-changes/README.md
```diff
@@ -0,0 +1,324 @@
...
+> <img src="img/works_tabs_concept.png" alt="works_tabs_concept" width="600"/>
+
+To retrieve these works from the catalogue API, we need the *single concept endpoint* to return a list of relevant narrower topics
+(in addition to the fields listed in the tables above):
+
+| Field name  | Details                                                                                                                                                                                                                                          |
+|-------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| broaderThan | Stored in the graph as SourceConceptNarrowerThan (edge direction reversed) and SourceConceptHasParent (edge direction reversed) edges. |
+
+We can then make separate API calls to the catalogue API to retrieve a list of works for each sub topic.
```
rcantin-w, line 128: I am interested in this; in `/search`, we also have tabs and in all designs, each tab has the X of results as a suffix. Because this means having to do multiple queries on every keyword change (and ignoring the added difficulty of filtering), we've always pushed back on it. I'd be interested in seeing how it's dealt with on your end and will keep an eye out 👀

rfcs/070-concepts-api-changes/README.md
rcantin-w, line 209: This is me not having been aware of concepts work for a bit so forgive; does it mean if any of these ids are accessed in URL, they return the same result page?

rcantin-w (commented):
This was a really good read, as someone who hasn't focused on this in a long time I found it very easy to follow, very clear. Thank you!

rfcs/070-concepts-api-changes/README.md
StepanBrychta: You're right, those two should match. I updated the image to say `narrowerThan`.

rfcs/070-concepts-api-changes/README.md
```diff
@@ -0,0 +1,324 @@
...
+  ],
+   
+  // The MeSH label takes priority over any other labels
+  "label": "Sanitation",
+  
+  // All labels from all source concepts can be included (though they might not be needed).
+  // Even if we do not include them in the API response, they should still be indexed in case we
+  // want to use them in the future for querying.
+  "alternativeLabels": [
+    "public sanitation",
```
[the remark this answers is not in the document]
StepanBrychta: At the moment we aren't planning to show alternative labels to users. The example response includes them just because the current API returns them too. But capitalising them for consistency is probably still a good idea!

rfcs/070-concepts-api-changes/README.md
StepanBrychta: That's the idea, yes. But we've since changed the approach (see [this comment](https://github.com/wellcomecollection/docs/pull/120#discussion_r1983368364)).

I updated the 'Connecting concepts across sources' section and the API response example section to reflect the new approach.

In general, different concept pages corresponding to 'linked' concepts would still show the same works/images, but there is an exception when it comes to label-derived concepts which link to more than one source concept.

rfcs/070-concepts-api-changes/README.md
StepanBrychta: Good point. We haven't yet agreed on what approach to use when deciding which related themes/people to surface in cases where there are many to choose from. I listed a few options in the RFC, but we will definitely need to revisit this.

kenoir (approved):
The added examples are great 👍

PR #121, opened 2025-04-08
Python build and deploy

paul-butcher:
## What does this change?

A plan for addressing inconsistencies in Python build systems: https://github.com/wellcomecollection/platform/issues/5955

## How to test

Read it, does it make sense?  Implement the recommendations, does it simplify things?

Once implemented, it will be easier to manage Python projects across all our repositories.

The point of writing this is to mitigate the risk of adding yet another build tool and/or process.

### DIFF
--- a/rfcs/070-python_builds/README.md
+++ b/rfcs/070-python_builds/README.md
@@ -11,7 +11,8 @@ Each time, we have written a new build and deployment process for them. For the 
 is born with has stuck, despite us finding better tools or methods when working on later projects. As such, we have
 ended up with multiple diverse mechanisms to check, build and deploy them.  Each with different standards and rules.
 
-This RFC proposes to harmonise those projects.
+This RFC proposes to harmonise those projects.  The projects are listed in 
+the [Appendix](appendix.md), alongside some description of their differences.
 
 ### Problems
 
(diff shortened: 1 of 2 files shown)


rfcs/070-python_builds/README.md
```diff
@@ -0,0 +1,159 @@
...
+Although each project will have its own pyproject.toml files, there can also be 
+some commonality.
+
+### Common Github Actions
+
+Three Actions are to be defined in the .github repository:
+
+#### Check
+A "Python Check" Github Action in the .github repository
+
```
agnesgaroux, line 117: Can we separate the check and test in different actions? Are there cases where we'd want to run the tests but not the checks, eg. merging to main?

rfcs/070-python_builds/README.md
```diff
@@ -0,0 +1,159 @@
...
+### Converting existing projects
+
+We should gradually migrate existing projects to the new common approach, starting with the 
+[Catalogue Pipeline inferrers](https://github.com/wellcomecollection/catalogue-pipeline/tree/92727715888204ca82b86cc0fbf478e5ca46f2dc/pipeline/inferrer).
+
+Migrating the pipeline inferrers should exercise most of the challenges outlined above, which will
+demonstrate the value of the approach and simplify any subsequent migrations.
+
+Initially, the shared github actions can be defined in the catalogue pipeline project. There is sufficient
```
agnesgaroux, line 146: I think it'd be good to move things to `.github` sooner rather than later

agnesgaroux (commented):
I'm not super familiar with the python ecosystem to have an opinion about things like dependencies management but this plan sounds sensible 👍 
A lo of the github actions we currently use still rely on scripts that are scoped to the particular project the action runs in. These scripts are very similar in a lot of places, are we going to try and extract them into something shareable too?

rfcs/070-python_builds/README.md
paul-butcher: We could do, but I think that adds the possibility of something being accidentally omitted, and requires extra configuration.  

I don't think there are any situations where we would want CI to do one but not the other.  There are still opportunities for style checking to fail on merge the same as tests, particularly if code is shared between projects with and without type checking.

I'm quite keen on the majority of CI stuff being hidden away.  My dream scenario would be for any project's workflow not to say much more than "build it", and all the complexity of what that means is defined elsewhere.

paul-butcher:
> I'm not super familiar with the python ecosystem to have an opinion about things like dependencies management but this plan sounds sensible 👍 A lot of the github actions we currently use still rely on scripts that are scoped to the particular project the action runs in. These scripts are very similar in a lot of places, are we going to try and extract them into something shareable too?

Ideally, yes. There is some conflict between the way some tooling resolves configuration (i.e. it looks in the project directory or the user's home), and the hassle involved in you or I keeping our own workspaces up to date with some scripts shared in a separate repository.  This unfortunately means we have to tolerate some duplication across projects, but I'm very keen to eliminate any duplication in builds as part of all this.

PR #122, opened 2025-06-02
RFC Clean-up

kenoir:
## What does this change?

Preparatory work as part of https://github.com/wellcomecollection/platform/issues/6016

The RFCs folder contained duplicate numbers, unnecessary status', inconsistent last modified and an out of date summary.

This change normalises the RFC starting sections, adds last modified dates using `git` and removes reference to status which is not useful and uniformly out of date.

Where simple summaries were not included they have been generated using `GPT 4.1`.

We have included an updated summary, sorted by last modified date for easier navigation and a script to regenerate as required.

In addition we've added some description of what an RFC should look like.

## How to test

- [Review the change](https://github.com/wellcomecollection/docs/blob/rk/rfc-cleanup/rfcs/README.md), does it make sense?

More consistent, easier to navigate RFCs.

This may break some external links as RFCs are re-assigned numbers where duplicates occurred.

### DIFF
--- a/rfcs/README.md
+++ b/rfcs/README.md
@@ -2,6 +2,12 @@
 
 An RFC is a place to discuss possible changes to the Wellcome Collection platform.
 
+## Table of contents
+
+- [What is an RFC?](#what-is-an-rfc)
+- [How do I format an RFC?](#how-do-i-format-an-rfc)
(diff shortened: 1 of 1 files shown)


agnesgaroux (approved):
Love a good clean-up! 👍

PR #145, opened 2026-04-14
RFC 086: Item viewer refactoring plan

rcantin-w:
## What does this change?
https://github.com/wellcomecollection/wellcomecollection.org/issues/9816

Originally in https://github.com/wellcomecollection/wellcomecollection.org/pull/12937 but David mentioned RFCs and I realised maybe this is where it should live instead. 

The Item Viewer has had a long life and is becoming more and more complex as we go. We need to consider
- Audio works
- Video works
- Image works
- PDF works
- Mixed media works
- Born Digital works

And then:
- No Javascript
- Restricted access works 
- Sensitive materials

And more!
So we thought it'd be nice to refactor it fully, focusing on readability, shared data (using context more for example), and maybe splitting the component further based on what should display/not. It's quite a complex piece of work and as it stands, because of the complexity of IIIF manifests, only Gareth feels comfortable building upon it. David and I thought we'd pair on the refactor to increase our knowledge and comfort level with it, and it starts with a plan. This plan was originally written by Claude after following some of my instructions. I then went through it, questioned it, rejigged it, added considerations extensively... I'm sure there is more to amend/consider/take on, so comments v welcome!

[Read it here](https://github.com/wellcomecollection/docs/blob/rfc-086-item-viewer/rfcs/086-item-viewer-refactor/README.md).

### DIFF
--- a/rfcs/086-item-viewer-refactor/01-overview.md
+++ b/rfcs/086-item-viewer-refactor/01-overview.md
@@ -1,0 +1,94 @@
+# Overview
+
+[← Back to Index](./README.md)
+
+## Problem Statement
+
+The IIIF Viewer component tree has significant duplication of logic and calculations across multiple components. Values like `currentCanvas`, download options, image services, and derived state are computed independently in different files, leading to:
+
+- Code duplication (~60% of calculations repeated)
+- Harder maintenance (changes needed in multiple places)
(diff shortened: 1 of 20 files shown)


rcantin-w:
This was approved by @gestchild and @davidpmccormick so I'll merge.

PR #146, opened 2026-05-13
RFC 087: kiosk mode for wellcomecollection.org

rcantin-w:
- For https://github.com/wellcomecollection/wellcomecollection.org/issues/13074

> We will be offering two experiences in the museum via iPads. We want to use the website for those as it means content stays up-to-date and our styles and components can be used, removing the need to build something new and having to maintain it. This iPad experience will display the website in what we'll call "Kiosk mode".

This RFC was written in an hour, so feel free to poke holes. We're very much wanting to start building prototypes but them but I want us to have thought through a few different things first, especially how we recognise the website is in kiosk mode/not.

[Enjoy](https://github.com/wellcomecollection/docs/blob/087-kiosk-mode/rfcs/087-kiosk-mode/README.md).

### DIFF
--- a/rfcs/087-kiosk-mode/README.md
+++ b/rfcs/087-kiosk-mode/README.md
@@ -16,8 +16,8 @@ Lauren has created Notion pages with more information about these experiences, I
 - HotJar survey to be added conditionally to Kiosk mode
 
 ## Exhibition-specific curated content
-[Notion documentation: kick-off](https://www.notion.so/wellcometrust/T-R-in-exhibition-digital-engagement-kick-off-3576687658a1807b9fb2c88f221bb6be)
-[Notion documentation](https://www.notion.so/wellcometrust/Digital-engagement-in-Tenderness-Rage-35e6687658a18097aa32dae82b7508f0)
+- [Notion documentation: kick-off](https://www.notion.so/wellcometrust/T-R-in-exhibition-digital-engagement-kick-off-3576687658a1807b9fb2c88f221bb6be)
+- [Notion documentation](https://www.notion.so/wellcometrust/Digital-engagement-in-Tenderness-Rage-35e6687658a18097aa32dae82b7508f0)
 
 ### Dev work considerations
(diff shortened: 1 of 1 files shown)


kenoir:
My vote is option 3 — the cookie with a structured value, not a classic toggle.

The deciding factor is requirement 2 (device ID in GA). A plain toggle is boolean-only, so it can't carry the device identifier — you'd need a second mechanism alongside it, adding complexity for no gain. URL params are fragile across page navigation (the RFC already flags this concern with query param stripping on search pages).

A structured cookie like `WC_kioskMode={"device":"RR-Ipad1"}` solves all four requirements in a single mechanism:

- Kiosk mode detection: cookie exists _then_ kiosk mode is on.
- Device ID for analytics: read directly from the cookie value into the dataLayer.
- Cookie banner / consent: handled the same way a toggle would.
- Conditional HotJar: same as above.

It also steps away from implying standard toggle behaviour (which is boolean and well-understood), giving the Dash a clear signal that this is something different that needs its own UI. And if future needs arise (e.g. which exhibition, content warning config), the structured value accommodates them without requiring new infrastructure.

The existing toggle system's cookie plumbing can likely be reused to get there, so the implementation cost over a plain toggle should be modest.

davidpmccormick:
My gut reaction was we could accomplish what we need to with existing toggles alone. They are cheap to create and e.g. `toggle_iPad1` and `toggle_iPad2` could readily be pressed into service

```ts
${iPad1 || iPad2 && <p>kiosk stuff</p> }
```

I suppose you could say its slightly hacky (only really because of 'work in progress' semantics), but if we don't _currently_ need to do anything else (or have any concrete plans to do so) is it worth the additional work for a future that might not exist?

But it is possible that my boundaries between pragmatism and laziness are quite blurry, and I certainly wouldn't object to building a more stateful system.

davidpmccormick (approved):

rcantin-w:
I think because of the fact I've wanted to try this before, I'm tempted to try the new style of cookie out. We can pause before using it a second time and see how useful/efficient we've actually found it to be and revert back to tried and tested ways of doing it if we think it's the right decision then. In the meantime it feels like the time to try it.

If the need for the ID goes away, and we hold nothing else in that cookie, then we can also revert back to regular toggle BUT create a new type. We currently have them as "experimental" and "permanent". Maybe we create a third type and they still have their own section in the Dash, like A/B tests.

PR #153, opened 2026-06-15
RFC 088: Migrating identity, requesting and items APIs from Sierra to FOLIO

kenoir:
## Preview

View the rendered RFC on this branch:

- [RFC 088 README](https://github.com/wellcomecollection/docs/blob/rk/identity-requesting-rfc/rfcs/088-folio-identity-requesting-migration/README.md)
- [Rendered API contract (openapi.md)](https://github.com/wellcomecollection/docs/blob/rk/identity-requesting-rfc/rfcs/088-folio-identity-requesting-migration/openapi.md) · [openapi.yaml](https://github.com/wellcomecollection/docs/blob/rk/identity-requesting-rfc/rfcs/088-folio-identity-requesting-migration/openapi.yaml)

---

## What does this change?

Adds **RFC 088: Migrating identity, requesting and items APIs from Sierra to FOLIO**, plus its supporting API contract and docs tooling. This is documentation only; there are no production code changes.

The RFC describes how we move the identity, requesting and item-availability APIs that power `wellcomecollection.org` from our current LMS (**Sierra**) to its replacement (**FOLIO**). It covers:

- **Architecture.** A parallel, FOLIO-backed **v2** identity API fronted by Auth0, built as Python Lambdas behind API Gateway (user + m2m Lambda authorizers validating Auth0 JWTs). The website's identity webapp BFF is the only production consumer. v2 is stood up alongside v1 (versioned by hostname) and accepts the existing v1 Auth0 audience so already-issued tokens keep working.
- **API surface: v1 → v2.** Route-by-route dispositions for the identity/requesting/item-availability endpoints, a condensed Sierra → FOLIO endpoint/module mapping, and the embedded API contract (linked to `openapi.yaml` / `openapi.md`).
- **Migration plan.** A per-request website toggle (reusing existing server-side API-base-URL/key toggle precedent, with cookie overrides and no redeploy), lazy patron migration within the existing Auth0 tenant (Auth0 captures password hashes at login while Sierra is still available), and a single coordinated cutover window in which identity, requesting and the LMS operational move happen together. The guiding principle is to change one variable at a time: the implementation changes (Sierra → FOLIO), the contract does not.
- **Open questions** to resolve before cutover: identifier translation for requesting; requesting business rules vs FOLIO circulation rules; Auth0 account removal after a FOLIO-side deletion; barcode/role format; and running the new items API alongside the existing v2 catalogue API items endpoint (with a candidate direction of a v3 catalogue API serving only the new FOLIO-backed items endpoint).

The RFC is written to be self-contained and openly accessible, reproducing the architecture, contract, migration plan and open questions in full so it does not depend on the closed discovery/prototype repository.

Files added:

- `rfcs/088-folio-identity-requesting-migration/README.md`: the RFC document.
- `rfcs/README.md`: refreshed RFC listing table (RFC 088 row added).
- `rfcs/088-folio-identity-requesting-migration/openapi.yaml`: the extracted OpenAPI 3.1 spec for the v2 identity API (users / m2m / items routes only).
- `rfcs/088-folio-identity-requesting-migration/openapi.md`: generated human-readable rendering of the spec, browsable on GitHub without a Swagger/Redoc renderer.
- `render_docs.py`, `pyproject.toml`, `.python-version`, `.gitignore`, `uv.lock`: a self-contained `uv` project that validates `openapi.yaml` and regenerates `openapi.md`.

## How to test

- Read `rfcs/088-folio-identity-requesting-migration/README.md` and review the architecture, v1 → v2 surface, migration plan and open questions.
- (Optional) Regenerate the rendered contract and confirm no diff: from the RFC directory, `uv run python render_docs.py`: this validates `openapi.yaml` against the OpenAPI spec validator and rewrites `openapi.md`.

No measurable runtime success criteria; this is a documentation RFC. Success is the RFC being reviewed and providing a clear, self-contained plan and API contract that the team can align on ahead of the Sierra → FOLIO cutover.

- The migration risks themselves (lazy patron migration, the single coordinated cutover, identifier translation, requesting business rules, account-deletion propagation, items-API parallel run) are enumerated in the RFC's Risks and Open questions sections and are intended to be the subject of review.
- The OpenAPI spec is validated and the rendered Markdown is generated from it, reducing the chance of the contract and its human-readable rendering drifting apart.


rfcs/088-folio-identity-requesting-migration/README.md
```diff
@@ -0,0 +1,616 @@
...
+- `username` carries the Auth0 user_id;
+- `externalSystemId` carries the email (or, for a not-yet-linked migrated patron, the legacy patron
+  id);
+- the FOLIO user **UUID** is the internal join key (surfaced to the app as the
+  `folio_user_id` claim) and never appears in the public contract.
+
+A patron's FOLIO `active` flag mirrors **identity completeness**: a patron is active only once they
+have a real name *and* a verified email. New patrons are created **inactive**, carrying v1's literal
+placeholder names (`Auth0_Registration_tempFirstName` / `…tempLastName`) until registration
+completes; activation is reconciled on every login.
```
agnesgaroux, line 181: Is it worth expending on 
- minimal Folio patron 
- what gets sync'd into auth0 

Also is the value of `externalSystemId` firmly decided on?

rfcs/088-folio-identity-requesting-migration/README.md
```diff
@@ -0,0 +1,616 @@
...
+`auth0|p11215550`) and no `folio_user_id`. The enrichment call matches the FOLIO record by
+`externalSystemId`, confirms the email, sets the FOLIO `username` to the Auth0 user_id, normalises
+`externalSystemId` to the email, and writes the FOLIO UUID back to Auth0 metadata so later logins
+take the fast path. No FOLIO custom fields are involved.
+
+Because FOLIO cannot verify passwords, **Auth0 captures the password hash at login while Sierra is
+still available** (the custom-database connection's login script). This is why the public cutover
+opens a credential-capture window (see the migration plan).
+
+```mermaid
```
agnesgaroux, line 275: Isn't Sierra involved in the lazy migration? should it appear in the diagram?

rfcs/088-folio-identity-requesting-migration/README.md
```diff
@@ -0,0 +1,616 @@
...
+```
+
+### Browsing and requesting items
+
+Requestability is decided in **two stages** because no single source has the whole answer:
+
+1. **Catalogue pre-filter (cheap, API-key only).** `/items` returns an instance's items with a
+   `requestable` flag computed from catalogue data alone: the item has **no open request** (we never
+   queue holds, so any existing page/hold/recall makes it unavailable), is **not suppressed**, and
+   has a **circulating status**. This needs no user identity.
```
agnesgaroux, line 316: Doesn't that diverge from the existing items api response? 
The works page needs the location's access status/method but currently doesn't use a `requestable` flag to decide whether to display the request item button

rfcs/088-folio-identity-requesting-migration/README.md
StepanBrychta, line 181: If first name and last name are required fields which can only be set once, would it make sense to require users to specify them during the registration step instead of collecting required data in two stages (username/password first, first/last name later)? This way we wouldn't have to use the magic string placeholders.

This would have UX implications and would require frontend changes, so maybe it's something to consider for the future instead of implementing it now.

agnesgaroux: ➕

kenoir:
Note this resolves: https://github.com/wellcomecollection/platform/issues/6345

rfcs/088-folio-identity-requesting-migration/README.md
kenoir: Yes, I think it would make sense to update the UX to simplify this step, but I think it's out of scope for the migration to make any UX changes.

kenoir: Good question. `externalSystemId` is **not** firmly decided; it's an open question. I've added a note in the identity-model section flagging that what it carries (and whether it's even the right place for the migration join key) may still change before cutover. (71ee1ff)

rfcs/088-folio-identity-requesting-migration/README.md
kenoir: You're right, Sierra is involved: Auth0's custom-DB login script verifies credentials against Sierra and captures the password hash while Sierra is still available (the credential-capture window). I've added a Sierra participant to the lazy-migration sequence diagram so it matches the prose above it. (71ee1ff)

rfcs/088-folio-identity-requesting-migration/README.md
kenoir: Good catch. The `requestable` flag is **additive**, not a replacement: the works page keeps using the location's access status/method to decide whether to show the request button, it's also a prototype affordance rather than a final design; how requestability is ultimately determined is still open (see the requesting business rules open question). I've added a note clarifying both. (71ee1ff)

agnesgaroux (approved):
Let's do it 💪

PR #156, opened 2026-06-18
RFC 089: Identifiers API

kenoir:
## Preview

View the rendered RFC on this branch:

- [RFC 089 README](https://github.com/wellcomecollection/docs/blob/rk/identifiers-api-rfc/rfcs/089-identifiers-api/README.md)
- [Rendered API contract (openapi.md)](https://github.com/wellcomecollection/docs/blob/rk/identifiers-api-rfc/rfcs/089-identifiers-api/openapi.md) · [openapi.yaml](https://github.com/wellcomecollection/docs/blob/rk/identifiers-api-rfc/rfcs/089-identifiers-api/openapi.yaml)

---

## What does this change?

The proposal is a small, read-only Identifiers API. Wellcome Collection gives every catalogue thing (a work, an image, an item) a stable public "canonical" id, and keeps a registry recording which underlying source ids that canonical id was built from. This API does one job: given a canonical id it returns the source id(s) behind it, and given a source id it returns the canonical id (optionally with its siblings). It only ever reads that registry; it never creates or changes ids.

It exists because of the Sierra/CALM to FOLIO/Axiell migration. As records move between systems a single canonical id accumulates several source ids over time (an original plus inherited "predecessor" aliases), and a couple of internal services sit right at the edges where that translation has to happen: the IIIF viewer needs to turn old b-numbers and CALM refs into the canonical id it presents under, and requesting needs to turn a canonical item id into the FOLIO UUID a hold is placed on, and back again. The guiding principle is that everything public speaks canonical and source ids only appear at those two edges (ingest and the FOLIO boundary). Rather than have each consumer re-derive the mapping or query the catalogue by source id, this API is the single shared place that translation lives. Because the main running cost is database queries, it is also the natural place to cache aggressively (at the edge, to keep requests off the database) and to attribute that database cost to the consumers driving it.

How it relates to the other RFCs:

- It reads the ID Registry defined in **RFC 083** (stable identifiers). RFC 083 owns the data and all the writes; this API is just a read-only window onto it.
- It serves the IIIF/DDS lookup that **RFC 085** (IIIF identities, open PR #143) describes, where the canonical Work id becomes the IIIF URI and older identifier forms redirect to it.
- It is the concrete "service" answer to the open question in **RFC 088** (the Sierra to FOLIO identity/requesting migration, open PR #153) about how requesting should translate item ids. RFC 088 left that access mechanism open (a direct database read, a sync, or a service); this RFC proposes the service and is for RFC 088 to ratify.

The RFC is written to stand on its own and carries the API contract alongside it, so it can be reviewed without access to the closed discovery/prototype repository where the working prototype lives.

Files added:

- `rfcs/089-identifiers-api/README.md`: the RFC document.
- `rfcs/README.md`: refreshed RFC listing table (RFC 089 row added).
- `rfcs/089-identifiers-api/openapi.yaml`: the OpenAPI 3.0 spec for the two lookup operations (the source of truth).
- `rfcs/089-identifiers-api/openapi.md`: generated human-readable rendering of the spec, browsable on GitHub without a Swagger/Redoc renderer.
- `render_docs.py`, `pyproject.toml`, `.python-version`, `.gitignore`, `uv.lock`: a self-contained `uv` project that validates `openapi.yaml` and regenerates `openapi.md`.

## How to test

- Read `rfcs/089-identifiers-api/README.md` and review the contract, architecture, caching strategy and open questions.
- (Optional) Regenerate the rendered contract and confirm no diff: from the RFC directory, `uv run python render_docs.py`: this validates `openapi.yaml` against the OpenAPI spec validator and rewrites `openapi.md`.

No measurable runtime success criteria; this is a documentation RFC. Success is the RFC being reviewed and providing a clear, self-contained contract and architecture that the team can align on, and a decision on whether this service is the access mechanism for identifier translation in RFC 088.

- The design risks themselves (the caching strategy and the database cost it controls, the unmet FOLIO-item ingestion dependency, item canonical-id stability) are enumerated in the RFC's Open questions section and are intended to be the subject of review.
- The OpenAPI spec is validated and the rendered Markdown is generated from it, reducing the chance of the contract and its human-readable rendering drifting apart.


rfcs/089-identifiers-api/openapi.md
```diff
@@ -0,0 +1,118 @@
...
+| `sourceSystem` | path | yes | string | Source system identifier type. |
+| `value` | path | yes | string | The identifier value within the source system. Must be a single, URL-encoded path segment. Source-id formats are heterogeneous, so this is not pattern-validated; unknown values resolve to 404 rather than 400. |
+| `type` | query | no | string (enum: Work, Image, Item) | Ontology type. Part of the lookup key; defaults to Work. |
+| `include` | query | no | string (enum: siblings) | Set to "siblings" to also return all source identifiers sharing the resolved canonical id. Omit for the canonical id alone. |
+
+**Responses:**
+
+| Status | Body | Description |
+|---|---|---|
+| `200` | n/a | A mapping exists for the supplied tuple. |
```
paul-butcher: Should this be:
 
```suggestion
| `200` | IdentifierSet\|CanonicalIdRef | A mapping exists for the supplied tuple. |
```

rfcs/089-identifiers-api/README.md
```diff
@@ -0,0 +1,412 @@
...
+
+- **`type`** is `Work` | `Image` | `Item`, defaults to `Work`, and is a real key component: a
+  canonical id can carry rows of differing types (cross-type predecessors are allowed).
+- **`isAlias`** is `false` for the earliest-`createdAt` row (the original) and `true` for later,
+  inherited rows. The API derives it so consumers do not have to.
+- **`sourceSystem`** is an open set, not enum-constrained, so a system the gateway has not been told
+  about resolves to `404` rather than a spurious `400`.
+
+**Status codes:** `200` found; `304` conditional GET (matched `ETag`); `400` malformed `canonicalId`
+or an unsupported enum value (rejected at the gateway); `404` no mapping: an unknown id, an unknown
```
agnesgaroux: Is this 
```
`400` [...]  an unsupported enum value
```
consistent with
```
- **`sourceSystem`** is an open set, not enum-constrained, so a system the gateway has not been told
  about resolves to `404` rather than a spurious `400`.
```

rfcs/089-identifiers-api/README.md
```diff
@@ -0,0 +1,412 @@
...
+From [RFC 083](../083-stable_identifiers/README.md), two tables:
+
+- `canonical_ids`: the uniqueness registry. PK on `CanonicalId`. Supports pre-generation
+  (`Status` free/assigned).
+- `identifiers`: the mappings. PK on `(OntologyType, SourceSystem, SourceId)`, FK to `canonical_ids`,
+  secondary index `idx_canonical` on `CanonicalId`.
+
+Consequences for the two lookups:
+
+- **Forward** (canonical → sources) rides `idx_canonical` and returns N rows: cheap, but not a
```
paul-butcher:
```suggestion
- **Forward** (canonical → sources) reads `idx_canonical` and returns N rows: cheap, but not a
```
typo?

rfcs/089-identifiers-api/README.md
```diff
@@ -0,0 +1,412 @@
...
+- **API key in the `x-api-key` header, validated by API Gateway** before the request reaches the
+  Lambda (`x-amazon-apigateway-api-key-source: HEADER`; the `ApiKeyAuth` security scheme in the
+  spec). The gateway enforces keys with no custom authorizer code, and the key identifies the
+  consumer so database cost can be attributed per consumer.
+- **A per-consumer throttle bound to the stage** is a safety valve that caps how many cache-misses
+  one consumer can drive into the database (rate-limiting to protect the backend, not a quota that
+  bills usage). The keys, the throttle, and their stage binding are **not** in the OpenAPI body;
+  they are separate API Gateway resources in Terraform, so re-importing the definition does not
+  disturb them.
+- A **gateway-level regex** on `canonicalId` (`^[a-hjkmnp-z][a-hjkmnp-z2-9]{7}$`) rejects malformed
```
paul-butcher, line 257: Might we want something similar in the reverse lookup?  Validate sourceSystem against a list?

It's almost certainly overkill to try to validate source system/id pairs (e.g. sierra-system-number/b{0-7}.)

rfcs/089-identifiers-api/README.md
```diff
@@ -0,0 +1,412 @@
...
+
+2. **The FOLIO-item ingestion dependency (RFC 088).** The requesting translation (canonical item id
+   ↔ FOLIO item UUID) has no data in the registry yet: `folio-item-id` identifiers are absent. This
+   API cannot serve that translation until the catalogue pipeline ingests FOLIO items and the ID
+   Minter records `folio-item-id` rows. Confirm the pipeline change with the catalogue-pipeline
+   workstream.
+
+3. **Item canonical-id stability through the FOLIO migration.** Items are minted canonically today,
+   but the canonical id must survive Sierra → FOLIO via RFC 083 predecessor inheritance at item
+   level (a FOLIO item UUID added as a predecessor of the Sierra item number). RFC 083's transformer
```
paul-butcher: Should this be the other way around?

```suggestion
   level (a Sierra item number added as a predecessor of the FOLIO item UUID). RFC 083's transformer
```

rfcs/089-identifiers-api/README.md
```diff
@@ -0,0 +1,412 @@
...
+   `sourceSystem` (`?q=b18035978`). The reverse path makes `sourceSystem` a required key component
+   (it is part of the primary key), so a bare-value lookup cannot use the primary-key prefix and
+   would need a secondary index on `SourceId`. Decide whether to support it, and at what cost. Not
+   part of the committed contract.
+
+5. **`isAlias` vs `obsolete` (RFC 085).** RFC 085 wants identifiers tagged by type and possibly an
+   `obsolete` flag (source system retired). That is adjacent to `isAlias` (inherited predecessor) but
+   not identical. Reconcile the two so one representation serves both consumers.
+
+6. **The `type` enum vs reality.** The live registry holds types beyond `Work` / `Image` / `Item`
```
paul-butcher: Can we populate this enum with the types we _need_ to use the API for (those three), and extend it as we need, rather than worrying about the question up front?

rfcs/089-identifiers-api/README.md
```diff
@@ -0,0 +1,412 @@
...
+   Minter records `folio-item-id` rows. Confirm the pipeline change with the catalogue-pipeline
+   workstream.
+
+3. **Item canonical-id stability through the FOLIO migration.** Items are minted canonically today,
+   but the canonical id must survive Sierra → FOLIO via RFC 083 predecessor inheritance at item
+   level (a FOLIO item UUID added as a predecessor of the Sierra item number). RFC 083's transformer
+   changes are described at bib/work level; item-level predecessor emission needs confirming with the
+   pipeline workstream.
+
+4. **Bare-value reverse lookup (RFC 085).** The DDS wants to query by bare value without a
```
agnesgaroux: We get multiple rows with identical `SourceId` though right? eg
```
WHERE SourceId = 'b19871831'
AND OntologyType = 'Work'
AND SourceSystem = 'sierra-system-number';

WHERE SourceId = 'b19871831'
AND OntologyType = 'Work'
AND SourceSystem = 'mets';
```
sierra canonical id -> adyxgsj4
mets canonical id -> ugpnzjrn

We can't tell which one the caller wants, unless we or they know the source system

rfcs/089-identifiers-api/README.md
```diff
@@ -0,0 +1,412 @@
...
+   throttle carries more weight); the concrete `max-age` values for the bounded (migration) and
+   relaxed (post-switchover) phases; whether the `ETag` should stay a weak validator from
+   `(row_count, max(createdAt))` or move to a content hash; concrete **per-consumer throttle limits**
+   (the database safety valve); and the **cost-attribution mechanism** (edge/access logs keyed by API
+   key, or CloudWatch). The prototype emits `Cache-Control` (`max-age=300` forward / `include=siblings`,
+   `max-age=86400` on the immutable bare reverse lookup) and a weak `ETag`, and honours
+   `If-None-Match` with a `304`, as **prototype defaults, not contract decisions**, and as response
+   headers only (no real edge or stage cache).
+
+2. **The FOLIO-item ingestion dependency (RFC 088).** The requesting translation (canonical item id
```
agnesgaroux, line 362: Can we get on with transforming folio items? so we can clarify this sooner rather than later?
What if the item ids are HRID in the aoi-pmh feed? 
Is there a point where we should think about having HRID and UUID in the identifiers DB so as to be able to convert from one to the other? 
(okapi does it as an alternative
`/instance-storage/instances?query=hrid=={hrid}` or 
`/instance-storage/instances?query=id=={uuid}` )

rfcs/089-identifiers-api/README.md
```diff
@@ -0,0 +1,412 @@
...
+  cache keeps the most traffic furthest from the database, so it is preferred; the stage cache is
+  kept only as a possible secondary layer (see [Caching](#caching)).
+
+---
+
+## Open questions
+
+Each has a prototype direction but an unsettled integration point.
+
+1. **Caching and cost.** The cache placement (edge/CloudFront as primary vs the API Gateway stage
```
paul-butcher, line 346: Can the caching and cost questions be "suck it and see", or do we need to think more carefully about when and where we expect the lookups to happen and how that might change the access patterns?

Current expected clients include:

*  Producing the metadata document upon ingestion of digitised material (probably mostly unique or thereabouts)
* The Items API (unless this is handled within the pipeline) more chance of repeated requests

rfcs/089-identifiers-api/README.md
```diff
@@ -0,0 +1,412 @@
...
+want.) The API Gateway **stage cache** remains available as a secondary layer, but it sits behind the
+gateway, so it saves less than an edge hit. This is a candidate, not a decision.
+
+**Freshness is direction- and time-dependent.**
+
+| Response | Freshness | Treatment |
+|---|---|---|
+| Reverse, bare (`source → canonicalId`) | Immutable once minted | Long TTL, even mid-migration. |
+| Forward (`canonicalId → sources`) | Alias set can grow during migration | Bounded TTL + ETag. |
+| Reverse, `include=siblings` | Carries the canonical → sources set | Same as forward. |
```
paul-butcher, line 286: If we offer a reverse + specific sibling, it could be more cacheable.

e.g. `?include=sierra-system-number`

This may not always be true, I suppose, but as the id database has migrated from a 1:1 mapping of old system ids to canonical ids, I think this means that there would continue to be a 1:1 mapping from ids in one scheme to ids in another. 

That said, doing so may be undesirable or bring along other complications:

One of the envisaged use cases (digitisation) is that, given a folio identifier, we can find the corresponding bNumber for digitisation, but if that does not exist, use the canonical identifier.  So we'd either want to always return canonical if it exists, or include canonical in the list of requested returns.  In which case, we could not rely on the absence of the requested sibling(s) to provide a 4xx response.

I think that might make it a bit tricky to generalise, because if we are going from a new system id to old system id (like this), then we can expect the result to be unchanging, but old to new might just mean it's not been migrated yet.

rfcs/089-identifiers-api/openapi.md
kenoir: The cell was `n/a` because `openapi.md` is generated and the renderer (`render_docs.py`) only handled single-`$ref` bodies, not `oneOf`. The source `openapi.yaml` already declares `oneOf: [CanonicalIdRef, IdentifierSet]`, so I fixed the renderer to emit `oneOf`/`anyOf` bodies as the alternatives joined by an escaped pipe and regenerated the file. The cell now reads `CanonicalIdRef \| IdentifierSet` (linked, in spec order). Fixed in 87b7c38.

rfcs/089-identifiers-api/README.md
kenoir: These are consistent, but the wording was too loose. `type` (`Work` | `Image` | `Item`) is the only enum-constrained parameter, so a bad `type` is the `400` case; `sourceSystem` is an open set, so an unknown one is `404`. I've named the field explicitly: "an unsupported `type` enum value". Fixed in 30be607.

rfcs/089-identifiers-api/README.md
kenoir: Changed to 'reads' in e2c9202.

rfcs/089-identifiers-api/README.md
kenoir: It was backwards. The Sierra item number is the predecessor that lets the new FOLIO item UUID inherit the existing canonical id, matching the work-level Sierra-original plus folio-instance-alias case. Swapped in 70274aa.

rfcs/089-identifiers-api/README.md
kenoir: This matches the committed contract; the spec already enums to exactly those three. I've reframed open question 6 to record the decision: scope `SourceIdentifier.type` to the three types the API needs today and extend the enum on demand, rather than modelling the full registry up front. Done in f053339.

rfcs/089-identifiers-api/README.md
kenoir: Good instinct, but I'd rather not validate here. Any list we check against would need to track the source systems in the registry, and that set is owned by the ID Minter (RFC 083) and keeps growing through the Sierra to FOLIO migration. `folio-item-id` isn't in the registry yet, so a list would block it until we updated and redeployed.

And an unknown system still has to reach the read to find out there's no mapping, so the `404` covers it anyway. I'll keep `sourceSystem` an open set.

rfcs/089-identifiers-api/README.md
kenoir: Added the two clients to open question 1 in ba64d52: digitisation metadata ingestion fetches mostly unique ids, the Items API is more likely to repeat. The actual TTL and throttle numbers can stay suck-it-and-see once we have real traffic, but the access patterns pull in opposite directions, so they are worth recording up front.

rfcs/089-identifiers-api/README.md
kenoir: Good idea, and it fits the freshness model. Two wrinkles worth flagging: the cache win only really lands in the new-to-old direction (folio to bNumber, which is the digitisation case), since old-to-new might just be un-migrated yet and stays bounded; and because the registry is one-to-many it returns a filtered set rather than a single value, with an absent sibling coming back as canonical plus a `200`, not a `404`.

I've folded it into open question 4 (bare-value reverse lookup) in 20ccec9, since the two are the same kind of projection and are best settled together rather than widening the `include` enum now.

rfcs/089-identifiers-api/README.md
kenoir: Yes, exactly. Added this to open question 4 in 5816ebb: the bare value isn't just expensive to index, it can be ambiguous, since the same `SourceId` can sit under different source systems and resolve to different canonical ids. So a bare-value query may have to return multiple candidates or ask the caller to disambiguate. The committed contract keeps `sourceSystem` required for exactly this reason.

rfcs/089-identifiers-api/README.md
kenoir: Good shout. Folded the HRID/UUID question into open question 2 in a2fe582: FOLIO records carry both forms, so we need to confirm which the OAI-PMH feed delivers. The registry can already hold both per item, so the Minter could record both and we'd serve HRID to UUID directly, but okapi resolves the two natively, so storing both is an optimisation rather than a requirement.

On getting on with the FOLIO item transforms, agreed it would clarify this sooner, but the scheduling sits with the catalogue-pipeline workstream rather than this RFC. The dependency is recorded in open questions 2 and 3 so it won't get lost.

agnesgaroux (approved):

rfcs/089-identifiers-api/README.md
```diff
@@ -0,0 +1,445 @@
...
+   serve HRID ↔ UUID translation directly; since okapi already resolves the two forms natively,
+   storing both is an optimisation rather than a requirement.
+
+3. **Item canonical-id stability through the FOLIO migration.** Items are minted canonically today,
+   but the canonical id must survive Sierra → FOLIO via RFC 083 predecessor inheritance at item
+   level (a Sierra item number added as a predecessor of the FOLIO item UUID). RFC 083's transformer
+   changes are described at bib/work level; item-level predecessor emission needs confirming with the
+   pipeline workstream.
+
+4. **Decided: no bare-value reverse lookup for now (RFC 085).** RFC 085's identity service is WorkID-level:
```
tomcrane, line 378: I can see the secondary index argument _against_ supporting this, but for a consumer, it's not as clean as it could be. Consider an incoming IIIF manifest request to the DDS (iiif-builder):

```
iiif.wellcomelibrary.org/presentation/{identifier}
```

What does the DDS do with this? We have no other context at this point, we don't know what source system `{identifier}` corresponds to. This is the first point of contact.

The DDS current behaviour has lots of knowledge of what identifiers look like and how they work. First, we see if it's a b number, or a [permitted variation on a b number](https://github.com/wellcomecollection/iiif-builder/blob/main/src/Wellcome.Dds/Wellcome.Dds.Common/WellcomeLibraryIdentifiers.cs). If the identifier is a permitted b number variant form, we normalise it to the standard form:

```
// 7-digit form:
https://iiif.wellcomecollection.org/presentation/b1803597
// calculate missing checksum (8) and redirect to https://iiif.wellcomecollection.org/presentation/b18035978
```

_This was needed pre wc.org, when there were generated links from the Sierra catalogue to wellcomelibrary.org viewer pages using the pure numerical id `1803597`. Now the Sierra catalogue is kept hidden from the public, so 7-digit links are unlikely, but the code still supports them._

We then see if we have the manifest stored under this key in S3. If so, great, we return it.

If it's not in S3 at the normalised b number form, or if it wasn't even a b number, then the DDS [queries its Manifestations table](https://github.com/wellcomecollection/iiif-builder/blob/1596f5672836f26514dcde163bf6efafe94ed5a2/src/Wellcome.Dds/Wellcome.Dds.Repositories/DdsContext.cs#L152-L159), populated when the item was processed: information about its identities that DDS hangs on to but is not **authoritative for**. This allows us to handle things like:

```
https://iiif.wellcomecollection.org/presentation/zjytxny8
// redirect to https://iiif.wellcomecollection.org/presentation/b18035978

https://iiif.wellcomecollection.org/presentation/dpv7wsth
// redirect to https://iiif.wellcomecollection.org/presentation/b19995271

https://iiif.wellcomecollection.org/presentation/SB/1/1/298
// redirect to https://iiif.wellcomecollection.org/presentation/b19995271
// Note these last two are the same redirect, both work ID and Calm Ref redirect to the b number it was digitised under
```

In the future, the workID form becomes the canonical one, and we will redirect all other forms to that.

If we have a _bare-value_ lookup - "tell us what you know about 'b19995271'" - then the DDS can defer to an authority for the identity. If not, we'll still have to do some tricks locally.

When we see `b19995271` or `b1999527` for `{identity}`, we know it's a b number and not a work id, because we can run it through the code linked above. But DDS has to retain that b-number-aware code to do so. And then, it can *specifically* ask for 

```
/v1/identifiers/by-source/sierra-system-number/b19995271?type=Work
```

or if we see `SB/1/1/298` for `{identity}`, the presence of `/` characters may trigger us to ask for:

```
/v1/identifiers/by-source/calm-ref-number/SB/1/1/298?type=Work
// identity service needs to handle the unbounded trailing {**path} elements!
```

... and if the worst came to the worst, we could ask for each source system in turn to see if we have a match. The more identifier format knowledge we retain in the DDS, the more likely we ask for the right source system on the first attempt.

Consider the string `b3288596`

Is this a work ID or is it a b number? It's actually valid under both schemes. It matches the 8-character work ID reduced alphanumeric set, AND expands to a valid b+8-digits b number:

[iiif.wellcomecollection.org/presentation/b3288596](https://iiif.wellcomecollection.org/presentation/b3288596)

We don't know if we need to redirect or or not, until we've made a call to the identity service to resolve it.

How many work IDs are there that are a `b` followed by 7 digits? Probably not many. And it's even more unlikely that any of them are real Sierra b numbers. And if there are any of those, it's astronomically unlikely to be the same work! 

The question is, how much knowledge of identifier forms do we want DDS to retain? It could defer everything to the identity service (with a short term memory cache), always resolving `{identity}` using the service and never deducing anything about it. If on receipt of the query results, `{identity}` turns out not to have been the canonical form, then redirect to the canonical form. This reduces the knowledge the DDS has, AND works for any future identifier scheme, but at the expense of that secondary index!

kenoir: > The question is, how much knowledge of identifier forms do we want DDS to retain?

Some, and I think the floor is set by the data rather than by this API. The "defer everything
to this service" option doesn't quite work at this level, even if we paid for the secondary
index, for two reasons.

First, normalisation stays with the DDS either way. The registry only holds the forms the
minter minted from: b18035978 is a row, 1803597 and b1803597 are not. So the
checksum/variant handling in WellcomeLibraryIdentifiers.cs has to run before any lookup,
bare or qualified. A bare-value endpoint would remove the need to guess the source system,
not the need to normalise.

Second, a bare-value lookup can enumerate the ambiguity but it can't resolve it, and the
ambiguous case is routine rather than rare. The registry is pre-merge: as @agnesgaroux pointed out
above, every digitised b-number sits under both sierra-system-number and mets, resolving
to two different canonical ids (the public work and the invisible METS work), and both rows
are type=Work so the type filter doesn't split them. Which of those canonical ids is the
public one is decided by the merger, and the registry doesn't know that, so this API has no
basis to rank the candidates. "Tell me what you know about b19995271" comes back as two
labelled matches and the DDS still holds a precedence rule (prefer sierra-system-number
over mets). We'd have swapped format-parsing for a selection policy, plus the index. And
the b3288596 collision straddles the canonical and source id spaces, so a bare source
lookup doesn't settle it either; the same policy decides.

PR #158, opened 2026-06-26
RFC 088: Record the barcode sequence service

kenoir:
**Preview (rendered on this branch):** [RFC 088 README](https://github.com/wellcomecollection/docs/blob/rk/rfc-088-barcode/rfcs/088-folio-identity-requesting-migration/README.md) · [openapi.md](https://github.com/wellcomecollection/docs/blob/rk/rfc-088-barcode/rfcs/088-folio-identity-requesting-migration/openapi.md)

Amends the merged RFC 088 to record how new patrons get a library-card barcode, replacing the now-superseded description.

## Why

RFC 088's **Open Question 4 (Barcode and role)** still described the old approach: new users received the bare Auth0 id as their barcode, with an open task to "verify the 24-character barcode format". That approach was replaced during prototyping. New patrons are now minted a numeric, sequential card number by a small **sequence service** (a DynamoDB atomic counter) at first login, which keeps the printed and scanned card representation numeric, as it is today. The RFC was therefore wrong on this point; this corrects it and resolves the open question down to the parts that genuinely remain open.

## What changed

- **Open Question 4** rewritten to record the numeric-sequence decision. What stays open is narrowed to agreeing the concrete seed, prefix and width with the LMS migration and the card supplier, confirming reader support (scanners and self-issue kiosks), plus the existing patron-group-to-role confirmation. The stale "24-character barcode format" concern is removed.
- **Login and enrichment** gains a short note that `/m2m/enrich` mints the barcode when the FOLIO record has none (allocate-once, best-effort, retried on next login rather than blocking it).
- **API surface** gains a row for `POST /m2m/sequences/{name}/next`.
- **OpenAPI spec** carries the new route in `openapi.yaml`, with `openapi.md` regenerated via `render_docs.py`. All referenced components already existed, so the addition is self-contained.

The RFC index `Last modified` date is synced via the generator script. No other RFCs are affected.

## Scope note

The sequence service is a sub-feature of the v2 identity API, not a standalone service: it shares the m2m Lambda, API Gateway and authorizer, lives inside the folio-api spec, and has no consumers outside the identity API. So it belongs in RFC 088 rather than a new RFC.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rfcs/088-folio-identity-requesting-migration/openapi.md
```diff
@@ -422,6 +423,36 @@ token with scope `enrich:read`.
 | `200` | [`EnrichResponse`](#enrichresponse) | Enrichment payload for the Auth0 action. |
 | `404` | [`IdentityError`](#identityerror) | No Folio user could be resolved. |
 
+#### `POST /m2m/sequences/{name}/next`
```
agnesgaroux, line 426: Are we going to use this for anything other than creating patron barcodes? If not, why not give it a more descriptive name? This is both generic and specific ("returns it formatted as a barcode") so a bit confusing

kenoir: We might, it does sequences, what it outputs isn't barcode specific so i'm not to too fussed about the name.

PR #163, opened 2026-08-05
RFC 089: amend for the RFC 091 preservation-identifier decision

kenoir:
## Preview

- [Amended RFC 089 README](https://github.com/wellcomecollection/docs/blob/rk/rfc-089-preservation-id-amendment/rfcs/089-identifiers-api/README.md)

---

## What does this change?

Amends the merged RFC 089 (Identifiers API) to reflect the decision recorded in RFC 091 (#159): preservation identifiers stay the canonical identifiers for digital objects, and nothing is minted at ingest. RFC 089's framing was written against the pre-inversion RFC 085, in which the canonical Work id was to become the IIIF Manifest URI, and two parts of it no longer described reality.

- **The canonical-first principle gains its deliberate boundary.** Digital objects speak preservation identifiers publicly (the b-number, CALM ref, or newly assigned id that Goobi, Archivematica and the storage service share), a permanent identifier space alongside canonical and source ids rather than an exception to be migrated away. Canonical-first continues to cover catalogue-level entities.
- **The consumer list is rewritten.** Requesting (RFC 088) is unchanged and remains the primary driver. The digitisation ingest app (RFC 091) is added: it reverse-looks-up `folio-instance/<folio-id>` with siblings to find a `sierra-system-number` predecessor, which is the cross-migration/post-migration signal. The DDS moves to a reduced role: the Work id to preservation identifier binding it needs for redirects is merge-derived, so it lives in the catalogue API and the DDS's own records, not this registry; what remains for this API is source-identifier resolution and classification, on the unchanged contract.
- **The read-only stance is recorded as settled.** The secured resolve-or-mint extension the earlier draft of RFC 091 proposed was the only write path ever suggested for this API, and it is withdrawn.
- **Open question 4 (no bare-value reverse lookup) carries the post-merge review outcome**: normalisation of variant forms stays in the DDS because the registry only holds minted forms, and a bare lookup could enumerate ambiguous matches but not rank them, since which canonical id is public is merger knowledge. The decision to keep `sourceSystem` required stands, now with the consumer-side reasoning recorded.

The API contract, architecture, caching strategy and data model are untouched; this is a framing amendment, not a design change.

Files changed:

- `rfcs/089-identifiers-api/README.md`: the amendments above, plus a refreshed last-modified date.
- `rfcs/README.md`: listing table regenerated (the 089 row moves with its new date).

## How to test

- Read the [amended RFC](https://github.com/wellcomecollection/docs/blob/rk/rfc-089-preservation-id-amendment/rfcs/089-identifiers-api/README.md), in particular the canonical-first principle, the consumers, and open question 4.
- The OpenAPI spec is untouched, so no regeneration of `openapi.md` is needed.

No runtime criteria; success is RFC 089 and RFC 091 describing the same system, with no reader able to infer from RFC 089 that the Work id becomes the manifest URI or that a mint endpoint is coming.

- Documentation only; no production code or infrastructure changes.
- The amendment narrows the stated DDS dependency on this API. If the DDS integration later needs more than source-identifier resolution, that is a new requirement to raise against the contract, not something this amendment forecloses.

## Notes for reviewers

- Reads best alongside RFC 091 (#159), which records the decision this amendment propagates, and the RFC 091 discussion notes on the RFC 085 branch, which informed the DDS-side framing.
- RFC 085 (#143) will be amended separately to invert its "Use the Work ID" position.


agnesgaroux:
👍

